PR #2, opened 2017-05-09
STRIPES-360 https for placehold.it

julianladisch:


julianladisch:
Replace by local image: https://issues.folio.org/browse/STRIPES-377

PR #11, opened 2017-07-19
Stcom 21 arrow functions

zburke:
Clean up ESLint complaints about how arrow functions are used. See STCOM-21.


MikeTaylor:
Down to 264 lint errors! Woohoo! :-)

PR #23, opened 2017-07-24
Add support for unsaved changes to EditableList. STCOM-26.

mkuklis:
@JohnC-80 and @zburke this PR introduces a new version of `<EditableList>` with a support for navigation check in cases when the unsaved changes are detected. @JohnC-80 the work is based on `FieldArray` (thank you for recommending it). It allowed for easier state management for each item inside the form. 

I tried to keep the current API for `<EditableList>` aligned with the previous version so it stays compatible with the apps. It seems like things are working well but it would be great if you could review it. Thank you!


mkuklis:
Thank you for the review @zburke and for catching these extra issues! It appears that there is another flag in `redux-form` called `keepDirtyOnReinitialize` which keeps the dirty fields unchanged on reinitialization. I found it via this discussion: https://github.com/erikras/redux-form/issues/370 
This should fix these additional issues. @JohnC-80 I'm going to merge this PR to master. I hope you don't mind ;)

JohnC-80:
Michal,
I like the cleanliness of things, but I believe I’ve encountered a bug.
If you edit/save an item with a change that will change the sorting order of the list, the item keeps its place, but the fields seem to update/show values from the updated order. To reproduce:
Patron groups has  groups in this order:
faculty – Faculty Member
graduate – Graduate Student
staff -  Staff Member
undergrad – Undergraduate Student

Click edit on the second item - graduate – add a 2 to the beginning of the name  “2graduate” (that would send it to the beginning) and save. Now the fields read
faculty – Faculty Member
2graduate – Faculty Member
staff -  Staff Member
undergrad – Undergraduate Student

It’s like it’s used the description value for the 2nd item from the proper sort order – The appropriate values are still there somewhere ( I can edit > cancel on the first and sedond items and they appear as they should be)
Just not being properly rendered/updated in the right places.

mkuklis:
@JohnC-80 Thank you for your review and for catching this issue. Unfortunately it looks like this started happening after I added `keepDirtyOnReinitialize`. It also looks like `keepDirtyOnReinitialize` doesn't play well with our own `navigationCheck`. I'm going to remove it for now. It means that what @zburke has discovered will be still reproducible but it shouldn't be a big deal (if it is we may take another take on this :)).

PR #27, opened 2017-07-27
Lint: FilterControlGroup. See STCOM-21

zburke:


MikeTaylor:
Why was it necessary to change a <label> to a <div>?

zburke:
Lint issues the error, "Form controls using a label to identify them must be programmatically associated with the control using htmlFor". See https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md. 

IOW, a label has to be associated with one and only one input element so you can't use a label for a checkbox group or a radio button group. We could wrap the whole thing in a `<fieldset>` with the old `<label>` replaced with `<legend>`. Do you think that would be better?

MikeTaylor:
Hmm. I'm pretty sure that's a change in ESLint since I wrote this code, because I think it passed lint cleanly back in the day. But, fine -- so long as your solution doesn't affect the rendering, it's fine by me. I was just curious.

PR #34, opened 2017-08-07
use "title" prop for stringValue. Refs STCOM-36

zburke:
If a non-string value such as a valid React Element is handed to a cell
for rendering and that element has a title property, use it instead of
displaying a warning.

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -432,8 +432,13 @@ class MCLRenderer extends React.Component {
       }
 
       if (typeof value === 'object') {
-        console.warn(`Formatter possibly needed for column '${col}': value is object`, value);
-        stringValue = null;
+        if (React.isValidElement(value) && value.props.title) {
+          stringValue = value.props.title;
+        }
+        else {
+          console.warn(`Formatter possibly needed for column '${col}': value is object`, value);
+          stringValue = null;
+        }
       }
 
       if (typeof value === 'boolean') {


MikeTaylor:
Idle curiosity: what is the use-case for this?

zburke:
This follows from the conversation on #coreteam last week: https://folio-project.slack.com/archives/G51D075GU/p1501855026664784 in which we noted that passing a non-string element (a <select> object in this case) to MCLRenderer causes it issue console warnings because it can't decide what to use for a title property. 

The cause of the test failure isn't clear, so it likewise isn't clear this change will resolve the failure, but this does prevent the warning in the case where we have a title property.

MikeTaylor:
Oh, I see -- so it's not new functionality, it's better detection of certain accidental conditions.

mkuklis:
@zburke this looks very nice!

JohnC-80:
Ideally the solution for this would not warn for passing any component through to the grid - with or without a 'title' prop. If you pass <Button> without a title, the warning still shows up. Further - I think if a component is set up to render/use a 'title' prop, the MCL should just let it.

MikeTaylor:
Sure, but we know from experience that we often end up inadvertently passing things into MCL that are not what we intend. From my perspective as a _client_ of MCL, the most important thing is to get a clear and explicit error-message when I make a mistake.

JohnC-80:
Absolutely - in this case what I think we want is to **not** warn a client when they're doing something that's fine - i.e. passing a component through to the grid. You should be able to send a component through to the grid without any noise on the console. The MCL was warning if it found *any object there (react components included) and it really shouldn't behave that way. It should still warn if a plain js object is passed in - it doesn't know what to do with that and there's no suitable default.

JohnC-80:
As far as more explicit, helpful warningss, maybe a link to [Formatter documentation](https://github.com/folio-org/stripes-components/blob/master/lib/MultiColumnList/readme.md#formatter) would be helpful there.

JohnC-80:
For info's sake, It appears now that attempting to render a plain js object will draw out an error from React itself - suggesting an add-on that I haven't seen before. That's definitely an unnecessary hoop I don't want a dev going through since we have formatters... so catching/not attempting to render might be a next iteration here... 
image

zburke:
Restating to make sure I understand what we want: 

1. components which can be successfully rendered should pass through MCL without warning, regardless of whether they have a title
1. other objects passed to MCL will generate a warning. 

Do we care about the `title` property at all? This whole thing started not because of trouble rendering the object but because MCL wanted to apply a `title` property to the `<div>` it wraps around every cell value. Do we want to extract and use that property in the wrapper div when available, or do we not care about it since the rendered element will also have a `title` property?

JohnC-80:
Your 2 points are correct. 
Regarding the title - It's fine that the title prop is applied the cell's div **if** the title prop exists - but any given component may or may not have a prop named 'title' to access. The MCL should not base whether or not it sends a console warning about using formatters based on the presence of that title prop on the component.
The title attribute works nicely for values that could be truncated, translated, or cut off somehow. The `stringValue` variable is applied to both the `title` attribute and the way that the row will be 'spoken' through a screen reader ([Line 465](https://github.com/folio-org/stripes-components/blob/master/lib/MultiColumnList/MCLRenderer.js#L465)) when someone is tabbing through the list - so it's nice for rendered components to have a way to supply that via a `title` prop. If `stringValue` is null - it simply won't render the attribute/won't speak the value on initial focus of the row.

zburke:
OK. I'm going to close this PR in favor of another that copies the title, when available, onto the wrapper div and does not issue console warnings when displaying a component.

PR #38, opened 2017-08-07
do not warn on component display. Fixes STCOM-36

zburke:
There is no need to issue a warning when displaying a React component as
it will be properly rendered.

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -431,7 +431,9 @@ class MCLRenderer extends React.Component {
         stringValue = value;
       }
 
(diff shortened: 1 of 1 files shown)


JohnC-80:
Minor thing - would it be better to check for the object type on the outside, then check if it's a valid component and set the string accordingly? This code will basically be checking if any and every value passed through is a component - I'm assuming it's better to know its an object before the possibly less efficient `React.isValidElement` call. I know, I know... *dodges thrown object.

zburke:
No worries; that's a legit criticism.

JohnC-80:
Slick. I'll merge this guy in after the demo today. Thanks!!

PR #39, opened 2017-08-09
Fix compilation errors in index.js

cowboyd:
`lib/index.js` currently re-exports the `default` export from several packages as its own. However, every JS module can only have at most a single `default` export. This was giving rise to errors like:

```
ERROR in ../stripes-components/index.js
Module build failed: SyntaxError: Only one default export allowed per module. (3:8)

  1 | /*form elements*/
  2 | export {default} from './lib/Button';
> 3 | export {default} from './lib/Checkbox';
    |         ^
```

This change re-exports each referenced component module as a named
export so that you can import them by name. E.g.

```js
import { Button } from '@folio/stripes-components';
```

> Note: `InputList` does not appear to be in the codebase anymore, so
> that export was removed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,27 +1,25 @@
 /*form elements*/
-export {default} from './lib/Button';
-export {default} from './lib/Checkbox';
-export {default} from './lib/RadioButton';
-export {default} from './lib/Select';
-export {default} from './lib/Textfield';
-export {default} from './lib/InputList';
+export { default as Button } from './lib/Button';
+export { default as Checkbox } from './lib/Checkbox';
+export { default as RadioButton } from './lib/RadioButton';
+export { default as Select } from './lib/Select';
+export { default as Textfield } from './lib/Textfield';
 
 /*data containers*/
-export {default} from './lib/KeyValue';
-export {default} from './lib/LayoutGrid';
-export {default} from './lib/MultiColumnList';
+export { default as KeyValue } from './lib/KeyValue';
+export { default as LayoutGrid } from './lib/LayoutGrid';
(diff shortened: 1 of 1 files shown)


MikeTaylor:
@JohnC-80, does any client code even use this top-level index? I am guess it does not, otherwise how could it have been working?

Is there any reason not to go ahead and merge this?

JohnC-80:
Everything references via the @folio scope - so no, nothing's using it as far as I know. I'm for merging it I just haven't had a chance to try it out yet.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

MikeTaylor:
I think we're at cross purposes. We'll keep needing the `@folio` scope whatever we do. Using that top-level will only be the different between ```import Pane from '@folio/stripes-components/lib/Pane'``` and ```import Pane from '@folio/stripes-components'``.

I won't be getting Outlook for Android, but thanks for the suggestion.

JohnC-80:
Merged! Thank you! I agree with this opening us up to move things around within the library. Sorry for short comments/late response on this... I'm technically 'off work' today.

MikeTaylor:
No problems! In this PR, we integrated a fix, reached a decision on Stripes "best practice" and learned something about GitHub notifications. I call that a good payoff!

skomorokh:
I agree that it would be nice to import everything from the top level, however, this will probably wind up with us importing the whole of stripes-components up front unless we're quite careful. Might be okay though as we have the luxury of these imports being purely ES6 modules without any CommonJS.. All theoretical so far as we've not yet reimplemented code-splitting/async loading.

https://advancedweb.hu/2017/02/07/treeshaking/

PR #44, opened 2017-08-15
show overflow in wrapper elements. Refs UIU-128.

zburke:
The wrapper elements around a list, in addition to the individual cells,
need to allow visible overflow in order for the cells' overflow settings
to bubble up. This is necessary to allow overflow in the trailing cells
of a list; otherwise their overflow is simply cropped by the wrapper
element.


JohnC-80:
I'm hesitant to remove overflow from the container entirely - I think overflow: hidden could at least be placed there if the virtualize prop is present. The .scrollable div receives the overflow: hidden style dynamically if there's a width or height prop present.

zburke:
Changing the container's overflow was primarily to deal with a dropdown menu that was being clipped by its parent element, but #48 handles that, making this irrelevant.

PR #54, opened 2017-09-15
Upgrade ESLint and fix broken rules

cherewaty:
I noticed on our builds of `thefrontside/ui-eholdings` that we were still getting `WARN: 'Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs'`, even though I thought it had been resolved with https://github.com/folio-org/stripes-components/pull/45.

It turns out I missed one - in searching for `React.PropTypes`, I didn't notice an instance of `import React, { PropTypes } from 'react';`.

That would've been caught by the ESLint `react/no-deprecated` rule provided as part of `eslint-config-airbnb`, but the error was not being thrown. I upgraded all the `eslint`-related packages, and then there were errors in my editor and when running `yarn lint` (`react-no-deprecated` also needed to know the target `react` version to pop up).

I fixed all the linting errors, with only a few uses of `eslint-disable-line`.

### Next Steps
It would be awesome if we ran the linter as part of a continuous integration environment, so future pull requests to `stripes-components` with ESLint errors could not be merged.

### Follow-up Work
I turned off `jsx-a11y/label-has-for` since it had a lot of breakages that needed extensive refactors with probable css changes. The rule was usually thrown because `<label>` elements weren't wrapping their associated checkboxes, radio buttons, etc. https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,6 +5,9 @@
     "import/resolver": {
       "webpack": {}
     },
+    "react": {
+      "version": "15.6.1"
+    },
   },
   "env": {
     "browser": true,
@@ -36,6 +39,7 @@
         tags: [],
         roles: ['tabpanel', 'dialog'],
       },
-    ]
+    ],
+    "jsx-a11y/href-no-hash": "off",
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -41,5 +41,6 @@
       },
     ],
     "jsx-a11y/href-no-hash": "off",
+    "jsx-a11y/label-has-for": "off",
   }
 }
--- a/lib/Accordion/Accordion.js
+++ b/lib/Accordion/Accordion.js
@@ -25,7 +25,6 @@ const propTypes = {
 };
 
 class Accordion extends React.Component {
-
   static defaultProps = {
     header: DefaultAccordionHeader,
     separator: true,
--- a/lib/AuthorityList/AuthorityList.js
+++ b/lib/AuthorityList/AuthorityList.js
@@ -102,7 +102,6 @@ class AuthorityList extends React.Component {
       </Paneset>
     );
   }
-
 }
 
 export default AuthorityList;
--- a/lib/Datepicker/Calendar.js
+++ b/lib/Datepicker/Calendar.js
@@ -88,7 +88,7 @@ class Calendar extends React.Component {
 
     moment.locale(this.props.locale);
 
-    let inputDate;  // initial date for state
+    let inputDate; // initial date for state
     let cursorDate;
     if (this.props.selectedDate === '' || this.props.selectedDate === 'undefined') {
       inputDate = null;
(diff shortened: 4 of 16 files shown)


.eslintrc
```diff
@@ -36,6 +39,8 @@
         tags: [],
         roles: ['tabpanel', 'dialog'],
       },
-    ]
+    ],
+    "jsx-a11y/href-no-hash": "off",
```
cherewaty: https://github.com/airbnb/javascript/issues/1474

JohnC-80: Thank you! Agreed on next steps.

I currently disagree with the idea that wrapping everything is the way to go - I feel like we could settle on 
```
"jsx-a11y/label-has-for": [ 2, {
      "components": [ "label" ],
      "required": {
          "every": [ "id" ]
      },
      "allowChildren": false,
  }],
```
for now - All of the labels do explicitly pass htmlFor.
It's odd that there isn't an 'or' option to these - that it wants both wrapping and id's.
I'm inclined *not* to wrap all of the fields (especially text and select fields) in label tags unless we have some explicit cases where assistive technology won't acknowledge html's basic 'for' attribute... hoping to shed light on these things when the accessibility SIG cranks up.

MikeTaylor:
This PR looks good to me _except_ for the many major-version upgrades of dependencies. I'm not saying we shouldn't make those changes, but that if we do we should co-ordinate them across all the 21(!) stripes-foundation and ui-module repos.

Can we merge all the linting fixes on the code itself, but hold off on the eslint version-upgrades until we have an integrated plan?

JohnC-80:
Didn't the necessary linting fixes arrive as a result of the version-upgrades?.... hopefully merging those fixes won't trip the current versions....

cherewaty:
I'm hesitant to separate the version upgrades from the fixes, since the ESLint ruleset defined by those dependencies changes with the new versions. I made the upgrades to begin with because the earlier versions were not catching the `react/no-deprecated` violation.

Since the only upgrades are ESLint-related `devDependencies`, no libraries/apps that rely on `stripes-components` (or even `stripes-core`) should have any issues with the new dependencies. They should be able to continue with their own linting configuration, even if it's behind `stripes-components`.

.eslintrc
cherewaty: I agree that wrapping text and select fields seems odd. I need to do some deeper investigation on this rule.

MikeTaylor:
OK, I went ahead and accepted this. I guess it's a wake-up call that we need to find a better way to co-ordinate our use of dependent modules across Stripes.

JohnC-80:
Yeah... We'll figure it out... Hopefully, the changes between eslint-plugin upgrades won't be too crazy and it won't be too painful to keep those resources in a single place rather than maintain parity between modules. Growning pains!

PR #55, opened 2017-09-22
Use eslint-config-stripes

cherewaty:
Switches to using shared FOLIO eslint config: https://github.com/folio-org/eslint-config-stripes


lib/MenuItem/MenuItem.js
```diff
@@ -18,7 +18,7 @@ const handleItemSelect = (e, props) => {
 };
 
 
-const MenuItem = props => (<div onClick={e => handleItemSelect(e, props)}> {props.children}</div>);
+const MenuItem = props => (<div onClick={e => handleItemSelect(e, props)}> {props.children}</div>); // eslint-disable-line jsx-a11y/no-static-element-interactions
```
cowboyd: Do we need to follow this up with a PR to handle it in it the `jsx-a11y` approved way?

cherewaty: Yes we do.

cowboyd: Ok, let's make sure we've got an issue open somwhere. Otherwise, I'm 👍

JohnC-80: Just a heads-up, I've got a branch 'dropdown-revisions' that resolves the lint issue.. had to do a bit of a surgery there. `<MenuItem>` is a new addition to the new `<Dropdown>` component set. so there's some work-in-progress there as we integrate the component and swap out the old react-bootstrap Dropdown instances...

PR #61, opened 2017-10-02
push all props down to generated components. Fixes STCOM-84

zburke:
In an effort to appease ESLint by removing unused variables, we changed
the way some props destined for a child component were pulled from
`props` by using `separateComponentProps()` but this add some unintended
consequences. Namely, by requiring that those props be explicitly named,
we prevented tools like redux-form from implicitly handling their own
props on the components they managed. IOW, things broke. Things broke
BADLY.

Replacing the naively-conceived `separateComponentProps()` with
`omitProps()` feels kludgy since we're back to hard-coding a list of
props to ignore instead of but since doing it automatically didn't
actually work, the brute-force approach wins for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change history for stripes-components
 
+## 1.9.0 IN progress
+* Really restore missing props to generated component in `<TextField>` and `<Button>`. See STCOM-83. Fixes STCOM-84.
+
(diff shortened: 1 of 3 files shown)


MikeTaylor:
Is there a lesson we can learn from this?

zburke:
I view this as more of a kick in the pants reminding us of something we already know (messing with internals is risky) rather than teaching us something altogether new. And it's a reminder that better testing would serve as well, especially on core libraries.

PR #68, opened 2017-10-11
STCOM-40, STCOM-68 enabler

JohnC-80:
Fix issues with Dropdown component and associated components (<MenuItem>, <DropdownMenu>) and add appropriate classes to Button (dropdownActive, dropdownItem).

Additionally, add props to <Pane> component to set overflow and padding of the Pane's content `<div>`.


JohnC-80:
Agreed! @TG-EBSCO has been a great help in the process!

PR #70, opened 2017-10-13
Add <EntrySelector>

MikeTaylor:
Copied from ui-circulation so it can also be used in other modules.

Fixes STCOM-92.


MikeTaylor:
I can't believe we now have this set up so we can't commit code that has a blank line somewhere that ESLint doesn't like it.

This kind of bs inevitably ends up with people putting `/* eslint-disable */` at the top of all their source files.

ESLint is a servant, not a master.

PR #74, opened 2017-10-19
Refactor EntrySelector. Fixes STCOM-97

mkuklis:
This PR refactors `EntrySelector` to support UICIRC-15 and UIU-276.


mkuklis:
@MikeTaylor would you mind reviewing it. We need this to continue with both UICIRC-15 and UIU-276. Thanks!

PR #76, opened 2017-10-20
Fix connect in Settings. Fixes STCOM-99

mkuklis:
This PR moves `stripes.connect` from `render` into `constructor` so the connection happens only once for each component.


JohnC-80:
Adding @MikeTaylor who has much more experience with the Settings component than I do.

PR #81, opened 2017-11-06
Stcom 110 datepicker non us

JohnC-80:
Fixes issue with datepicker miss-handling dates for non-US locales.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ## 2.0.0 (IN PROGRESS)
 
 * Refactor `<EntrySelector>` to make it more reusable. Fixes STCOM-97.
+* Fix bug in `<Datepicker>` mis-handling non-US date formats. Fixes STCOM-110.
 
 ## [1.9.0](https://github.com/folio-org/stripes-components/tree/v1.9.0) (2017-10-13)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v1.8.0...v1.9.0)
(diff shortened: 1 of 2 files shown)


zburke:
This fixes [STCOM-110](https://issues.folio.org/browse/STCOM-110) part 2, parses selected dates incorrectly, but not part 1, wonky month-calendar widget display. Arguably, part 2 is the more import part since you can type in a date formatted according to the current locale and that's now handled correctly. 

But calendar display is still off for some months when the locale "en-gb" is selected. October 2017 is missing October 1 and contains a full week of November; the first two weeks of January 2017 are still completely out of sorts (the first week is 2017 but is missing January first, the second week is from 2018, and the rest are from 2016).

JohnC-80:
Thanks! I don't know why I was thinking they were separate issues.. I will close this for now and re-submit once the calendar rendering is sorted out.

zburke:
Actually, I would say split the ticket and merge these changes. I only filed one issue because I was thinking/hoping that one fix would rule them all, one fix would find them, one fix would bring them all and in the darkness---, uh, um sorry. 

If these are two separate issues, then let's have two separate tickets and merge this PR since it contains a legit fix. Sound good?

JohnC-80:
Sounds like a plan... I think I'm getting close on the part 1 fix.

zburke:
Calendar-month display wonkiness filed in https://issues.folio.org/browse/STCOM-115.

PR #87, opened 2017-11-13
facilitate additional fields in editable lists that are reactive components

wwelling:


JohnC-80:
Thanks, William! The root of this - is it the need to use inputs other than text inputs for Editable list items? Could you please post an example of what the EditableList API would look like with this in place? (this component really needs a readme.) 
`<AddressEdit>` takes an approach where fields default to TextField, but can be overridden via a fieldComponents prop - bottom of list at https://github.com/folio-org/stripes-components/tree/master/lib/structures/AddressFieldGroup#configuration-props

mkuklis:
@wwelling what is the difference between `visibleFields` and `additionalFields` here?
@JohnC-80 this is partially my fault for missing the readme. I will add it after this work is done.

wwelling:
@mkuklis visibleFields refer to actual properties on a given record. Example: ["group", "desc"] => usergroups.group and usergroups.desc.

Now what if you need to acquire data from backend and aggregate date into a single field. This is where additionalFields come in.

An additional field is an object with the gloss and the actual React.Component class. The are connected at the top level and constructed while being pushed into the fields of the ItemView.

wwelling:
@JohnC-80 edit is not a concern at this time. Only view of complex fields is needed and accomplished here.

wwelling:
The only limitation is that the order. VisibleFields will be first and then additionalFields. I felt this was somewhat out of scope for the feature I was assigned. Would not be a very difficult enhancement later.

mkuklis:
@wwelling this looks very nice! I remember I heard about this requirement (ability to display the totals). I'm glad we have it now :). Yes the only comment I had was what you just mentioned (about the order of fields). I was thinking perhaps we could change `visibleFields` and make them more customizable (so you could specify if you want to make the given field editable or if it's a reactive/connected/read-only field). Not sure if it's worth it now. I think we can always refactor if we have to.

mkuklis:
@wwelling I would also wait with merging before we hear from @JohnC-80 .

wwelling:
@mkuklis thanks. I will wait on John to comment. The ordering should be easy to code in when the use case comes up.

JohnC-80:
Sorry guys.. looking at the PR's across all of the repos for this... The way ControlledVocab passes in the records from stripes-connect as contentData seems a bit on the hard-wired side... it seems odd that the list would need more 'connected' points besides its single source... I would like for those more involved with stripes-connect to weigh in as well. I'm for versatility on these components, but at the same time, if it's really a stripes-connect issue, I would rather solve it there. @skomorokh thoughts on this one?

wwelling:
Thanks @JohnC-80. It does seem like your concerns are not directly related to the changes I have made, but more of an overall design for the components involved. I made an effort as to not make any design changes and approached the features by extending current functionality. There may be better ways, but more than likely require design changes.

JohnC-80:
@wwelling Yes - I definitely gathered that in your approach of adding-to rather than changing. Thanks you for that! I try to keep some reference of how the component API's work in other places and repeat those patterns if we can - That said, another way this may end up being ultimately resolved is through use of <MultiColumnList> rather than <List> as part of <EditableList> - it has `formatters` that go down to the actual content of the 'cells' - so the ordering, content, even the edit mode do get covered there. You can see how that works here... https://github.com/folio-org/stripes-components/tree/master/lib/MultiColumnList#formatter - I've got to add in column headers for https://issues.folio.org/browse/UIU-264

wwelling:
Thanks @mkuklis, @JohnC-80, and @skomorokh for the reviews. I will merge this as soon as the backend supporting [raml](https://github.com/folio-org/raml/pull/30) and [mod-users](https://github.com/folio-org/mod-users/pull/40) are approved. @JohnC-80 this was a good learning experience for me and I am not concerned of inevitable change. @skomorokh I was not aware of a plan to have a service facilitate complex fields for lists. That may have been more complicated and definitely more difficult to maintain. I hope this will suffice for the time being. I look forward to [GraphQL](http://graphql.org/) integration. There will be many simplifications on the front end.

PR #89, opened 2017-11-14
Set up Storybook development environment

cherewaty:
## Purpose
"[Storybook](https://github.com/storybooks/storybook) is a development environment for UI components. It allows you to browse a component library, view the different states of each component, and interactively develop and test components." Storybook should make component development and styling a breeze, since each component can easily run in isolation.

## Approach
We only set up a single story for the first component in alpha order, `Accordion`. We also pulled in the `storybook-readme` addon to show off the READMEs already present in `stripes-components`.

`yarn storybook` will spin up the Storybook environment at [http://localhost:9001](http://localhost:9001).

### The global CSS problem
For a component like the `Accordion` to render 100% visually styled in Storybook, it needs to know about styles previously defined globally in `stripes-core` like `font-family`, `letter-spacing`, etc. To facilitate Storybook's completeness, we moved the few global styles defined for FOLIO to `stripes-components` (another PR incoming for `stripes-core` to import those styles from `stripes-components` instead of defining them itself).

We should be vigilant about avoiding bloat in `global.css`.

#### Alternative
Another possible approach: don't define _any_ global styles, and have every component be 100% responsible for its own styles. It's the purer, most micro-servicey strategy, with zero style cascading. The `stripes-components` repo could contain mixins to help component authors manage things like `font-family`.

###### lib/mixins.js
```
  --stripes-font: {
    font-family: 'Source Sans Pro';
    font-size: 14px;
    letter-spacing: 0.04em;
  };
```

##### lib/Accordion/Accordion.css
```
  .accordion {
    @apply --stripes-font;
  }
```

##### Why not the alternative?
- More CSS delivered in the bundle
- More boilerplate in component CSS
- Consuming applications would have to also use the mixins to ensure consistency
- App developers have to remember to include the base styles in every component they build in their own apps
- Proven federated design systems like [Salesforce Lightning](https://www.lightningdesignsystem.com) and [Shopify Polaris](https://polaris.shopify.com) still define a limited set of styles in the global namespace

## Next steps
- Instead of defining global styles in `stripes-core`, `stripes-core` should import the global styles from `stripes-components`.
- Merge contents of `variables.css` in `stripes-core` into `variables.css` in `stripes-components`.
- Install cool Storybook addons like [knobs](https://github.com/storybooks/storybook/tree/master/addons/knobs), [actions](https://github.com/storybooks/storybook/tree/master/addons/actions), and [a11y](https://github.com/storybooks/storybook/tree/master/addons/a11y).
- Write stories for every component.
- Figure out a home for style guide information that FOLIO designers can collaborate on.

## Screenshots
stripes-components-storybook

### DIFF
--- a/.storybook/addons.js
+++ b/.storybook/addons.js
@@ -1,0 +1,2 @@
+import '@storybook/addon-options/register';
+import 'storybook-readme/register';
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,0 +1,14 @@
+import { configure } from '@storybook/react';
+import { setOptions } from '@storybook/addon-options';
+
+setOptions({
+  name: 'FOLIO Stripes',
+});
+
+const req = require.context('../lib', true, /\.stories\.js$/);
+
+function loadStories() {
+  req.keys().forEach((filename) => req(filename));
+}
+
+configure(loadStories, module);
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,5 +1,6 @@
 import { configure } from '@storybook/react';
 import { setOptions } from '@storybook/addon-options';
+import '../lib/global.css';
 
 setOptions({
   name: 'FOLIO Stripes',
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,0 +1,78 @@
+// you can use this file to add your custom webpack plugins, loaders and anything you like.
+// This is just the basic way to add additional webpack configurations.
+// For more information refer the docs: https://storybook.js.org/configurations/custom-webpack-config
+
+// IMPORTANT
+// When you add this file, we won't add the default configurations which is similar
+// to "React Create App". This only has babel loader to load JavaScript.
+const postCssImport = require('postcss-import');
+const autoprefixer = require('autoprefixer');
+const postCssCustomProperties = require('postcss-custom-properties');
+const postCssCalc = require('postcss-calc');
+const postCssNesting = require('postcss-nesting');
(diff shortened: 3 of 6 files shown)


cherewaty:
cc @rasmuswoelk @skomorokh @JohnC-80

JohnC-80:
I'm not quite on the side of being completely without global css rules yet. 'global.css' has been present from the start and thus far it's been fairly conservative - adding mainly utility classes that could be used outside of shared components for general-purpose layout needs. Things don't get added there very frequently -  It only just began to take on some updates as getting rid of the bootstrap dependency became a solidified reality. 
Now that you mention it, developer experience in stripes components is a good subject - are there good facilities within storybook for adding additional documentation for things existing outside of the components themselves? Docs for topics such as any mixins, utility classes, utility functions, etc? This is a good step, but lots of work still ahead of us on this, in any case...

cherewaty:
I had to explore the idea of going without any global CSS, and I wasn't ready to make that leap either. The existing `global.css` was pretty conservative, and if we leave it that way, it seems like a good pattern to continue.

I haven't found a simple drop-in Storybook plug-in for docs (style guide, mixins, utilities, etc.) that aren't tied to a specific component, but that doesn't mean one doesn't exist. We could also roll our own without too much work.

cherewaty:
@JohnC-80 could you merge? (I don't have merge permissions on this repo) 😁

JohnC-80:
Done!
Yeah.. even if something just simply displayed a markdown file in full and set up a link in the side menu, that would be a nice thing that would go a long way.

PR #93, opened 2017-11-17
Set up stylelint

cherewaty:
## Purpose
We'd like to treat our CSS as a first-class citizen with our JavaScript. Part of that means sticking to reasonable, consistent conventions. `stylelint` can give us the same anti-bikeshedding and bug prevention in CSS we already have with `eslint` in JS.

## Approach
`stylelint` will now run with `eslint` as part of `yarn lint`.

Developers can install a `stylelint` plugin in their editor for faster iteration: https://github.com/stylelint/stylelint/blob/master/docs/user-guide/complementary-tools.md#editor-plugins

Several of the errors would never have been checked in if `.editorconfig` was being followed by the editor: http://editorconfig.org/#download

Most editors' beautify tools will format CSS exactly like `stylelint-config-standard` recommends.

I started to make a `stylelint-config-stripes`, so that FOLIO Stripes apps could have a shared config like `eslint-config-stripes`. Unfortunately, `stylelint` relies too heavily on absolute paths for a shared config to be set up in the same way as `eslint`. The `stylelint-config-standard` ruleset is much simpler than the `eslint-config-airbnb` ruleset, so we should be ok just relying on `stylelint-config-standard` in all FOLIO Stripes apps. When `stylelint` gets some refactoring to support global shared configs the same way `eslint` does, we can come back and create `stylelint-config-stripes`.

### Frequent errors
A sampling of the most common rules found in `stylelint-config-standard` that threw errors or warnings in `stripes-components`:
- `indentation`
- `no-missing-end-of-source-newline`
- `number-leading-zero`
- `block-opening-brace-space-before`
- `rule-empty-line-before`
- `max-empty-lines`
- `no-eol-whitespace`
- `selector-list-comma-newline-after`

## Learning
[Stylelint](https://stylelint.io)
[Full list of out-of-the-box stylelint rules](https://github.com/stylelint/stylelint/blob/eec8dd4b179f2b4c0b3b55415a42189a901c01b0/lib/rules/index.js)

## Screenshot
Sample `stylelint` output:

sample stylelint output


cherewaty:
Related `stripes-core` PR: https://github.com/folio-org/stripes-core/pull/106

JohnC-80:
I like what I see so far with stylelint.... a few other css linters out there just barfed over the 'css-next-y' syntax. Another bullet to bite, but one more step to production-readiness.

cherewaty:
@JohnC-80 this should be mergeable

JohnC-80:
Good deal. Merged.

PR #94, opened 2017-11-20
Create Storybook stories for Modal and TextField

cherewaty:
Sets up Storybook stories for variants of `Modal` and `TextField`.

localhost_9001__selectedkind modal selectedstory dismissible 2c 20with 20label full 0 down 1 left 1 panelright 0 downpanel react_storybook 2freadme 2fpanel ipad

### DIFF
--- a/lib/TextField/TextField.stories.js
+++ b/lib/TextField/TextField.stories.js
@@ -1,0 +1,56 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react'; // eslint-disable-line import/no-extraneous-dependencies
+import TextField from './TextField';
+
+storiesOf('TextField', module)
+  .add('with defaults', () => (
+    <TextField />
+  ))
+  .add('with type', () => (
+    <TextField type="email" />
+  ))
+  .add('no border input style', () => (
+    <TextField inputStyle="noBorder" />
+  ))
+  .add('rounded input style', () => (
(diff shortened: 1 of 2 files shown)


lib/TextField/TextField.stories.js
```diff
@@ -0,0 +1,56 @@
...
+
+storiesOf('TextField', module)
+  .add('with defaults', () => (
+    <TextField />
+  ))
+  .add('with type', () => (
+    <TextField type="email" />
+  ))
+  .add('no border input style', () => (
+    <TextField inputStyle="noBorder" />
```
cherewaty: Looks like a couple different ways to get a non-bordered or rounded input? Probably needs a de-duping, but I didn't want to tackle any refactoring in this PR, just give us better tools to refactor when we need to :)

JohnC-80:
What's more the more useful path for these - including lots of different examples for each and every prop that a component has, or including an add-on like knobs that would allow viewers to tweak props themselves - given appropriate descriptions of each prop - hopefully with a place the resulting code could be copied for pasting into their app?

cherewaty:
Short answer: both.

Long answer: Knobs will be fun, and provide a better documentation experience. Writing lots of examples of the variants is probably better for future visual regression testing purposes. Maybe I'll throw a couple knobs in the next time I do a story, but I think we should continue writing out lots of explicit examples of variants.

PR #97, opened 2017-11-21
Add react-tether to Datepicker. Fixes STCOM-125.

mkuklis:


mkuklis:
@JohnC-80 I added `TetherComponent` to `Datepicker`. This is a very similar setup to what we have in `Dropdown`. Things seem to work and the Datepicker shows above the accordion now but could you please take a quick look at those changes. Thank you!

zburke:
@cowboyd Yes, we can run lint as a check rather than a comment, but we don't want to because then we'll just have code littered with `// eslint-disable-next-line`. Our goal is to use lint to encourage good style, not to require unyielding obedience.

cowboyd:
> Yes, we can run lint as a check rather than a comment, but we don't want to because then we'll just have code littered with // eslint-disable-next-line

Or just fix the error? 

The code is already [littered with `//eslint-disable-next-line`](https://github.com/search?q=org%3Afolio-org+%2F%2Feslint-disable-next-line&type=Code). It looks to me like being relaxed  about linting has not prevented this. It has actually encouraged it.

> Our goal is to use lint to encourage good style, not to require unyielding obedience.

In my experience the best way to achieve this is to make good style the default while allowing explicit overrides; not the other way round.

MikeTaylor:
"The error" is not always an error. And sometimes it _is_ an error, but in the judgement of the developer best left visible to return to subsequently, rather than masked out with an `eslint-disable` comment.

Anyway, this is how we do it. If you want to make ESLint-compliant a hard requirement for committing to your own repos, you're welcome, of course.

MikeTaylor:
BTW., @cowboyd, I'd appreciate your thoughts on erroneous ESLint errors like the one at https://github.com/folio-org/ui-users/blob/UIU-303/Users.js#L56 which is reported as "'onSelectRow' PropType is defined but prop is never used" (`react/no-unused-prop-types`) even though it clearly _is_ used at https://github.com/folio-org/ui-users/blob/UIU-303/Users.js#L221

I guess you guys must have run across this kind of mistaken inference a ton. How do you address it? Sometimes I've explicitly used `this.props.foo` rather than `props.foo` even though `props` is set to `this.props`, just to shut lint up. But that seems wrongheaded.

MikeTaylor:
(At least this problem [is documented](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md#false-positives-intermediate-variables) -- but I'd rather it were fixed!)

MikeTaylor:
([This comment](https://github.com/yannickcr/eslint-plugin-react/issues/816#issuecomment-317211331) seems to say "We're not going to fix this.")

cowboyd:
@MikeTaylor My instinct is always to disable rules that cause more headaches than the errors that they catch. It feels to me like this is a rule that takes a lot of effort to make smart enough to actually understand what is happening. As it is, it just doesn't understand (and yet feels entitled to have a breaking opinion on) a good portion of perfectly fine code.

This particular rule smells a lot to me like trying to implement a static type checker using eslint which isn't really designed to do that. It turns out that implementing a full type inference engine is non-trivial and so this rule half-asses it instead, which feels like a bad compromise.

My experience is that `PropTypes` does a really good job of diagnosing problems in the console both in tests and in development, so problems with bad PropTypes rarely get checked in. 

I guess the question is, has anybody experienced accidentally checking in bad uses of properties?

cherewaty:
Issue filed re: build check https://issues.folio.org/browse/STCOM-135

JohnC-80:
@cowboyd "PropTypes being a thing in the first place" might be the response from some... but I'm for them, myself. The main issue at hand is the janitorial process of linting, to enforce or not...
Is build failure the real answer to this? Shouldn't build failures be used for more legitimate reasons? Or will we see Build Failures also become cheapened like code reviews were at one point (a saddening thing.)
Having the lint results in comments at least allows for some easy inspection - thus, it's a good compromise. It's up to everyone involved to pay attention to the linting comments - and admittedly it was my fault for merging this white-space issue- lesson learned.
The false positives that are present in rules such as `no-unused-proptypes` make it certain that these things should not be heavily depended on. The fact that it doesn't seem to really protect you from any real show-stoppers makes it even worse. **I'm for disabling this rule.**

skomorokh:
> "PropTypes being a thing in the first place" might be the response from some.

That's me! I think they're a fine thing for library components that other people are going to be using (like here in stripes-components), but private components within an application, especially inline, ehh, it's a bit much. 

The `no-unused-proptypes` is helpful in that, for those cases where we do want to maintain proptypes, we don't accumulate baggage over time from props we've stopped supporting but neglected to remove from the proptypes. I think if it wasn't so broken it might be worth keeping around as a warning rather than an error, a solution I offered to STCOM-135: set it up to stop commits with lint we're definitely not okay with but still optionally provide some more verbose commentary as warnings. But it does sound pretty broken so it's probably just annoying.

MikeTaylor:
Agreed entirely. PropTypes are a machine-readable description of (part of) the API of a component. They are genuinely useful in library components that are going to be used by other code, especially when the user is a different coder from the person who made the library component. But when one person is building an application of several components, writing PropTypes for those purely internal APIs feels like busy-work to me.

With that said, when we do use PropTypes at all, I favour retaining `no-unused-proptypes` despite its well-documented imperfections. It's irritating, but I think the value outweighs the annoyance.

PR #99, opened 2017-11-22
Add button styling and story

rasmuswoelk:
This is my first pull request to stripes-components - feel free to mention if I should be doing anything differently.

You won't see a lot of commits because I initially just cloned the original repo instead of forking it so I just copied my files after forking.

**Changes**
- Added story to button (I have experimented with some additional sub folders to keep the component folder as clean as possible and to make the documentation more manageable)
- Updated some button styling (focus, hover and active styling etc.)
- Added two new styles (warning and success)
- Added mixin for easily changing or adding button styles
- Added two new post css plugins (css-rucksack and postcss-mixins). Important: These needs to be added to stripes-core.

Let me know if you need any more info :)


rasmuswoelk:
Updated the pull request with the following:

- Removed css-rucksack and postcss-mixins after was made aware that we need to follow CSS4 features as much as possible
- Fixed all linting bugs

cherewaty:
@rasmuswoelk you should click that link to sign the CLA, then it won't bug you about that on future PRs.

cherewaty:
https://github.com/samhaengcom/stripes-components/blob/f793d86fe9660a7d70d40f6011ad5ed128fa1015/lib/Button/Button.css#L432 expands the height of the buttons at smaller sizes, which can cause situations like this:

screen shot 2017-11-23 at 10 11 28 am

Is that the behavior we want?

rasmuswoelk:
@cherewaty 

It's probably not the exact behaviour we want however clickable elements should be at least 42x42px on mobile. The design will probably also change on mobile once we get to it. For now we could remove that breakpoint.

Note: The breakpoint and the change of height was there when I started looking at the component (as well as a lot of other classes and functionality) so I didn't wan't to remove it. We should probably re-visit the styling and the props of the component, remove unnecessary styling/logic and then fix it wherever the changes might break something. E.g. The component has two props for removing margin: marginBottom0 and bottomMargin0.

Regarding the CLA: I have clicked on the link and agreed several times. I am not sure why it won't update or save.

lib/Button/Button.css
```diff
@@ -44,41 +47,66 @@ a.button {
...
+  /**
+  * Button Style: Primary
+  */
+
   &.primary {
     background-color: var(--primary);
     border: 1px solid var(--primary);
     color: #fff;
 
-    &:hover {
```
JohnC-80: Removing the hover styles - any particular reason? Intended? Just curious.

PR #100, opened 2017-11-23
Added story and updated styling for MultiColumnList

rasmuswoelk:
I have added a story and changed styling according to [ux.folio.org/universal-navigation/uxi-3](ux.folio.org/universal-navigation/uxi-3) on the MultiColumnList component.

Among other changes:
- Padding and text styles
- Background colors on even rows
- Better hover, focus and active styles

**Note!**
I merged my forked master with my other branch (add-button-styling-and-story) so there are some files here related to the Button-component which is getting merged in another pull request. In the future I won't merge my branches into my own master before they are merged in to the upstream-repo. 

**New addons:**
I added two addons for storybook which I'm not using in this case but they can be useful in other stories and I will probably make use of these in the future. 

[storybook-actions](https://github.com/storybooks/storybook/tree/master/addons/actions) (Detect and writes out onClick actions) and [storybook-knobs](https://github.com/storybooks/storybook/tree/master/addons/knobs) (Enables some options for adding toggles etc. E.g. Easily toggle different styles or modals etc.)


lib/MultiColumnList/MCLRenderer.css
```diff
@@ -30,40 +30,61 @@
...
-    &.selected {
-      color: #fff;
-      background-color: color(var(--primary) shade(8%));
-
-      &.cell {
-        color: #fff;
-        background-color: color(var(--primary) shade(8%));
-      }
-    }
+  /* Active style (we have to be a bit specific to avoid using !important ) */
```
cherewaty: Thanks for staying away from `!important` :)

lib/MultiColumnList/stories/basic.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
cherewaty: I like the architecture pattern you're using with the stories broken up into several pieces.

rasmuswoelk:
Linting errors fixed.

PR #104, opened 2017-11-27
Adjust devDependency linting for .stories.js files

cherewaty:
## Purpose
Storybook is currently a `devDependency` of `stripes-components`, since consuming applications don't need to pull in Storybook.

The `import/no-extraneous-dependencies` ESLint rule was breaking on `.stories.js` files that need to import pieces of Storybook, because Storybook is simply a `devDependency`. We were disabling the rule on lines that violated it, but that won't scale well.

## Approach
This `.eslintrc` change will affect files that end in `.stories.js`:
```
"import/no-extraneous-dependencies": ["error", {
  "devDependencies": ["**/*.stories.js"]
}]
```
Any file that matches that pattern will not complain about dependencies anymore. All other files will continue to lint the same as before.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,5 +8,10 @@
     "mount":true,
   },
   "extends": "stripes",
-  "parser": "babel-eslint"
+  "parser": "babel-eslint",
+  "rules": {
+    "import/no-extraneous-dependencies": ["error", {
+      "devDependencies": ["**/*.stories.js"]
+    }]
+  }
 }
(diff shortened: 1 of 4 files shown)


cherewaty:
@rasmuswoelk  this means you shouldn't have to do `// eslint-disable-line import/no-extraneous-dependencies` anymore on stories.

MikeTaylor:
LGTM

PR #108, opened 2017-11-29
Updated icon component + story + global.css changes

rasmuswoelk:
In this pull request I have primarily worked on the Icon-component but I have also added some changes to global.css and a couple of other components.

nov-29-2017 14-04-16

**Icon-component**
- Removed inline sizing of icons and added size prop instead with sizes: small, medium and large (medium is default). The sizes of the icons will probably need to be adjusted.
- Removed styling related to buttons from icon. This will be controlled via. the button component (update coming)
- Removed center-prop since this should be controlled by the parent components instead
- Moved icons out of the component itself and into a separate file called icons.js. This way we have a single source which also is used to list the icons in the story. Easier to add, remove and edit icons. And we have a better overview of available icons in the story.
- Added story with knobs. You can change color and size of icons.

Note: The updates of the icon component might bring some breaking changes here and there. We will have to fix those as we stumble upon them. The size of the icon is default 18x18px "medium" which was the default size before.

**Globals.css**
- Added font inherit to input, buttons etc. since they down inherit as default
- Added reset for CSS. Note: This can (and probably will) mess up some of the styling here and there but it's important to have it from the start. It helps us normalize styling for different browsers. We can fix styling issues as we go.

**Misc.**
I have updated a couple of other components that needed to be updated as a result of changing globals.css and the Icon-component changes etc.


lib/Icon/stories/BasicUsage.js
```diff
@@ -0,0 +1,34 @@
...
+          icon={icon}
+          color={colour}
+        />
+      </div>
+      <p>{icon}</p>
+    </li>
+  ));
+
+  return (
+    <div style={{ padding: '15px' }}>
```
cherewaty: There are a few alternative strategies here so you won't have to repeat this inline style for the Storybook renderer: https://storybook.js.org/addons/introduction/

lib/Icon/Icon.css
```diff
@@ -9,24 +9,50 @@
   }
 }
 
+.root svg {
+  height: inherit;
+  width: inherit;
+}
+
 :global .stripes__icon {
```
cherewaty: Is there a way around this being `:global`?

lib/Icon/icons.js
```diff
@@ -0,0 +1,41 @@
...
+  'down-caret': ({ style, className }) => (<svg xmlns="http://www.w3.org/2000/svg" focusable="false" viewBox="0 0 26 26" style={style} className={className}><polygon points="13 20.6 1.8 7.3 4.1 5.4 13 15.9 21.9 5.4 24.2 7.3 " /></svg>),
+  'up-caret': ({ style, className }) => (<svg xmlns="http://www.w3.org/2000/svg" focusable="false" viewBox="0 0 26 26" style={style} className={className}><polygon points="13 5.4 24.2 18.7 21.9 20.6 13 10.1 4.1 20.6 1.8 18.7 " /></svg>),
+  'down-triangle': ({ style, className }) => (<svg xmlns="http://www.w3.org/2000/svg" focusable="false" viewBox="0 0 26 26" style={style} className={className}><polyline points="16.8 6.1 9 15.3 1.2 6.1 " /></svg>),
+  'up-triangle': ({ style, className }) => (<svg xmlns="http://www.w3.org/2000/svg" focusable="false" viewBox="0 0 26 26" style={style} className={className}><polyline points="1.2 13.9 9 4.7 16.8 13.9 " /></svg>),
+  'spinner-ellipsis': () => (<div className={dotSpinner.spinner}>
+    <div className={dotSpinner.bounce1} />
+    <div className={dotSpinner.bounce2} />
+    <div className={dotSpinner.bounce3} />
+  </div>),
+};
```
cherewaty: I've had a lot of success with SVG Store before: https://github.com/svgstore/webpack-svgstore-plugin (although I've only used it with broccoli, not webpack). It cleverly merges all your SVG files into a single sprite file. This current strategy is ok for now, but I'd rather serve up SVGs outside the `stripes` JS bundle.

JohnC-80: Will that work for inline SVGs? That's what we have right now since aspects like fill can be controlled with style.

rasmuswoelk: I have worked with sprites before. You can change color in css by "color" instead of fill. Only issue can be a slight "blink" in icons on load/refresh. I'm not sure if this can be prevented? Other than that it's a nice way to handle icons. It can always be implemented later on.

lib/Icon/Icon.css
rasmuswoelk: Hmm. I think it's a good idea to have a global class for this one. E.g. in buttons I wan't to be able to target the icons for matching colors on different button styles. My idea was to use this class for that.

lib/Icon/stories/BasicUsage.js
rasmuswoelk: Nice. I'll take a look at that :)

lib/Icon/icons.js
rasmuswoelk: The cool thing about being able to use color in CSS to change icons is that they automatically will inherit their parents color value.

PR #125, opened 2017-12-12
Pane header changes

rasmuswoelk:
**Note:** The new icons in #123 is merged into this pull request. 

The changes in this pull request is made to be backward compatible with the current PaneHeaders in various apps but they will look a bit broken. I will submit a pull request for the ui-users / smart-components in order to make them look nice again (by using the IconButton for icons in PaneHeaders etc.).

**Changes**
- Largely rewritten default functionality
- Added paneSub-prop (so you can add it directly instead of adding them via. the paneTitle-prop)
- Added IconButton-component which is used in the PaneHeader
- Added story for PaneHeader (located under Pane in StoryBook - I thought it made sense to keep everything related to Panes here even though the components has their own folders)
- It automatically adjusts margin so that the centered part is always 100% centered (currently it will be off center if there are different widths content in lastMenu and rightMenu)
- Added ellipsis to both paneTitle and paneSub
- Added new close icon for close action


lib/PaneHeader/PaneHeader.js
```diff
@@ -1,110 +1,179 @@
...
+    this.setCenteredMargin();
+  }
+
+  /**
+   * Set horizontal margin on centered content area
+   * The centered content is absolute to make sure that the content
+   * is always centered in the header - no matter what content there is to the left and right.
+   * We add margin so that the centered content won't overflow the menu areas.
+   */
+  setCenteredMargin() {
```
cherewaty: Is there a way to do this centering with purely CSS and no JS? Calculating these things at runtime is likely to be pretty brittle.

lib/IconButton/IconButton.css
```diff
@@ -0,0 +1,25 @@
+/**
+ * Pane Menu Icon
+ */
+
+@import '../variables';
+
+.root {
```
cherewaty: How about `.iconButton` instead of `.root`? Should make debugging CSS easier when the outputted classes aren't all `root-{hash}`.

lib/PaneHeader/PaneHeader.js
rasmuswoelk: It would be better to do it without JS for sure. I tried finding a solution but couldn't find a bulletproof one.

@cherewaty The solution looks good except that the right/left section doesn't expand once the content is wider than the section which causes the buttons to go on two lines.

image

lib/IconButton/IconButton.css
rasmuswoelk: I agree on that. I had just seen this kind og naming in a lot of other components so I suppose I  thought it was some kind of convention you guys was using (which I also found harder to debug). I'll change that root into iconButton.

lib/PaneHeader/PaneHeader.js
cherewaty: @rasmuswoelk since the JS solution is working (even if it's not ideal), you should just proceed with it. It's isolated enough to refactor later to be just CSS.

rasmuswoelk: I agree. Can I merge?

PR #128, opened 2017-12-13
Fix checkbox radio button sizes

rasmuswoelk:
On the request of Filip I have updated the checkboxes and radio buttons styling (made them smaller)

### DIFF
--- a/lib/Checkbox/Checkbox.css
+++ b/lib/Checkbox/Checkbox.css
@@ -1,10 +1,13 @@
 @import '../variables.css';
 
+:root {
+  --checkbox-size: 12px;
+}
+
 .root {
   display: flex;
   align-items: center;
   transition: background-color 0.2s ease;
-  margin-bottom: var(--controlMarginBottom);
   position: relative;
 
   &.rootFullWidth {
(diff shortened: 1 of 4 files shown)


lib/Checkbox/Checkbox.css
```diff
@@ -1,10 +1,13 @@
 @import '../variables.css';
 
+:root {
+  --checkbox-size: 12px;
+}
+
 .root {
```
cherewaty: `.checkbox` instead of `.root`? (same comment as https://github.com/folio-org/stripes-components/pull/125#discussion_r156392872)

lib/RadioButton/RadioButton.css
```diff
@@ -1,10 +1,13 @@
 @import '../variables.css';
 
+:root {
+  --radio-button-size: 12px;
+}
+
 .root {
```
cherewaty: `.radioButton` instead of `.root`?

rasmuswoelk:
@cherewaty CSS classes renamed from .root to .checkbox and .radioButton 👍

PR #145, opened 2017-12-22
expose a ref on TextArea for access by an ancestor. Refs STSMACOM-4

zburke:
In order to implement @mention autocompletion in a `<TextArea>` component, we need to expose a ref on that component so the autocomplete handler can grab the element and correctly position the dropdown menu relative to it.


skomorokh:
Urgh. With redux-form we're constantly putting all form contents in the redux store, we should be able to do better than this.

Perhaps we should expose all the form state via the mapStateToProps in stripes-connect? Later we can restrict it to only the module's forms when we (somehow) partition the form namespace (which we need to do to prevent collisions... which we need to prevent if we want to have some form values persist when you navigate away)

zburke:
The context of this PR is specifically to support access by @mattkrick/react-githubish-mentions, so it's less about getting the field's value than it is about getting the field itself in order to correctly position the dropdown menu. Because the menu is implemented in a portal to prevent clipping, it needs access to the actual element. 

Does that make you feel any better about this change? Or worse? If worse, do you have a suggestion about how we might accomplish STSMACOM-4?

skomorokh:
Ya, definitely a lot better--I was indeed reacting to the "grab the value" part of your comment. Though I don't know we currently have a better solution for that without a ref either so we probably should still investigate exposing the form data. Either way, I also should have been clear that I meant that as "eventually" and not to prevent this PR; a comment in the context of this PR not in opposition to it.

I shudder at the notion of element positioning because I cannot stomach the compromises required, do what you gotta do. Personally I would constrain my layout choices to those which flexbox cleanly enables and which we can perform without ever needing to manipulate the DOM from JS (eg. no refs ever), but the result would probably be "this looks like it was designed by a programmer" :)

PR #148, opened 2017-12-28
Datepickers should only show a single clear icon

cherewaty:
Errored datepicker fields were showing two clear icons.

### Before
screen shot 2017-12-27 at 5 04 00 pm

### After
screen shot 2017-12-27 at 5 06 20 pm


cherewaty:
cc @elrickvm

PR #154, opened 2018-01-01
Adding AppIcon component

rasmuswoelk:
This component is work-in-progress. I am waiting for the icon functionality to be worked out.

Read more in the discussion here: https://issues.folio.org/browse/STCOR-117

Until this gets worked out the AppIcon will be used as a placeholder for app icons in a future update of the main header of the application. Once the app icon functionality is in place we can update it accordingly. Until that it will replace the green, purple and orange icons in the header: 

image

**Some notes:**
- This component is both for the default app icon and the popover icon. A prop determines which styles is used.
- The AppIcon-component comes with a iconSize-prop for changing the size of the icon. This is because the icon will be used in different sizes all over the application in the future. Below is screenshots from prototypes. Also, you can see the icon implemented in the UXI-prototype here: ux.folio.org/universal-navigation/uxi-3/

**Examples of usage in the future**
image

image

image


rasmuswoelk:
I think I will remove the popover-icon feature from this component to keep it more simple (and also because it's not really needed with the upcoming changes I'm submitting for the MainNav). 

Hold on with the merge :)

PR #165, opened 2018-01-03
Give explicit class names instead of .root

cherewaty:
When debugging markup and writing tests for the browser, it'd be more helpful to have explicit class names for components, instead of `.root--${MODULE_HASH}`.

Many `stripes-components` already follow this pattern.


JohnC-80:
Thanks! LGTM

PR #169, opened 2018-01-05
STCOM-175 actually disable filters

JohnC-80:
Functionally disable the filter checkboxes in a disabled filter group.


JohnC-80:
Brutal, jenkins, brutal... alright, linting.....

PR #170, opened 2018-01-06
Establish global primary link color

cherewaty:
### Purpose
Since we have a place for a limited set of global styles, the color of `<a>`s makes sense to define on the global level.

### Questions
- Should there be a different hover state? Background? Underline? Darken the primary color? Lighten the primary color?
- Should there be a `:visited` state? My inclination is that coloring visited links purple (or any other type of indication it's been visited) is more detrimental than useful in an application like FOLIO. Other opinions welcome.

### DIFF
--- a/lib/global.css
+++ b/lib/global.css
@@ -1,4 +1,5 @@
 @import "reset";
+@import "variables";
 
 * {
   box-sizing: border-box;
@@ -40,7 +41,12 @@ html {
 }
 
 a {
+  color: var(--primary);
   text-decoration: none;
+
+  &:visited {
+    color: var(--primary);
+  }
 }
 
 strong {


rasmuswoelk:
@filipjakobsen What do you think?

filipjakobsen:
There will be a global variable for the link color. By default it will derive the main interaction color. The “visited” link color I think is useful, and this will also be derived from the main interaction color, but with a tweaked hue.

Links in text should have a border-bottom, and on hover, the underline should disappear and the text become less opaque (effectively lightening or darkening it depending on background color)

If text links open a new window with a non-FOLIO site, they should be followed by an icon indicating this.

The focus style will be in place when we implement the general focus styles which we have been designing.

filipjakobsen:
Jeffrey, why do you think a visual :visited indicator would be detrimental?

filipjakobsen:
@cherawaty

filipjakobsen:
@cherewaty

cherewaty:
@filipjakobsen link styles you mention sound excellent 👍 

My thought on `:visited` was that FOLIO is an application used all day, every day. Seeing that I visited a link a few days ago may actually be confusing when I've completely switched the context of the holding or user I've been looking for. I have no evidence to back this idea up.

filipjakobsen:
@cherewaty That's a relevant point to consider. I would say we should follow the web convention to begin with — having the :visited style there. And at the same time be very aware of the challenge you mention, what impact it has and if there is a distinction between external and internal links, in-app and cross-app links, and in general how it feels once we try it out.

How does that sound to you?

filipjakobsen:
@cherewaty discussed this in our weekly call and agreed my proposed suggestion above makes sense. Correct me if I misunderstood that, @cherewaty 🙂

cherewaty:
I'll make a follow-up pull request for this to define a `:visited` color in `global.css`. We'll stick with the web convention for now (always a good bet).

cherewaty:
Follow-up: https://github.com/folio-org/stripes-components/pull/177

PR #174, opened 2018-01-09
Added SearchField-component

rasmuswoelk:
The SearchField component is an universal component that will be used in different places across the application to prevent duplicate code. It's simple for now but can be updated later on with more functionality if needed.

**Changes:**
- Added SearchField component and story under "structures"
- Updated Icon component with id prop. Made it a stateless component.
- Added new props to TextField (ariaLabel) and TextFieldIcon (id)

### DIFF
--- a/lib/structures/SearchField/SearchField.js
+++ b/lib/structures/SearchField/SearchField.js
@@ -1,5 +1,5 @@
 /**
- * SearchFieldUsage
+ * SearchField
  *
  * A universal search field component
  */
--- a/lib/structures/SearchField/SearchField.js
+++ b/lib/structures/SearchField/SearchField.js
@@ -23,6 +23,7 @@ const propTypes = {
   clearSearchId: PropTypes.string,
   onClear: PropTypes.func,
   selectedIndex: PropTypes.string,
+  searchableIndexesPlaceholder: PropTypes.string,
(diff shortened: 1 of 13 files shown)


lib/structures/SearchField/SearchField.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * SearchFieldUsage
```
doytch: Super minor but is this comment supposed to just be `SearchField` for autodocs?

doytch:
@MikeTaylor the Codex Search looked solid in the demo today! One thing: it looks like you're using the `<select>` functionality of `FilterPaneSearch` to do the selection of what the search field is searching (title, ISBN, etc). 

So @rasmuswoelk , I think we'll need that same functionality in this SearchField component as well if we want to move it off of the header as per [STCOM-164 #1](https://issues.folio.org/browse/STCOM-164)

rasmuswoelk:
@doytch I haven't seen the demo yet - I will check it out and get back to this 👍

rasmuswoelk:
@doytch I have updated the component to support searchable indexes. Can you check it out and let me know if it supports what you need?

doytch:
Slick! Looks good man!

PR #175, opened 2018-01-09
Replace reset.css with normalize.css

cherewaty:
## Purpose
We've now had a little over a month with the 2011 edition of [Eric Meyer's reset CSS](https://meyerweb.com/eric/tools/css/reset/) as part of the `stripes` CSS (https://github.com/folio-org/stripes-components/pull/108). I've been running into cases where I want to write some basic body copy or a headline-paragraph combo, and I have to write all new typographic styles (esp. margin and line-height) for those pieces. This does not encourage consistency, and makes for a much heavier CSS bundle. Ideally, I should be able to write any vanilla HTML and have it receive good defaults.

## Approach
[Normalize.css](http://nicolasgallagher.com/about-normalize-css/) is an alternative to a reset stylesheet. It leaves typographic defaults while still fixing inconsistencies and browser-specific bugs. It's the industry standard in CSS resetting/normalization, used by Bootstrap, HTML 5 Boilerplate, GitHub, Gov.UK, etc. It's an actively-maintained, healthy open source project.

## Risk
There should be very low risk of broken functionality with this change. However, there is almost 100% chance of visual regressions, particularly when it comes to headlines. The sooner we can make this change (while other CSS is being written), the less painful it will be.

I've attached commits to make line-heights more consistent throughout the CSS and fix the PaneHeader titling that now needs its margins zero'ed out.

## Alternatives
- Stick with the `reset.css` we've got, but add a lot more re: margin, padding, line height, etc. to `global.css`.
- Completely remove any CSS resetting/normalization code (an approach I'm seeing come into vogue, but that might be too far of a leap).

## Questions
- 1.15, the `normalize.css` line-height, is still pretty conservative. We might want to consider 1.2 (or even higher), but a consistent 1.15 across the board is better than a base of 1 and inconsistent declarations on top of that.

## Screenshots
From macOS Chrome. Other browser/OS combinations could see slight variations in the first screenshot.

#### With no reset.css
no-reset-at-all

#### With reset.css (current)
current

#### With normalize.css instead (after this PR)
normalize


cherewaty:
Definitely want a 👍 from @rasmuswoelk before merging.

JohnC-80:
I don't have any strong opinion between out-of-box resets - there's always some overriding you have to do to achieve what you want. Browser consistency is an ongoing fight that's rarely worth the measures some systems take only to throw their hands up and go native on future revisions. Indeed, it would be nice to have basic html behave expectedly. We've previously had some complaint about `<ul>`s not being bulleted by default, so @cherewaty, you're not alone.

rasmuswoelk:
I see the complications of Eric Meyers reset. Me and Filip decided on using this reset. I don't have any problems with switching it to normalize.css.

@filipjakobsen ?

filipjakobsen:
What happens when (not if) we want to support the ability of the end user to change font-size, font-weight, letter-spacing and line-height universally? How do we control non-component html elements with Normalize? Global styling? If so, why is it better than resetting? I thought we wanted to discourage the use of regular, non-component elements? How does this play along with that philosophy?

These are not rethorical questions - I am just thinking out loud. 

Those questions remain with me. In addition, I recommend that the default line-height be 1.3 rather than 1.15.

filipjakobsen:
(I have no objection to Normalize in itself. I just want to make sure we don’t overlook the things I mention above, and that we don’t change it if we in reality still have to style everything from scratch to control it with theming)

rasmuswoelk:
I agree with John regarding the fact that no matter what reset you choose you will have to adjust things. In that regard it doesn't matter to me what reset we use.

I will also agree on Filip's points. If we take the headers (h1-h6) as an example. I would rather try to encourage 3rd party developers to use our Headline-component than using plain h1-h6 tags. 

This can be achieved by having a thorough design guide / documentation that demonstrates why developers should use our components and not having base styling on header tags. Having semi-nice looking base styling as default on HTML elements will probably not encourage developers to follow our component conventions. Or is it just me?

By using our components 3rd party developers has to do less work of styling and we are more in control.

JohnC-80:
If you let raw HTML go through looking not-so-great, you're going to see modules with pieces that look not-so-great because, despite the nice things components provide, there's always a likely need for basic HTML and basic body copy. Both should look acceptable in the end. `<Headline>` is a good component for what it provides in styling options, as styling headers can be a pain, but it's always only there as a suggestion over an h-tag - and there should be no problem if a developer uses one or the other. tag styles, as in the reset, are always considered global by component-scoped css. They still affect everything they're declared for. Originally, global.css was the home for styles of this nature - and could swimmingly include css variables in those style definitions so that the effect of the variables could carry right through.

filipjakobsen:
I trust you guys with the technical decision about native elements over components. I just thought that with components, we'd want to provide our own set of components, like Headline 1-6 components, and a Paragraph component, etc. But I guess that's not what we're doing then. In which case, I don't have any opinion on Reset vs Normalize and this change is fine by me.

This will force us to have good styles in place for all the native elements, which will require the same type of work as making our own components, so I am not sure I see the benefit, but you guys are the experts on this. So whatever you decide, I'm on board.

cherewaty:
Thanks all for the thoughts. Sounds like we have enough of a consensus to merge.

Follow-up items:
- Make the standard line-height 1.3, not 1.15.
- Keep in mind that developers working with `stripes` applications are not prevented in any way from using raw HTML elements vs. provided components.
- Explore how to communicate when it's appropriate to use a raw HTML element vs. a provided component.

PR #187, opened 2018-01-16
Use min-width media queries instead of max-width

cherewaty:
## Purpose
Justification added to `CSSinStripes.md`:

`stripes-components` provides [pre-set breakpoints](../lib/variables.css) that you can use to adjust styles at different browser window widths.
```
@media (--mediumUp) {
  .button {
    height: 2em;
  }
}
```

Use `min-width` breakpoints like `smallUp`, `mediumUp`, and `largeUp` when possible. There are `max-width` breakpoints available, but the cases for their use should be rare. Leaning on `min-width` promotes a mobile-first CSS workflow, where the narrowest screen size is the default. Layout can gradually be added as the width increases with breakpoints like `--mediumUp` and `--largeUp`.

This:
```
x ≥ 321px, x ≥ 641px, x ≥ 1025px, x ≥ 1441px
```
is easier to debug and reason about than this:
```
x ≤ 640px, 641px ≤ x ≤ 1024px, 1025px ≤ x < 1440px, x ≥ 1441px
```

## Approach
Styles that previously used breakpoints like `--small` and `--medium` have been flipped. Now whatever was the smallest screen width style is the default, and styles get modified as the screen size increases.

## Risk
There is a slight chance of visual regressions, but they should be very easy to fix if they come up.

## Follow-up
The majority of the media queries in `stripes-core` have to do with the `MainNav`, so I'll wait until https://github.com/folio-org/stripes-core/pull/152 is merged to go apply this same pattern in `stripes-core`.

### DIFF
--- a/lib/Paneset/Paneset.css
+++ b/lib/Paneset/Paneset.css
@@ -3,8 +3,8 @@
 .paneset {
   width: 100%;
   height: 100%;
-  position: absolute;
-  display: flex;
+  position: relative;
+  display: block;
   flex-direction: row;
   align-items: stretch;
   justify-content: flex-start;
@@ -21,7 +21,9 @@
   }
 }
 
-@media (--small) {
-  position: relative;
-  display: block;
+@media (--mediumUp) {
+  .paneset {
+    display: flex;
+    position: absolute;
+  }
 }
--- a/docs/CSSinStripes.md
+++ b/docs/CSSinStripes.md
@@ -18,8 +18,8 @@ We also make use of [PostCSS](https://github.com/postcss/postcss) and a few of i
 CSS files can be written as normal. For class-naming, camelCase is recommended, but not enforced. These files should live in the same directory as the component's .js file.
 ```
 /* ComponentName.css */
-.className{
-    color: #090;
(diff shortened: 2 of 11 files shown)


lib/Paneset/Paneset.css
```diff
@@ -21,7 +21,9 @@
   }
 }
 
-@media (--small) {
-  position: relative;
```
cherewaty: This appears to have been a bug. These CSS attributes were not previously modifying anything.

JohnC-80:
@filipjakobsen Do we need to modify these breakpoints further? Previously we had discussed setting a small breakpoint at 800px.

filipjakobsen:
@JohnC-80 

As long as we can easliy adapt these later universally, I think anything is fine.

In the prototype we did on http://ux.folio.org/universal-navigation/uxi-3/ , 800 and 1200 were suitable. But again, if we can just adjust recurringly, I'm fine with any breakpoints at this stage.

One thing that I recommend is that we have an additional breakpoint for very large screens, perhaps around 1600+. I think that would be quite useful.

PR #191, opened 2018-01-17
Use `<MultiColumnList>` in `<EditableList>`

JohnC-80:
Fixes STCOM-190
EditableList grows a few new features with this PR:

- takes advantage of `<MultiColumnList>`'s `formatter` API.
- some columns can be set to 'read-only'
- column headers
- makes use of Promises so that the display need not be prematurely updated (STCOM-151)
- additional error-handling for Promises.
- tracks editing status via internal array, rather than assigning a temporary key to the data objects.

Pre-req for https://github.com/folio-org/ui-users/pull/218.


lib/structures/EditableList/EditableListForm.js
```diff
@@ -64,86 +66,269 @@ const propTypes = {
...
+    this.getVisibleColumns = this.getVisibleColumns.bind(this);
+    this.getReadOnlyColumns = this.getReadOnlyColumns.bind(this);
+
+    this.testingId = this.props.label.replace(/\s/g, '').toLowerCase();
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (!isEqual(this.props.initialValues, nextProps.initialValues)) {
+      this.setState({
+        status: nextProps.initialValues.items.map(() => ({ editing: false, error: false })),
```
mkuklis: @JohnC-80 it looks like the same thing is already happing in `buildStatusArray`. I wonder if we could reuse `buildStatusArray` here and just pass `nextProps` to it. That way you could also reuse it in line 100.

JohnC-80: Great suggestion! Will roll these up into a single function.

PR #192, opened 2018-01-17
Re-adding button margin bottom

rasmuswoelk:
I'm re-adding the default margin on buttons since it's default behaviour and a pattern for a lot of other components.

Since it was there before the prop is already added to a lot of buttons across the system so the change shouldn't cause a lot of problems.

I found some buttons in some places that needed the prop and added them in these pull requests:
- https://github.com/folio-org/stripes-smart-components/pull/78
- https://github.com/folio-org/ui-users/pull/219

They should be merged in before this gets merged in. 

I will update if I come across any other places that needs it. Let me know if you spot any buttons with bottom margin that shouldn't have it and I'll fix it - thanks!


cherewaty:
There are definitely schools of thought that believe atomic components like buttons shouldn't declare any margins, and I tend to agree with them in this case ([Salesforce Lightning](https://www.lightningdesignsystem.com/components/buttons). [Bootstrap](http://getbootstrap.com/docs/4.0/components/buttons/)).

I'm all for margin declarations about relationships like `button + button`, but I'd rather all atomic components have `margin: 0`. _(Atomic in this case means not being composed of smaller components, a la [Atomic Design](http://bradfrost.com/blog/post/atomic-web-design/).)_ When composing atoms together into molecules or organisms, that's the point where we want to add margin. Margin is a function of relationships.

Could we give the `<Button>` component `margin: 0`, remove the `marginBottom0` prop, and rely on more complex components to define margin?

lib/Button/Button.css
```diff
@@ -19,6 +19,7 @@
   /* margin-bottom: var(--controlMarginBottom); */
   margin-left: 2px;
   margin-right: 2px;
+  margin-bottom: 15px;
```
JohnC-80: Why wouldn't this come from a variable along with all other form components so they could all have uniform spacing?

JohnC-80:
I'm actually in the [ZURB foundation](https://foundation.zurb.com/sites/docs/button.html) camp on this one. I prefer no margin-bottom as the override, rather than the default. Forms are the primary use-case that `<Button>` serves.  It's ideal to be able to drop it into a form and have it behave/space itself nicely with rows of inputs or other content below it. Prior to recent updates to the nasty instance form in `ui-inventory`, we saw what a default `margin-bottom: 0` on <Button> can do for layout - a glorious column of bled-together stacked, flat buttons (I'll try and dig up an image), since the dropped-in `<Buttons>` were not spacing themselves at all by default.
If it were not assigned margin by default, I feel like devs would find themselves having to specify quite often in otherwise simple, common situations, which can lead to some great inconsistency. If it's necessary, complex components can re-assign margins/styles as they need to already via the `buttonClass` prop - but for most instances where I'm building something complex, having no margin does work out best most of the time - that's easily set with the bottomMargin0 prop. Comfortable defaults. Simple overrides. Do we want to always have to specify for a visually comfortable layout? I'm inclined to say 'no.'

cherewaty:
@JohnC-80 I'll defer to you on this one - you've put a lot more thought into this than I have. Thanks for the explanation. You're very right about that `ui-inventory` instance form.

With that in mind, 👍 to these changes.

doytch:
Yeah I was talking about this with Rasmus a while back. While normally I'd be 100% on your side @cherewaty, @JohnC-80 explained why I favour this change here. Component libraries that are used by a bunch of different developers across varied use cases do well to not include opinionated styling. But in Stripes' case, the usage is fairly rigidly scoped and I think it make the DX a bit better by removing the need for a bunch of one-off CSS.

lib/Button/Button.css
rasmuswoelk: True - I'll fix it 👍

rasmuswoelk:
@JohnC-80 While making these changes I came across the fact that there are two classes on the button for defining zero-bottom-margin:

marginBottom0 and bottomMargin0

Doing a global search for these in stripes-components I found most instances of marginBottom0 being used (44 vs. 26 of bottomMargin0 - primarily related to Button's, so in other components marginBottom0 is used).

This must be a result of some re-factoring but can we agree on using marginBottom0 as the standard and try to phase out bottomMargin0?

JohnC-80:
Yes, @rasmuswoelk  `marginBottom0` matches the word-order of the css attribute it implements, so it makes the most sense to use. Let's go move forward with it.

PR #193, opened 2018-01-18
STCOM-167, STCOM-183 Added prop for href that turns IconButton into a link if needed

rasmuswoelk:


JohnC-80:
Awesome, Thanks @rasmuswoelk!

PR #209, opened 2018-01-31
Modify default button style

cherewaty:
## Purpose
The FOLIO button styles available take a very similar approach to [Bootstrap](https://getbootstrap.com/docs/4.0/components/buttons/) and [Foundation](https://foundation.zurb.com/sites/docs/button.html). That means there are a wide variety of styles available, but they are unprescriptive about use.

### Before
before

Weaknesses of the current strategy:
- Primary buttons (in solid blue) are the default if no specific style of button has been specified. That means there are frequently multiple primary buttons visible at the same time.
- Secondary buttons can be confused with a disabled state.
- Secondary buttons have a very similar visual weight to primary buttons. We're solely relying on color to distinguish between them.
- Hollow buttons don't have an explicit purpose.

## Approach
Based on a session at the FOLIO Madrid dev summit and a conversation with @filipjakobsen, here are new naming conventions and styles for buttons. It shifts to a more prescriptive set more similar to [MailChimp's Pattern Library](https://ux.mailchimp.com/patterns/forms) or [Salesforce's Lightning Design System](https://www.lightningdesignsystem.com/components/buttons/).

### After
after

- There's now a `default` button style different from the `primary` button style. When in doubt, use `default`.
- When in the context of a form, use the primary button to denote what will happen if the user submits the form.
- In almost every case, the primary button should only be used if typing Enter/Return will trigger it.
- "Error" has been renamed to "Danger" to better reflect when it gets used. It's intended for buttons where clicking them will cause something "dangerous" to happen.
- `danger`, `success`, and `warning` buttons should be used sparingly.
- `hollow` is no longer an available button prop.
- `transparent` and `secondary` are no longer available `buttonStyles`.
- The disabled state is now distinct from every enabled button style.

### Disabled buttons
disabled

## Risk
This _is_ a breaking change. A few quick searches around the `folio-org` revealed a small number of uses of `hollow`, `transparent`, and `error`. Module developers should comb over their buttons to ensure they're now using the right options.

### DIFF
--- a/lib/Button/Button.css
+++ b/lib/Button/Button.css
@@ -60,6 +60,39 @@
   */
 
   /**
+  * Button Style: Default
+  */
+
+  &.default {
+    background-color: transparent;
+    border: 1px solid var(--primary);
+    color: var(--primary);
+
+    & :global .stripes__icon {
+      fill: var(--primary);
+    }
+
+    &:focus:active {
+      transition: none;
+      box-shadow: 0 0 0 18px color(var(--primary) alpha(-30%));
+    }
+
+    &:focus {
+      transition: box-shadow 0.3s ease-out;
+      box-shadow: 0 0 0 3px color(var(--primary) alpha(-50%));
+    }
+
+    &:hover {
+      background-color: color(var(--primary) shade(8%));
+      color: #fff;
+
+      & :global .stripes__icon {
+        fill: #fff;
+      }
+    }
+  }
+
+  /**
   * Button Style: Primary
   */
 
@@ -77,62 +110,6 @@
     &:focus {
       transition: box-shadow 0.3s ease-out;
       box-shadow: 0 0 0 3px color(var(--primary) alpha(-50%));
-    }
-
-    &.hollow {
-      background-color: transparent;
-      color: var(--primary);
-
-      & :global .stripes__icon {
-        fill: var(--primary);
-      }
-
-      &:hover {
-        background-color: color(var(--primary) shade(8%));
-        color: #fff;
-
-        & :global .stripes__icon {
-          fill: #fff;
-        }
-      }
-    }
-  }
-
-  /**
-  * Button Style: secondary
-  */
-
-  &.secondary {
-    background-color: var(--secondary);
-    border: 1px solid var(--secondary);
-    color: #fff;
-
-    &:focus:active {
-      transition: none;
-      box-shadow: 0 0 0 18px color(var(--secondary) alpha(-30%));
-    }
-
-    &:focus {
-      transition: box-shadow 0.3s ease-out;
-      box-shadow: 0 0 0 3px color(var(--secondary) alpha(-50%));
-    }
-
-    &.hollow {
-      background-color: transparent;
-      color: var(--secondary);
-
-      & :global .stripes__icon {
-        fill: var(--secondary);
-      }
-
-      &:hover {
-        background-color: color(var(--secondary) shade(8%));
-        color: #fff;
-
-        & :global .stripes__icon {
-          fill: #fff;
-        }
-      }
     }
   }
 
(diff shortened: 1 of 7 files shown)


JohnC-80:
Here's a screenshot with the default button style applied where there were secondary buttons previously... I believe it does echo Filip's concern with the border-only buttons blending in with input fields...(especially text inputs in their :focused state.) IMO, it creates quite a bit of visual noise to see this.
image

cherewaty:
@JohnC-80 thanks for the screenshot, that's helpful 😁 . I think there's enough visual difference between the text fields and the buttons, but I'll let @filipjakobsen make that call.

JohnC-80:
Indeed, I've bowed out of the aesthetics game, for the most part. When a textfield is in focus, its border becomes the same blue as the border of the buttons... granted, those focus styles are WIP... but FWIW: 
textfield-focus

filipjakobsen:
Let’s go with these for now. The thing John mentions is something I’ve already discussed with Kimie Kester as a potential issue. We might update it in another PR later on. But in any case, we will want to tweak all the components and heir styling as we go along.

The most important part about this PR is that it deprecates certain button styles and changes the default button to be a secondary style.

MikeTaylor:
I agree that merging this is the way to go, so we can fail-early on the changed style names.

But I also agree very strongly with @JohnC-80 that we badly need to change the new "default" button style, which really is completely indistinguishable from the active text field. Come on, @filipjakobsen, you can come up with something better!

filipjakobsen:
@MikeTaylor 

Thanks for your input. As I wrote above we are aware of this particular challenge, and this component, along with all others, will be optimized recurringly. For now, the infrastructural part of this PR is of some urgency. A lot of components need styling attention, and we will work through it all. Sometimes it’s more efficient to work with styling once the infrastructure is in place. So this PR is the first step towards working more with the styling of this component.

MikeTaylor:
For sure.

JohnC-80:
@cherewaty, @filipjakobsen - The takeaway from conversations I've been a part of, `hollow` needs to be gone - no issue there - it's rarely used and barely breaking in the actual cleanup work to be done as result. No problems if we feel so strongly about such a thing... now things have swayed to where the former hollow appearance is now the default, removing the very distinguishable `secondary` style that was in place before, broke things further by employing a completely non-semantic `default` class rather than simply setting the existing `secondary` class to be the default.
The statement that 'infrastructure isn't in place' is a complete and total excuse in this instance - removing things that were functional didn't provide any more "infrastructure" than what was present before.

filipjakobsen:
@JohnC-80, when I use the term infrastructure in this context, I mean updating the set of buttons to ensure consistency in usage for the future. I might be using the wrong term.

filipjakobsen:
@JohnC-80 If you feel very strongly about this styling change, I suggest we just do the following:

1. Make all texts on all buttons Semibold (this will distinguish them further from input fields than the rounded corners already do)
2. Change the background color of Default to a 10% opaque --interactionColor, with --interactionColor text

Even if we make the changes above, no one should consider them set in stone. We will need to change styling in accordance with the overall style development as we go along.

Right now, I think it's just important that we change the button naming, deprecate the hollow and transparent styles and implement an unambiguous Disabled style.

screenshot 2018-02-01 15 56 53

MikeTaylor:
For whatever it may be worth, all of that sounds good to me.

JohnC-80:
Here's my suggestion - we put button designs through actual usability testing with users rather than making prescriptions.

filipjakobsen:
@JohnC-80 what if we keep the current styling, and make all the buttons semibold?

Then we can worry about changing styling later on, without rushing it

screenshot 2018-02-01 16 11 20

JohnC-80:
@filipjakobsen I definitely prefer the consistent flatness. Might as well just roll this one back, then! ( I kid... There were positives here, so future PR for the tweak )

cherewaty:
I'll let the dust settle on this for a day, and then I'll make a follow-up PR after the Feb. 2 FOLIO UX call. Thanks y'all.

PR #228, opened 2018-02-09
Upgrade to CI-based @folio/eslint-config-stripes

doytch:


doytch:
@JohnC-80 I'm done with switching this over to @folio/eslint-config-stripes so this is the current set of errors. Merge and clean up later? Last minute complaints about some of the linter rules? :D 

There're a bunch of these unused prop/state errors so I didn't want to just fix them all blindly in case they let the original authors catch any problematic behaviour.

JohnC-80:
Alright. We'll go on another linting pass. thanks!

PR #232, opened 2018-02-13
Update `<FilterGroups>` documentation for recent API changes

MikeTaylor:
Fixes STCOM-206.


MikeTaylor:
@JohnC-80 @zburke It's a bit horrifying that shifting to eslint-config-stripes has thrown up 37 ESLint errors in stripes-components. Do we have a Jira yet for fixing them?

JohnC-80:
I believe much of it was caused by updating the eslint/ eslint configs that have been used in eslint-config-stripes - Either way, I feel like there's almost always going to be more lint errors when you upgrade - the tooling we're using are moving targets as well - hooray, more false-positives! ( I haven't actually looked, but from the look of the errors I'm willing to bet there's some similar cases to the lovely way proptypes have been handled...)
https://issues.folio.org/browse/STCOM-207 is the issue.

MikeTaylor:
Oh, totally, yeah -- it's entirely due to the rules changing to be tighter, not because of the code getting sloppier!

PR #243, opened 2018-02-20
Clean up lint errors

cherewaty:
Lint errors were getting pretty noisy.

I didn't try to fix any errors for `react/no-unused-state lint` - I just disabled those lines. Without some kind of automated test, I wasn't familiar enough with the components to know if removing those states would break functionality.


MikeTaylor:
Please, no. Let's either fix the errors, or continue to live with the notifications.

There's no point in running Lint if we're just going to tell it to shut up whenever it's inconvenient.

skomorokh:
I think it's perfectly reasonable to want to explicitly flag intentional exceptions to our rules as such rather than let them hang around ambiguously.

I agree that if we find ourselves doing so very commonly for some particular rule we don't really have that convention and should alter our lint config, but for occasional exceptions I think it's useful to draw the distinction of it being intentional rather than incidental.

JohnC-80:
While I agree that it's been noisy since the upgrade, I do think that we need to actively examine and fix the errors (especially those flagged as `no-unused-state`)... So for this PR, if the commits for 'quieting' errors can be excluded, then we could probably accept this and decrease the noise pile a bit.

cherewaty:
How about this as an in-between solution? I can merge this and can create a JIRA issue to follow up and properly fix the lint failures for `react/no-unused-state`, `react/no-unused-prop-types`, and `eslint-disable-line no-restricted-globals`. I don't have the background knowledge on those components to confidently make those changes without test coverage, but if one of y'all can commit to that as a top priority, I'll gladly create the issue with all the details.

In the meantime, it's unhealthy to continue to have CI continue to comment on these errors on every pull request. It desensitizes us to them and makes it very difficult to detect when new errors pop up (example: https://github.com/folio-org/stripes-components/pull/242#issuecomment-366978316). Silencing them with a commitment to follow up is a better option than living with the clutter. A CI failure for linting should be a rare exception, not the norm.

JohnC-80:
@cherewaty There's an umbrella JIRA here already: please link any issues you create up to this guy: https://issues.folio.org/browse/STCOM-207

MikeTaylor:
> I think it's perfectly reasonable to want to explicitly flag intentional exceptions to our rules as such rather than let them hang around ambiguously.

I don't think _anyone_ disagrees with this.

That's not what's happening here. When we switched to using `eslint-config-stripes`, ESLint helpfully found for us a whole bunch of real errors in the stripes-components code: a lot of messing about with state that is not used.

Those are errors. We should fix them.

And if we think that they are, for some reason, errors that we don't care about, then we should modify the config to tell it we don't care -- not explicitly annotate each individual instance with a "I know this is wrong but let's pretend it's not" comment.

JohnC-80:
I've noted the components that need attention for state-updates and similar complicated eslint-errors in the STCOM-207 JIRA, so those can go quietly for now to reduce the noise on these PR's. #242's got some other hang-ups as Rasmus' lint setup just isn't working in his dev environment - I might just go in and lint it for him since it's an otherwise fine PR...

skomorokh:
Oh! Sorry Mike, I misunderstood the situation entirely. I definitely agree with you here, why spend time quieting real errors? We should fix those. It's not wrecking lint so completely on stripes-components, you can still see if you introduce new ones, especially when most of the lingering ones are the same. Do we really intend to leave them unaddressed for so very long?

That said, since you did spend the time quieting them and there are JIRA issues filed to make sure we don't forget, I think we should merge the quiescence commits and benefit from the work. Plus you did the work to fix all the other ones, seems churlish to reject your efforts to maximise linting utility.

The restricted global fix is trivial by the way:
https://github.com/airbnb/javascript#standard-library--isnan

PR #244, opened 2018-02-21
Updating AppIcon to support context

rasmuswoelk:
image

This update adds support for React Context in the AppIcon-component.

It relies on stripes being in context which allows it to check for the metadata-key to find icon(s) for an app specified by the "app"-prop.

**Usage:**
`<AppIcon app="users" />`

I also updated the readme with examples.

**Note:** I have preserved two other options of adding icons even though the context one will be the recommended. This is mainly because of some apps not being actual apps yet - such as "Settings". In this case I needed another way of passing the icon data down.

In the future we'll probably limit the options to only using the context part (assuming we don't decide to use another way of implementation).

I have been adding icons to various apps so that they are ready for this change. I will also submit a PR for stripes-core that will activate icons in the MainNav and afterwards we can start adding icons other places in the application.

### DIFF
--- a/lib/AppIcon/AppIcon.js
+++ b/lib/AppIcon/AppIcon.js
@@ -10,19 +10,38 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import css from './AppIcon.css';
 
-const AppIcon = ({ size, icon, style, children, className, tag }) => {
+const AppIcon = ({ size, icon, style, children, className, tag, app }, { stripes }) => {
   /**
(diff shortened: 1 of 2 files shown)


rasmuswoelk:
Note: There are a bunch of linting errors but most of them not related to this PR. I'm assuming these are getting fixed in: https://github.com/folio-org/stripes-components/pull/243

Also, I am currently not able to run `yarn run lint` properly on my local machine. I'm looking into fixing this issue ASAP.

PR #253, opened 2018-02-27
AccessibleFocus component for universal hover, focus and active styling

rasmuswoelk:
This is a big one and it will require some proper testing before getting merged. However, a lot of these changes are purely visual updates and not breaking changes.

## Purpose
To achieve universal hover, focus and active styling for most clickable/focusable elements through a single helper component.

This allows developers to quickly add our uniform styles to any custom component. 

It's implemented in all of our components (or at least it should be - might be missing some but I'll fix those later on if something should come up).

## The ideas behind the design & UX
These changes are made in cooperation with @filipjakobsen. Here are his words on the design and UX.

The purpose of the AccessibleFocus design is three-fold:

1. The accessibleFocus style means that when people increase contrast or change system background color with Theming, in the future, we can make sure the focus-styling is still very visible.
2. The accessibleFocus style means we maintain the styling of focusable elements in one place, so we — with a few exceptions — can control the focus styling in one place and keep it consistent across the system
3. The accessibleFocus style also means that we can ensure a very visible focus styling across all browsers, avoiding doubt or ambiguity in how the styling will look in different browsers.

The styling is meant to fit the general styling of the system, while still being conspicuous enough that ambiguity of what's in focus is not a problem.

The dot provides, in addition to the styled box, a very clear and 100%-contrast interface element which can easily be recognised even on things that already stand out in some other way through their styling.

## The ideas behind the implementation
I decided to go with a helper component instead of simply supplying helper CSS classes. 

In the current state of the component it's simply extending with some extra CSS classes that activates the different styles which means that the same result could be achieved with only CSS classes.

That being said I think it's the right way to have a dedicated component for the following reasons:
- It's easy to switch out with any existing element or component
- It provides a declarative interface with a simple set of props that can be used to achieve the desired outcome for most implementations
- We can always add new features or update styles very easily once it's implemented everywhere

You can check out the readme and examples in the storybook of stripes-components to learn more about how to use the component.

As mentioned above; the component is already implemented in most of our existing components so changes are that you won't ever need to use this component if your UI only contains of our pre-made components.

## Changes in this PR
- Created a `AccessibleFocus` component which supplies styling for :hover, :active and :focus
- Removed most of the current styles in different components and replaced with the new AccessibleFocus component
- Updated minor styling in a lot of different components to make them work with AccessibleFocus
- Minor tweaks to a lot of components on Filip's request

## Some changes that needs to be done
I have tested these changes as much as possible but there might be stuff we need to tweak afterwards.

One thing for an example is Accordions which currently have overflow: hidden; on expanded content. This crops of some of the styling - it isn't critical but it must be dealt with at some point.

image

Another thing is testing. I have only tested this out on Mac but I'll need to test it out properly on Windows too. I know @JohnC-80 is running windows so hopefully he can help me out on finding any weird quirks before this gets merged in.

## Other affected repos
In order to test this properly you'll need to checkout these PR's in the following repos:

- stripes-core: TBA!
- stripes-smart-components: TBA!

## Some gif's!
feb-27-2018 19-06-44

feb-27-2018 19-04-38

feb-27-2018 19-03-48

feb-27-2018 19-17-49


lib/Checkbox/Checkbox.js
```diff
@@ -200,7 +202,7 @@ class Checkbox extends React.Component {
 
     // If no ID is provided we add a random ID
     // (id is required for Checkboxes to work properly)
-    const inputId = this.props.id ? this.props.id : `checkbox-${uniqueId()}`;
+    const inputId = this.props.id || uniqueId('checkbox-');
```
zburke: The reason we changed checkboxes from `<input id="some_id" /><label htmlFor="some_id" />` to `<label><input /></label>` in #270 is that, currently, we do not have a way to guarantee the element IDs are unique (because the spec says they are supposed to be unique) _and_ knowable ahead of time (because we need to know the IDs in order to write tests). 

There is a long discussion of the pernicious effects of duplicate element IDs at https://issues.folio.org/browse/UIU-430, but as noted by @MikeTaylor in the follow up issues, before we can even contemplate checking IDs to ensure they are unique (https://issues.folio.org/browse/STRIPES-512) we need a way to generate repeatable module-ids (https://issues.folio.org/browse/STCOR-173) _and_ we have thread that value down to the components. None of this is to say, "unique element IDs aren't worth it" because obviously it's The Right Way to fix the problem in the long term. Rather, it's that the problem causes very real bugs with an immediate negative impact on end users; fixing unique IDs in the long-term is a lot of work, and fixing it in the short-term with nested label/input elements is a legitimate fix. 

If we need to move away from nested label/input in the long-term, that's fine, but I think we need to resolve STCOR-173 and STRIPES-512 first.

rasmuswoelk: @zburke Alright.. well.. For the new styling in this PR I need the ID's since nesting isn't an option for this custom active-, hover- and focus-styling.

It's a bit problematic if I need to wait before merging in this PR since these changes are long overdue (and I'm being asked by designers/UX'ers when it's up on testing).

Could we fix the current duplicated ID's as a start so we can merge this PR in?

zburke: I understand the desire to move ahead with styling for all elements before this PR diverges further from master. I definitely agree it needs to happen and the sooner the better, but fixing the duplicated IDs is not a small task and any app that uses ui-plugin-find-users will break if you restore an `id` to the `<Checkbox>`'s label. There is no way around it as things currently stand. The filters for finding proxies or sponsors in ui-users will break. The filters for attaching a user to a request in ui-requests will break. 

IMO, we should omit the changes to `<Checkbox>` in this PR and come back to it later, after [STRIPES-512](https://issues.folio.org/browse/STRIPES-512) and [STCOR-173](https://issues.folio.org/browse/STCOR-173) are resolved. 

Look, I get it. If we don't make all these elements consistent, it'll throw the whole feel off, like sitting down to a formal meal and finding plastic utensils on the linen napkin. Probably, this is a question for Cate or Jakub to decide, not a developer (function! function! function!) and a designer (form! form! form!).

rasmuswoelk: @zburke 

Alright :) It's of course sad to hear because I spent a lot of time making these styles work properly in order to ensure proper accessibility through clearly noticeable focus, hover and active styles.

For now I have updated Checkbox so that the styling still works on hover and active but the focus styling will not work until we can add id's on input/label (because CSS doesn't have a parent selector unfortunately). This is of course not ideal but it will have to do for now. Hopefully we can add these ID's to checkboxes soon.

But what about radio buttons - wouldn't those need to be updated as well or aren't they affected by this?

zburke: Hopefully we can solve the element ID problem soon and get your styles properly applied. I'm sure that work will not have been in vain!

WRT radio buttons, you're absolutely correct that this issue affects them too. It just so happens that we don't use any radio controls in plugin-find-user at present so we haven't encountered it.

JohnC-80:
I've added a group of folks for review on this PR... my gut says that having an exclusive component is a bit much for adding a focus style. If I'm using this to build a new component, I still have to write styles for everything else. It seems like taking advantage of css-modules' `compose` functionality from a shared file of focus-styles would accomplish necessary things here. Focus styles really don't need to be complicated ones that lean on z-index and :pseudo elements to make them work. Their settings shouldn't be so complicated as to require simplification through a component's API. This component seems to have a great deal on its shoulders, having to deal with *whatever children get thrown at it as well as whatever should land in the `component` prop. This seems like it would be difficult to test. The checkbox issue and the DOM dependency there also raise red flags for me... but perhaps some of you may prefer this component-driven approach. I know that FOLIO is a project with a great deal of scope and varied experiences/expectations among developers. Thanks for additional feedback from anyone I've marked as reviewers!

rasmuswoelk:
For the visual part: I'll upload the changes to a test server since several other branches in other repos are involved to save everyones time.

doytch:
This is pretty slick! It's _definitely_ beautiful and I have zero comments on that front! 💯 

Here's what I am thinking, though:
1. I generally lean John's way and prefer to do work like this in just CSS. However, I think the necessity of the `focusDot` and `boxOffset` functionality and the simplicity with how you've implemented it sways me. The alternative would be keeping track of all the various className modifiers you can chuck onto it which is more error-prone and has lower DX than this.
1. I think I'd prefer this uses render props rather than passing the component class as a prop.
1. Any thought about adding `tabindex` attributes onto the children by default?

Just some ideas and comments that come to mind. I think this is solid work overall.

skomorokh:
If css-modules has a facility that does what we want here I would strongly prefer if we investigate that instead. I imagine there will be other complex styling we want to reuse and it seems like layering additional components every time we want to do that is unnecessarily indirect if there is indeed a reasonable way to keep it pure CSS. Especially when it seems to still involve adding layers of DOM elements.

From the look of it, compose does provide the modularity we need, if we even need to go that far. And I don't see Mark's point about the DX--how is passing a prop of `boxOffsetStart="medium"` to a wrapping component particularly better than adding the corresponding class `boxOffsetStartMedium`?

doytch:
> And I don't see Mark's point about the DX--how is passing a prop of `boxOffsetStart="medium"` to a wrapping component particularly better than adding the corresponding class `boxOffsetStartMedium`?

Better error reporting via React props validation, really. And this is more personal preference, but I like being able to see the interface of a component by looking at it's JSX file and propTypes instead of going and digging for CSS classes.

rasmuswoelk:
I'm open for suggestions for alternative implementations.

If possible it would probably work with using `composes`  in PostCSS. 

You could also import the CSS classes in each file and combine them with classnames(). I think it will be a lot more difficult for 3rd party developers to use the styles in the correct way. 

Alternatively we could keep it as an internal tool. My guess is that our pre-made components that already implements AccessibleFocus - such as Button, NavList, IconButton etc. - will be sufficient in most cases so that developers won't even need to use the AccessibleFocus. 

However if there should arise some need for a custom button or similar it would be great if developers had an easy way of adding these styles - whether it being a component , a set of css classes or some PostCSS methods used as implementation.

Robdel12:
> To achieve universal hover, focus and active styling for most clickable/focusable elements through a single helper component.

I think any component that is built should already have those boxes checked (or they're not complete imo). And if you're building your own component you should have styles available to cover those bases. 

I guess my main hangup is why should this be a component? To me this is a CSS concern. It'd help me understand if I knew what the component did better than implementing a global `:focus` style.

rasmuswoelk:
@Robdel12 The original idea of the component vs. importing the CSS classes manually was to make a convenient helper for 3rd party- and folio-devs to add these styles without having to go through a lot of CSS classes to carefully select the ones that are needed in order to make it work properly.

The styles themselves are a bit advanced I will admit. The styles was build in cooperation with @filipjakobsen. They utilize pseudo states `:before` and `:after` to set both focus, hover and active styles as well as the optional focus-dot. The current settings is the positioning and offset of the focus-dot and the offset of the hover, focus and active styles box. In order to make these work it's required for the element that has these styles applied to have inner elements (this is because of an z-index issue). The component itself could be used to warn devs when it's not properly applied - e.g. a string is passed as the only child.

To illustrate the number of classes to choose from.. E.g. for the focus-dot there's a class for adding the dot itself,  a class for positioning (4 in total - one for each side). Then there's an additional option to change the offset of the focusDot - that's 3 additional CSS classes to choose from. And then there's the control of the offset - e.g. you want a larger offset in the right to align properly with other elements (to please basic rules of design).

All that being said: I don't mind changing the solution into importing CSS classes in each component if having a helper component makes people feel uneasy. I can see both pros and cons but as mentioned earlier; I'm really just interested in getting the new styles applied asap.

cherewaty:
@rasmuswoelk this PR will be a good historical document of the strategy, but can you close this and make a new PR for the CSS changes, with squashed-up commits? The approach has changed so dramatically since this PR opened that merging this will be very confusing for developers looking back a year or two from now.

rasmuswoelk:
@cherewaty Yes. Probably a good idea 👍

rasmuswoelk:
Cleaned PR: https://github.com/folio-org/stripes-components/pull/339

PR #262, opened 2018-03-06
Added styling for disabled FilterGroups and Accordions

doytch:
Disabled FilterGroups didn't have obvious styling to show they were disabled beyond the CheckBox itself. This adds background styling by adding a `disabled` prop to `<Accordion>`s themselves and CSS styles on that class.

Did it this way since we wanted both the header and `<FilterGroup>` styled and it'd let us just have the one set of rules. The ruleset is general enough to work for an Accordion in any parent that has a `major-padding` for its padding (like SearchAndSort).

Screenie:
screen shot 2018-03-06 at 3 26 55 pm

### DIFF
--- a/lib/Accordion/Accordion.css
+++ b/lib/Accordion/Accordion.css
@@ -15,9 +15,9 @@
 }
 
 .disabled {
-  /* Doing this math since the SAS pane has padding that surrounds Accordians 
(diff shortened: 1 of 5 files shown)


MikeTaylor:
That's some sweet work!

PR #267, opened 2018-03-07
pass inputProps to country fields for Typeahead

zburke:
`<Typeahead>`, used in the ui-users `<Autocomplete>` implementation, deprecated the `name` prop in favor of `inputProps`. Somewhere along the line, the `name` prop passed to `<Field>` in EmbeddedAddressForm is swallowed, but `inputProps` sails on through.

Refs [UIU-427](https://issues.folio.org/browse/UIU-427)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Proper documentation for `makeQueryFunction`. Fixes STCOM-221.
 * Make `makeQueryFunction` robust to failed substitutions. Fixes STCOM-225. Available from v2.0.3.
 * `makeQueryFunction` favours parameter-access via the anointed resource rather than the URL. Fixes STCOM-226. Available from v2.0.4.
+* Pass `inputProps` to country fields for consumption by `<Autocomplete>`. Refs UIU-427. Available from v2.0.5.
 
 ## [2.0.0](https://github.com/folio-org/stripes-components/tree/v2.0.0) (2017-12-07)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v1.9.0...v2.0.0)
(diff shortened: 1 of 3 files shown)


JohnC-80:
The `<Typeahead>` has been shouting for replacement for a while... I guess it finally broke through to the console. This doesn't happen to touch [UIU-298](https://issues.folio.org/browse/UIU-298) I doubt it does, but just curious there...

mkuklis:
@JohnC-80 I think we are missing country validation on user form and that's why we are running into UIU-298.

@zburke this looks good and we should merge it. I also agree with @JohnC-80 that `<Typeahead>` should go away. Perhaps we can replace it with a simple `Select` for now.

zburke:
No relation to UIU-298. 

In trying to combat the console warnings generated by `<Typeahead>` in ui-users, I realized the `name` prop is getting swallowed before it even gets that far, so while changing `<Typeahead ... name={label}>` to `<Typeahead ... inputProps={{ name: label }}>` quiets the warning, it is insufficient to place the correct `name` attribute on the field**. This PR, along with https://github.com/folio-org/ui-users/pull/278, address the latter issue, threading the fieldname through via `inputProps`. 

** Even though the field has an invalid name, e.g. `Country` rather than `personal.addresses[1].country`, it is somehow correctly handled when submitting the form. 

What do you think, @JohnC-80, @mkuklis -- is it worth making these changes or should we simply hold off until `<Typeahead>` is replaced?

JohnC-80:
FYI guys - this is what I've got going on... it's a select component with a filterable options list...
selectno2

mkuklis:
nice @JohnC-80 :) Thank you!

zburke:
Holding off per @JohnC-80.

PR #268, opened 2018-03-08
InformationIcon

rasmuswoelk:
ImformationIcon component for displaying some extra information or explanation to a feature or similar.

The styling might need to be updated a bit but this is a good start.

In this PR I also adjusted the styling of the Popover-component, added examples and added a couple of useful props to the component.

**Issue:** 
https://issues.folio.org/browse/UIORG-48

**Screenshot:**
informatioj-icon


lib/InformationIcon/InformationIcon.js
```diff
@@ -0,0 +1,57 @@
...
+      <IconButton
+        className={css.icon}
+        data-role="target"
+        icon="info"
+        size="small"
+        iconSize="small"
+      />
+      <div data-role="popover" className={css.wrap}>
+        <div className={css.content}>{content}</div>
+        <Button
```
Robdel12: Looks like we're using a button & treating it like an anchor link. Is there any reason (beyond style) for this? We lose a bit for accessibility by doing this.

rasmuswoelk: @Robdel12 Hi Robert!

Hmm. No I don't believe it does. This should probably be added in the Popover-component as a default feature.

rasmuswoelk:
@Robdel12 Hi Robert!

Hmm. No I don't believe it does. This should probably be added in the Popover-component as a default feature. I could look into this.

lib/InformationIcon/InformationIcon.js
rasmuswoelk: @Robdel12 You are totally right! It should be a regular anchor-tag but with the button styling. I'll change that - thanks!

JohnC-80:
2 cents worth: I think it might fit best in the `lib/structures` folder rather than among all of the building-block components. Also, perhaps a more descriptive name like `<InformationPopover>` might be best, since it is more than just an icon.

rasmuswoelk:
@JohnC-80 That makes sense. And yeah.. I had my doubts about the name. InformationPopover sounds good.

rasmuswoelk:
@JohnC-80 @Robdel12 

Some changes:
- Updated name to InfoPopover
- Moved into structures
- Button is now an anchor-tag (Had to adjust the Button-component slightly but now it supports regular links/anchor-tags which it didn't before)
- Added focus to Popover body/content on open

What do you guys think.. Can we merge it in? 😎

lib/Popover/Popover.js
```diff
@@ -63,13 +76,22 @@ class Popover extends React.Component {
...
+      const { open } = curState;
       const nextState = Object.assign({}, curState);
-      nextState.open = !curState.open;
+      if (!open) {
+        nextState.open = true;
+        this.timer = setTimeout(() => this.setState({ visible: true }), 25);
+      } else {
+        nextState.visible = false;
+        this.timer = setTimeout(() => this.setState({ open: false }), 100);
+      }
```
JohnC-80: hmm... is there a better way than setting timeouts to here to trigger animation?  Can we hook into React lifecycle to do this without having to trigger re-renders via additional `setState` calls 25ms or 100ms later?

lib/Popover/Popover.js
```diff
@@ -11,23 +13,28 @@ const propTypes = {
...
-  offset: 0,
+  offset: 5,
+  noPadding: false,
 };
 
 class Popover extends React.Component {
   constructor(props) {
     super(props);
 
+    this.timout = null;
```
JohnC-80: `timout` and `delay` - are these guys necessary?

lib/Popover/Popover.js
```diff
@@ -288,7 +261,7 @@ class Popover extends React.Component {
               ),
             );
             this.rendered_target = (
-              <div className={css.popoverTarget} ref={(t) => { this._target = t; }}>
+              <div className={classnames(css.popoverTarget)} ref={(t) => { this._target = t; }} key="target">
```
JohnC-80: A mild one, but I feel the inline `classname()` calls could be cleaned up. I'm not sure why a call is used here to apply only a single class...

rasmuswoelk: That's true. I think I had something else added but removed it and I never removed the classnames function. I'll remove it 👍

rasmuswoelk: Removed.

lib/Popover/Popover.js
rasmuswoelk: So the problem was that the parent of the <Transition> (the popover) was being mounted and unmounted by the React Portal before the Transition could do it's magic.

One solution was to do a setTimeout to make a small delay before rendering the popover. I see that this is kinda dirty so I tried finding another way.

I found that in React's <Transition /> component you can set the prop `appear` which triggers the transition once it renders (in this case; when the Portal is mounted). That solved the issue for the enter-transition but it would still disappear immediately without transition when closing the popover. 

I found out that you can listen for a `onExited`-event on <Transition> so when opening the Popover I flip the `open` and `visible` states to true but when closing I'm only flipping the `visible` to false which triggers the transition and once that's done (`onExited` callback) I'm flipping the `open`-prop to false to properly remove the popover/Portal from the DOM.

My plan is to make a transition-helper that we can re-use for components like Dropdown, Popover, Tooltip etc.

lib/Popover/Popover.js
rasmuswoelk: Removed.

lib/Popover/Popover.js
rasmuswoelk: Commit: https://github.com/folio-org/stripes-components/pull/268/commits/a2301cec84689dfe2f6fc1098b6d19935cadc457

JohnC-80: So I'm wondering - from the accessibility side of things - if it would be better to have things rendered, but visibly hidden (perhaps suspended via `shouldComponentUpdate`) as well as `aria-hidden="true"` and `tabIndex="-1"` etc... that way the container is always there so the appearance/disappearance of components isn't too much of a concern. @Robdel12 - any insight?

Robdel12: Yeah that sounds like a solid idea. That's usually what I do when working with css transitions.

rasmuswoelk: @Robdel12 @JohnC-80 We can most definitely do that! However, on a page with a lot of Popovers (e.g. a list with a lot of rows) it will render a lot of popovers into the DOM.

If this isn't an issue I can update the <Popover> Portal to be rendered but hidden (and implement the shouldComponentUpdate as well).

Robdel12: Oh interesting! It doesn't use the same container for each popover? I figured using a portal would send the popover markup to another part of the DOM (thus only creating one that gets rendered into when "open")

JohnC-80: I think it's a good idea to have a designated `<div>` where these things live... `#overlayContainer` or something (similar to `#ModuleContainer` that already exists... it would really help us manage focus in the case of Layers/modals needing to have their focus 'quarantined' to just their contents, overlays, and the main navigation.

Robdel12:
Nice! 🎉

PR #271, opened 2018-03-09
restore checkbox position. Refs STCOM-227

zburke:
Prior to this change, the UI element was 9999 pixels offscreen, so there must have been some additional CSS/JS magic going on here prior to #270 (See [UIU-430](https://issues.folio.org/browse/UIU-430)). 

Tagging @rasmuswoelk in case there is any restorative CSS work to make this pretty again now that we've made it functional again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 * Make `makeQueryFunction` robust to failed substitutions. Fixes STCOM-225. Available from v2.0.3.
 * `makeQueryFunction` favours parameter-access via the anointed resource rather than the URL. Fixes STCOM-226. Available from v2.0.4.
 * In `<Checkbox>`, nest HTML checkboxes inside labels instead of associating the labels by ID. Fixes STCOM-227. Available from 2.0.5.
+* CSS tweak for `<Checkbox>` to bring the UI element back on screen. Refs STCOM-227. Available from 2.0.6. 
 
 ## [2.0.0](https://github.com/folio-org/stripes-components/tree/v2.0.0) (2017-12-07)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v1.9.0...v2.0.0)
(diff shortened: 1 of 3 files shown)


rasmuswoelk:
@zburke The input itself is supposed to be off the screen because we made our own custom Checkbox-styling because styling of native inputs is very limited. The input itself still needs to be "visible" (we can't hide it with visibility: hidden; or display: none;) so that it can be toggled when the label is clicked.

But no worries - I'll fix it.

Another thing though: After `id` was removed from input and `htmlFor` was removed from `label` we are getting a linting error: `Form label must have associated control jsx-a11y/label-has-for`. This is an accessibility issue that we need to fix (i'll fix this as well since I'm looking to merge in #253 soon)

It's just weird that this error didn't come up by jenkins when the PR was merged in #270 🤔

zburke:
@rasmuswoelk, a couple things: 

1. The eslint a11y doc says [nesting by wrapping a control in a label tag](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md) is a legitimate way to associate a label with a control, which explains why there was no lint error when we ran lint locally and why QC didn't report one on t. It's odd to me that you're seeing one!
1. Additional context at https://issues.folio.org/browse/UIU-430 and in my comments on https://github.com/folio-org/stripes-components/pull/253/commits/6782a4b8212fdcb50021beb9a0b77aadb16b2e6b WRT #253. We don't have unique element IDs at present (ugh; our bad), nor do we have an easy way to generate them, but there are user-facing bugs this implementation solves. 

Let's talk more about the best short-term solution. Feel free to grab me on Slack, or file a JIRA issue so we can preserve the comments; whatever works best for you!

rasmuswoelk:
@zburke Ah I see. Okay, then it's weird I'm getting linting errors. I'll double check to see if it's still the case.. 

For #253 I actually need ID's on label/input in Checkbox so I hope it's okay that I re-add those. I have made sure that it always renders with a unique ID if no ID was specified. I also made a test for this (renders 100 Checkboxes and checks for duplicate ID's).

However, this solution doesn't take manually added duplicated ID's in to consideration so it would be good to have some solution for that as well.

Further discussion on Slack or Issues.folio.org 👍🏻

PR #298, opened 2018-03-26
STCOM-238 Update list of exports

cherewaty:
Per conversation at https://github.com/folio-org/ui-eholdings/pull/295#discussion_r176257629, we'd rather do:
```
import {
  Button,
  Icon,
  IconButton
} from '@folio/stripes-components';
```
than
```
import Button from '@folio/stripes-components/lib/Button';
import Icon from '@folio/stripes-components/lib/Icon';
import IconButton from '@folio/stripes-components/lib/IconButton';
```
... but the list of exports in `index.js` is out of date. I just added the components lacking to have `ui-eholdings` use the more elegant imports, but there are plenty more components in the library not being exported. I'll make a JIRA issue to make sure the rest of the components get added to the index.

As future components get added to the library, we should make sure they get added to `index.js`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,20 @@
-/*form elements*/
+/* form elements */
+export { default as Badge } from './lib/Badge';
 export { default as Button } from './lib/Button';
 export { default as Checkbox } from './lib/Checkbox';
+export { default as Datepicker } from './lib/Datepicker';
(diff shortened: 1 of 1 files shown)


cherewaty:
Issue for rest of the work: https://issues.folio.org/browse/STCOM-238

PR #299, opened 2018-03-27
STCOM-239

szkata:
Update the EntrySelector component to use translations.


lib/EntrySelector/EntrySelector.js
```diff
@@ -116,7 +116,7 @@ class EntrySelector extends React.Component {
     const LastMenu = addMenu || (
       <PaneMenu>
         <Button title={addButtonTitle} onClick={this.props.onAdd} buttonStyle="primary paneHeaderNewButton">
-          + New
+          {this.props.stripes.intl.formatMessage({ id: 'stripes-components.new' })}
```
doytch: Maybe I'm being daft but I don't see where `stripes-components.new` is being defined. My tip of `stripes-components` doesn't have a `translations` directory or any translation files. Should this be referring to `stripes-core.button.new`?

szkata: You are right, I fixed it.

PR #301, opened 2018-03-27
UIEH-212 Make Settings component responsive

cherewaty:
## Purpose
The `Settings` component was unusable on narrow screens. The UX pattern was such that if you selected a module in the leftmost pane, you'd be taken directly to the first item in that module's subnav, with no way to ever access the subnav.

Groundwork for https://issues.folio.org/browse/UIEH-212

## Approach
Adjusting for a narrow screen meant several changes:
- Assuming the first link in the list is where the user wants to go means the subnav won't appear on narrow screens. The automatic selection of the first nav item has been removed.
- There's now a back button in the `firstMenu` of the pane containing the module's settings pages.
- To get around the brittle logic of `activeLink` (makes some assumptions about url structure), consumers of the `Settings` component can now pass in an `activeLink` prop.

### Bonus
We have plans for three pages in the eHoldings settings, but the right order for those is not alphabetical. The ordering of the nav list items should be up to individual modules. I removed the automatic alphabetization.

## Module adjustments
Module developers shouldn't see any breaking changes, but they should add back buttons to their individual settings pages to take advantage of the new responsiveness.

## Follow-up
The pane containing the subnav now is much wider on wider screens, because the `Paneset`' is fairly limited in its props. Ideally, `flex-grow` and `flex-shrink` would be editable when dealing with those types of layouts, not just `flex-basis`.

### DIFF
--- a/lib/Settings/Settings.css
+++ b/lib/Settings/Settings.css
@@ -1,0 +1,7 @@
+@import '../variables.css';
+
+.settingsBackButton {
+  @media (--mediumUp) {
+    display: none;
+  }
+}
--- a/lib/Settings/Settings.js
+++ b/lib/Settings/Settings.js
@@ -56,14 +56,6 @@ class Settings extends React.Component {
       />);
     });
 
-    const defaultRoute = !Saved ? null : (
-      <Route
-        key={0}
-        path={props.match.path}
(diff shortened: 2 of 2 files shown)


JohnC-80:
Nice! Give me a chance to pull this down and have a look with the other modules that use it.

PR #303, opened 2018-03-28
STCOM-238 Update exports to include all components

cherewaty:
Encouraging
```
import { 
  Button, 
  Icon, 
  IconButton 
} from '@folio/stripes-components';
```
over
```
import Button from '@folio/stripes-components/lib/Button';
import Icon from '@folio/stripes-components/lib/Icon';
import IconButton from '@folio/stripes-components/lib/IconButton';
```

Finishes https://issues.folio.org/browse/STCOM-238


cherewaty:
I didn't investigate the tree-shaking - I'll take a look.

PR #304, opened 2018-03-28
use code -> name mapping in AddressFieldGroup

zburke:
The previous implementation of AddressFieldGroup mapped country-name to
country-name in its select-list of countries. This required every
component that used this one to translate from country-name to
country-code in order to properly store the country-code on the
back-end, and it broke down in the case of translated country names.

Fixes [STCOM-24](https://issues.folio.org/browse/STCOM-242)


MikeTaylor:
Ah, sorry -- those new ESLint errors are of course due to this config change: https://github.com/folio-org/eslint-config-stripes/pull/21

We could change the {{quoted-props}} configuration to "don't care, go nuts"; or we could fix the code to meet the new rule. The problem here is that the "consistent" setting is not merely a relaxation of the old "as-needed" -- it actively conflicts with it in cases like the present one where some properties need quotes and others do not.

What do you think?

zburke:
Eh, I agree that in your example in that PR, the code looks better with consistent quotes, even if there's not huuuuuge value in that consistency. Outside `ui-inventory/data/languages.js` which can probably be handled with a judicious regex search/replace, there are ~50 instances that need cleaning up, which is not much to manage. 

I say keep the rule and clean the code.

MikeTaylor:
OK, thanks!

PR #306, opened 2018-03-28
lint

zburke:
Now, I know what you're thinking. "Did his quotes match or not?" Well to
tell you the truth in all this excitement I kind of lost track myself.
But being this is ESLint, the most opinionated linter in the world and
would complain about pretty much anything, you've gotta ask yourself one
question: "Are the keys in that JavaScript object literal consistently
quoted?" Well, are they, punk?


MikeTaylor:
I laughed immoderately at your PR comment -- so much that I interrupted Fiona and Jonno's biology revision so I could read it to them. Then it turned out that neither of them had seen, or even heard of the scene from, Dirty Harry. So I had to explain that first, _then_ read the message. It was a whole thing.

PR #307, opened 2018-03-28
STCOM-194 maximize panes functionality.

JohnC-80:
To add this functionality to a pane, a developer simply has to use the `maximizeable` prop. It'll throw a 'maximize' action in the Pane's actions menu.
panemaximize


cherewaty:
I haven't seen this exact pattern before in a web app, so I have some questions:
- What happens when a user clicks the close X in the top left of a maximized pane? Does it "un-maximize" the pane and go back to a three-pane layout? Does it close the pane completely and go to a two-pane layout (I'm guessing this is true)? If that's the case, wouldn't a back button make more sense than a close X?
- What happens when the user uses the browser back/forward buttons, refreshes the page, or shares the URL? It looks like the state is only maintained within the component state, so the maximization wouldn't persist. Changes like that to layout that don't persist can be very frustrating.
- What if the "maximize" action were simply a link to another URL instead of a layout state action? That's JIRA's solution to the problem where you might want to see the third pane of a layout by itself. It seems like every pane that we would want to have maximization abilities should have its own URL anyway. We've done that on eHoldings - every detail view of a provider, package, or title that could be seen in the third pane of a search layout also has a url where it can be seen by itself. This solution would also solve the persistence problem, and fit in better with the natural usage of a browser.

JohnC-80:
Agreed. I think persistence of this is definitely a necessary future step - not so far away - having the url drive this shouldn't be too terrible. 
Dismount behavior is something necessary here - thanks for bringing up the close behavior. It should go back to a two-pane layout (having closed the third) and behave in the same way as if the Pane were un-maximized.
As far as the back button goes, my thoughts there were to keep the "un-Maximize" action in the same place where the maximize action lived... but something more immediate might definitely be better - @filipjakobsen ?

filipjakobsen:
The behavior I have designed for this feature (incl. un-maximize) is specified in this JIRA issue as an animation and still frames: https://issues.folio.org/browse/UX-114

@JohnC-80, @cherewaty, whichever technical solution you folks decide on for the URL, please remember that this setting is a per layout per user preference that can be toggled, and so changing URL's based on whether something is maximized potentially runs counter to how the UI is structured. Depending on how exactly the URLs currently work that may or may not be problematic.

JohnC-80:
@filipjakobsen Just to clarfy - in the instance where you'd be viewing a maximized pane in your browser and you want to send a link to that view to someone else, you wouldn't expect it to be maximized by the url, but in whichever state (maximized or un-maximized) the receiving user was viewing the module at.
.

filipjakobsen:
@JohnC-80 Yes, I would expect it to be equally possible and easy to share the link to the record from a non-maximized pane as from a maximized pane. The maximization of the pane ought not to have an impact on that.

cherewaty:
@filipjakobsen The problem I see trying to be solved: if I'm looking at a three-pane search layout, I want a way to see any one of the three panes in isolation.
- How to see the search/filter pane by itself: don't input a search query.
- How to see the results pane by itself: click the search icon at the top left of the pane to close the search/filter pane.
- Proposal for how to see the preview/detail pane by itself: click a link labeled "Maximize".
  On https://folio.frontside.io/eholdings/providers/697?searchType=providers&q=w&searchfield=title, clicking "Maximize" for the W.W. Norton & Company, Inc. detail pane would open https://folio.frontside.io/eholdings/providers/697. The full-width view at https://folio.frontside.io/eholdings/providers/697 would have a back button in the top left corner to return to the three-pane layout.

With those three actions, we solve the problem: every pane of the layout can be seen by itself.

The pattern built here is highly unconventional for a web app. I'm worried we're way overthinking the problem and building a solution that doesn't follow expected conventions in the browser, when we can solve the problem in a much simpler and expected way.

filipjakobsen:
Thank you for your comment, @cherewaty.

The problem that this is primarily designed to solve is for the users to consistently be able to view preview panes in a full-width layout, rather than per item to choose to view them in full view. This gives the users who prefer this maximum space for the results, and maximum space for preview. For the users who prefer a side-by-side view, the un-maximized pane is useful.

I am not sure if you have other needs for eHoldings, but this is the problem meant to be solved with the pattern I have designed.

filipjakobsen:
With the pattern I have designed, described above, the user would be returning to a 2-pane layout when closing the full-width preview pane, rather than a 3-pane layout. As demonstrated in the JIRA issue for this designed that can be found on https://issues.folio.org/browse/UX-114

ttamj:
@filipjakobsen, I agree with with @cherewaty.  This approach to maximizing seems very unconventional.  The concept of maximize is well established and this seems to run counter to expectations.  I also echo the desire to link directly to a detail view, particularly given opening search results in a new tab is so commonplace on the web.

I like the idea of giving users a preference between 2-pane/3-pane views.  Do we have (or plan to have) a dedicated space for user preferences in FOLIO?  I ask because my secondary concern here is this pattern conflates preference with toggle state, rather than allowing for a true default user preference.

JohnC-80:
For url functionality, routing within the ui-module should be sufficient enough to render a paneset with a single pane (or whatever it would like, for instance, to include the Notes pane as well.)
If a user wants to set a preference for a module within the module's settings to always show its detail panes in maximized mode, then it should be able to modify the hrefs to its result set for that display. The back button for the maximized view in either case, would probably function the same as the browser's back button. I feel like we should at least prototype this approach before pursuing this further at the component level.

filipjakobsen:
Hi @johnc-80, @cherewaty and @ttamj,

Khalilah is in the process of setting up a meeting on this topic, and I feel that it would be preferable to discuss this topic in that meeting, rather than trying to do it in this PR, since:
A. @johnc-80, I do not understand what you want me to prototype.
B. It seems to me that we are suggesting solutions to various different problems (temporary maximization; permanent maximization; URL behaviour; back button behaviour; interface layout; and the display of individual records), and that being the case, to my mind, it seems hard to reach agreement about the best path forward.

JohnC-80:
Closing this until we do have the meeting and hopefully decide on the path forward.

PR #312, opened 2018-03-30
provide title attribute to EntrySelector links. Fixes STCOM-244

zburke:
Fixes [STCOM-244](https://issues.folio.org/browse/STCOM-244)


zburke:
@JohnC-80, any problems you foresee with this?

zburke:
From @JohnC-80: 
> so this primarily for testing purposes? I'm afraid it might be an accessibility ding for things that are over-titled - they can end up repeating the title on a screen reader. like it'll say the title then say the actual text in the link. could you use a data-attribute rather than `title`? `data-test` or something? 

Rather than add markup exclusively to enable testing, we'll try to write tests that do a better job picking up on the markup that's already in place.

PR #314, opened 2018-04-02
UIEH-130 Create z-index management solution

cherewaty:
## Purpose
`ui-eholdings` uses `z-index` to manage layering of its search layout. The dropdown menu from the header avatar ends up underneath the third pane: https://issues.folio.org/browse/UIEH-130

Handling `z-index` can be notoriously finicky. The solution to a problem is frequently "set it to a very high number," but with many layers happening, that can get messy real quick.

## Approach
I've used this super clever Sass way of handling z-index on several previous projects: https://www.smashingmagazine.com/2014/06/sassy-z-index-management-for-complex-layouts/

Since we don't have Sass in our toolkit, I wanted to achieve a similar setup with PostCSS. With `postcss-functions`, I was able to create a `stripesZIndex()` that works off the same idea. We define an array of layers in one place, then every time we need to define a z-index we place a new layer in the array. If there's a problem with the ordering, that can be solved from one place, instead of having to track down every z-index defined across multiple repos.

Note: `stylelint` was incorrectly throwing `function-name-case` violations with my camelCased function name, so I just turned that rule off for now.

## Risk
There's a good chance this might break an existing z-index behavior, but it will make fixes to those types of problems very easy.

## Dependencies
Requires `stripes-core` PR: https://github.com/folio-org/stripes-core/pull/257

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -14,6 +14,7 @@ const postCssNesting = require('postcss-nesting');
 const postCssCustomMedia = require('postcss-custom-media');
 const postCssMediaMinMax = require('postcss-media-minmax');
 const postCssColorFunction = require('postcss-color-function');
+const postCssFunctions = require('postcss-functions');
 
 module.exports = {
   entry: [
@@ -72,6 +73,9 @@ module.exports = {
                 postCssCustomMedia(),
                 postCssMediaMinMax(),
                 postCssColorFunction(),
+                postCssFunctions({
+                  glob: path.join(path.resolve(), 'lib/sharedStyles/functions', '*.js'),
+                }),
               ],
             },
           },
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -15,5 +15,6 @@
         "compose-with"
       ],
     }],
+    "function-name-case": null
   }
 }
--- a/lib/Datepicker/Calendar.css
+++ b/lib/Datepicker/Calendar.css
@@ -5,7 +5,7 @@
   border: 1px solid #bcbcbc;
   padding: 4px;
   box-shadow: 0 4px 19px 0 rgba(0, 0, 0, 0.39);
-  z-index: 9999;
+  z-index: stripesZIndex('calendar');
 }
 
 .srOnly {
--- a/lib/Dropdown/Dropdown.css
+++ b/lib/Dropdown/Dropdown.css
@@ -15,7 +15,7 @@
 }
 
 .DropdownMenuTether {
-  z-index: 1000;
+  z-index: stripesZIndex('dropdownMenuTether');
   display: none;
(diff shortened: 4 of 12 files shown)


skomorokh:
I really like the general idea and think we need something like this. But the implementation doesn't sit well with me because it hoists things that should be private to a particular app to this global library. eg. I think it's more appropriate to position something like 'appContainer' at this level and provide a mechanism for apps to subdivide their own space rather than bring app-specific details like 'searchFiltersPane' into the same list as global facilities like 'modal' and 'toast'.

What if apps each could provide a key on the stripes object in package.json with an ordered list of named layers? Then we can make 'appContainer' as deep as the longest list (or choose an arbitrary max for simplicity) and build a mapping per-app from the data loaded in from stripes-config.

cherewaty:
@skomorokh you're right that ideally each module could handle its own layers, but I'd rather start with something simpler. This fixes immediate bugs and gives us the ability to iterate. When it starts to become a common problem that individual modules are relying on a globally-scoped list, we can do the not-insignificant architecture to handle that. I'm all about just-in-time abstractions over premature over-engineering.

skomorokh:
Eh, it's a matter of degrees. There's already a place for per-module metadata that can readily accommodate the list and aggregates it such that you can `import { modules } from 'stripes-config';` so it isn't really adding a lot of new architecture. You're 90% of the way to a general solution. And it's not terribly speculative/solution-in-search-of-problem to assume apps will make use of a facility for this, after all, CSS has z-index to begin with, you have app-level layers for ui-eholdings, etc.

When (do you really think it's an if?) someone needs to set a z-index in an app they'll have to choose between context switching back to solve this for the framework or some convenient workaround like repurposing `searchPaneVignette` as the layer for `budgetProjectionTooltip` or whatever. Or, worse, hardcoding some integers. So we'll accumulate stuff like that for a while.

On the other hand, we got this far without it so _shrug_.

MikeTaylor:
For what it's worth, I'm with Jason on this one (despite my tendency to think he has a habit of over-complicating simple fixes in search of perfection).

JohnC-80:
Eventually we'd like to get to the point where we've eliminated as many of the post-css plugins as we can due to the future-standards they support being implemented in browsers. I don't think this one would ever be eliminated.
If this is really the best way to handle this, do you think there's a way that we could put a cap on the ever-growing list? As we add new components, new ones get developed - even added by 3rd party devs - would it be best to have a smaller list of mapped levels with possibly varying offsets?

cherewaty:
Closing for now, with follow-up ticket: https://issues.folio.org/browse/STCOM-247

eHoldings is no longer blocking the logout dropdown menu with https://github.com/folio-org/stripes-components/pull/317; that was the original intent of UIEH-130

PR #319, opened 2018-04-06
STCOM-249 EditableList: Fixed Cancel button not resetting changed values

doytch:
Clicking the cancel button on a form managed by `<EditableList>` does not restore the original values. It removes the editable field, but leaves the changed value in place. Navigating away then causes the "unsaved changes" prompt to appear.

This change updates `redux-form`'s `initialValues` after every successful save, and the calls `reset()` on any Cancel button press.

### DIFF
--- a/lib/structures/EditableList/EditableListForm.js
+++ b/lib/structures/EditableList/EditableListForm.js
@@ -28,6 +28,14 @@ const propTypes = {
   */
   initialValues: PropTypes.object, // eslint-disable-line react/no-unused-prop-types
   /**
+   * Callback provided by redux-form to set the initialValues to something else.
+   */
+  initialize: PropTypes.func,
+  /**
+   * Callback provided by redux-form to set the field values to their initialValues
+   */
+  reset: PropTypes.func,
+  /**
    * Callback for saving editted list items.
    */
   onUpdate: PropTypes.func,
@@ -186,6 +194,9 @@ class EditableListForm extends React.Component {
         return newState;
       });
     }
+
+    // Reset the field values.
+    this.props.reset();
   }
 
   onSave(fields, index) {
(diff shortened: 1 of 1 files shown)


JohnC-80:
Thanks, @doytch! Would've thought the intialValues would come back anyways from a GET that happens after an item is saved - but I could be wrong there - and who knows how things change... 
Say we have a case where multiple items are in edit mode - calling reset at the form would probably reset all of the items in the list, right? It would be nice to pull those reset values only for the item that's been canceled, if possible.

doytch:
@JohnC-80, you're right that the initialValues do come back after a different item is saved. But that still means the edit-cancel behaviour is broken for the time before a different item is saved, assuming one even is saved.

And yeah, if multiple items are in edit-mode, then canceling one reverts the others (though they still stay in edit mode). Will hack around further on this to try and fix that. Also, Julian found and filed UIU-465 which is related so I'm gonna try and handle that in this PR as well.

doytch:
Looks like UIU-465 is gonna be handled in stripes-smart-components so I won't be making any changes to `<EditableList>` after all for that.

doytch:
So there's no built-in way of targeting resets in redux-form besides using form segments, which doesn't really apply for us here. The alternative looks like it would be maintaining a _separate_ "initialValues" that's actually the initialValues overlaid with the value of fields in edit-mode. That's pretty smelly IMO. Also pretty complicated for what I feel is not a common use case. So I think it makes sense to live with this PR merged for a while and decide later if we do want to do that.

Also got confirmation from Filip that we should in fact be showing the value on the server when not in edit mode. So **Cancel** should reset the value and not just turn off edit mode.

Will merge later today barring any -1s 😄

PR #320, opened 2018-04-09
Setup New Testing Environment

wwilsman:
## Purpose

This lays out the groundwork for creating new tests for all of the components. These new tests should live alongside the component in a `tests` directory, and end with `-test` to be picked up by Karma.

These tests will be run in a _real_ browser in CircleCI. Later, it should be fairly easy to run our same tests in Firefox, or in Browserstack, so that we can accurately cross-browser test all of our components.

## Approach

First, I removed the old test files. There were a few existing tests for the dropdown component, but not only could I not get them to pass locally, but the style of the tests does not work well with what is being proposed here.

Second, I created a testing "harness." For the **majority** of components, this harness _should be unnecessary_. The harness creates a new `Stripes` instance and provides contexts typically found in a stripes app such as a `Router` and providers for `Redux`, `Intl`, and `Apollo`.

To facilitate any necessary requests that must be made to fulfill the okapi redux store's expected shape, [pretender](https://github.com/pretenderjs/pretender) was added. Pretender allows us to fake network requests and provide test-specific data in return. For our components, this _should not_ be needed, but since I do not know how every component works, I added it just in case. 

Third, config for CircleCI was added for linting and testing the components. The tests are run using `stripes-cli`, but there is an expectation in `stripes-cli` that tests can only be performed on apps. So `"type": "app"` had to be added to the `package.json` file so that `stripes-cli` will run our tests.

And finally, I added some example tests around the `Accordion` component. I only wrote a few to get us started, but with the few I did write I found two issues. One issue is when providing an `id` the accordions stay closed by default regardless of the `closedByDefault` prop. The second issue happens when the component is interacted with quickly after mounting. During this phase the references do not exist yet, so a couple checks were added to verify that the `this.content` reference is made before it is needed.

The example tests were written with [BigTest](https://github.com/thefrontside/bigtest), which currently powers all of our tests for [ui-eholdings](https://circleci.com/build-insights/gh/folio-org/ui-eholdings/master).  **Quick video of ui-eholdings tests in action:** https://youtu.be/4B0B29FSGds

### TODO:
- [ ] Set up CircleCI integration for this repo


JohnC-80:
The chance we'll need that harness is super slim... now if we were doing testing for stripes-smart-components, that'd be a different story... they're much more "hooked up."
I'd feel better about not having it up front, rather than something that requires too much outside of stripes-components itself. We really only need to add all that if it's absolutely necessary.

While a few of these components do depend on the stripes object from react context, for the most they just function on whatever props they get fed. Those that do depend on the stripes object in context could perhaps lean on the new context API within their own tests to provide it... so they'd have to worry less about what the Harness was doing and stay concerned with themselves. Components will need to be updated to the new API eventually, anyhow.

As for CircleCI - my ignorance on this level of things cries out. On the website they have cost plans, and an open-source discount, etc.. just wondering about potential future costs for this - if there are any - would just need to inform the right people. Say we adopt this FOLIO-wide - what costs would we be looking at then?

As for integration of CircleCI on the repo, @funkymalc?

cowboyd:
> I'd feel better about not having it up front, rather than something that requires too much outside of stripes-components itself. We really only need to add all that if it's absolutely necessary.

This seems totally reasonable. In fact, it's probably a warning flag if something that should be as context free as a component is coupled to that much, and if something is hard to test without the harness, then maybe that's a sign of a deeper issue that should be addressed.

It does seem like this _is_ similar to the harness that should eventually be in stripes-core though.

> As for CircleCI - my ignorance on this level of things cries out. On the website they have cost plans, and an open-source discount, etc.. just wondering about potential future costs for this - if there are any - would just need to inform the right people. Say we adopt this FOLIO-wide - what costs would we be looking at then?

Circle gives four free containers to each organization to use on OSS repositories. Based on our experience and the level of activity of the folio project, we would probably start to bump up against that limit pretty soon if this were adopted project-wide. However, it's more than plenty for three or four repos to start out with.

.circleci/config.yml
```diff
@@ -0,0 +1,73 @@
...
+
+  test-chrome:
+    docker:
+      - image: circleci/node:latest-browsers
+    steps:
+      - attach_workspace:
+          at: ~/
+      - run:
+          name: Run tests
+          command: yarn test --karma.singleRun --karma.browsers=Chrome
```
cowboyd: Should we add Firefox to the list since it can be run inside a linux container and won't require browserstack?

cowboyd: Also, I seem to recall that there was an issue with stripes-cli and passing through karma command line options that might be /cc @ttamj

ttamj: @cowboyd That has since been addressed.  Stripes CLI will now pass options to Karma given they are prefixed with `karma.` and in camelCase.  For example, Karma's `--single-run` is defined as `--karma.singleRun`.

wwilsman:
I removed the harness component, and pretender along with it since these components shouldn't ever need to talk to the network.

cowboyd:
@skoczko Sounds good to me.  I'd gladly be a resource in that discussion. I have quite a bit of experience with the various alternatives like Jenkins, Circle, and Travis, (and others) and can speak to the tradeoffs between them: from performance and scaling, to UX and even pricing.

julianladisch:
Cross-referencing "CircleCI for testing ui-* and stripes-*": https://issues.folio.org/browse/FOLIO-1204

PR #321, opened 2018-04-10
Layout updates to ConfirmationModal

rasmuswoelk:
Updated ConfirmationModal to match original designs. Overlay and close button fires onCancel callback.

## Before

image

## After

image


JohnC-80:
Maybe it's just me, but the close button would be a bit overloaded here. If it could be removed, I feel it's best if it isn't there. On a modal with strictly informational content, it makes sense as a way to dismiss things - but to have it exist on a confirmation modal where the system is requesting a definitive answer from the user, I'm not sure. We can say "yes the close button fires onCancel" but is it right to assume that's what the user would always want when they close it? In that case, the user answers the question without answering the question. @filipjakobsen?

zburke:
@JohnC-80, I confess to having the same concern that a modal with a "close" button isn't really a modal anymore, but IIRC that's often how Windows presents things and didn't want to reveal my true nature as a pedantic Mac fanboi.

JohnC-80:
If Windows really needs something, it'll ask you again... maybe THAT's our solution here... will you allow this to happen? 
{ } Yes  { } No

:tada:

filipjakobsen:
Hi folks,

Some modals need close buttons. Confirmation modals do not. So, we can remove the close button from any confirmation modal, but not from all modals in general.

Thanks,

PR #323, opened 2018-04-11
Add ability to pass custom filter to Selection component. Fixes STCOM-251.

mkuklis:
https://issues.folio.org/browse/STCOM-251


JohnC-80:
Thank you, @mkuklis!

PR #327, opened 2018-04-16
MCLRenderer: Added safety checks

doytch:
In certain scenarios, the data provided to the MCLRenderer during the process of several re-renders can get it into some trouble without these sanity checks. For example, the amount of items to render will momentarily become NaN since we get a 0 averageRowHeight. 

This is part of the fixes for UIU-466


JohnC-80:
Thanks! @doytch!

PR #328, opened 2018-04-16
add hidden filter config. Refs UIU-400

zburke:
Allow filters to have hidden configuration values. For example, suppose
a filter group "species" existed and the query should *always* include
humans and allow the user to choose whether Ewok, Jawa, or Wookie
members should also be included in the search. Ewok, Jawa, and Wookie
would be included in the "values" array of the config element with
name: "species" and Human would be included in the array of the
hiddenConfig element keyed by "species".

Refs [UIU-400](https://issues.folio.org/browse/UIU-400)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 * Use country-code to country-name mapping in `<AddressFieldGroup>`. Fixes STCOM-242. Available from v2.0.8.
 * MCL should use column-titles, not column keys, for the aria-label field. Fixes STCOM-246. Available from v2.0.9.
 * Allow for selecting value programmatically in Selection component. Fixes STCOM-250.
+* `<FilterGroups>` support hidden constraints. Refs UIU-500. Available from v2.0.10. 
 
 ## [2.0.0](https://github.com/folio-org/stripes-components/tree/v2.0.0) (2017-12-07)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v1.9.0...v2.0.0)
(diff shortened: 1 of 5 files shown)


MikeTaylor:
Can you easily show me how the config looks for your ewok example?

zburke:
```
// filters exposed to user as checkboxes
const filterConfig = [
  {
    label: 'Species',
    name: 'active',
    cql: 'active',
    values: [
       { name: 'Ewok', cql: 'ewok-unique-id' },
       { name: 'Jawa', cql: 'jawa-unique-id' },
       { name: 'Wookie', cql: 'wookie-unique-id' },
    ],
  },
];

// filters that are ALWAYS applied and that the user cannot control
const hiddenFilterConfig = {
  species: [
    { name: 'Human', cql: 'human' },
  ]
};
```

MikeTaylor:
Why not just
```
const filterConfig = [
  {
    label: 'Species',
    name: 'active',
    cql: 'active',
    values: [
       { name: 'Ewok', cql: 'ewok-unique-id' },
       { name: 'Jawa', cql: 'jawa-unique-id' },
       { name: 'Wookie', cql: 'wookie-unique-id' },
       { name: 'Human', cql: 'human', hidden: true },
    ],
  },
];
```
?

MikeTaylor:
Thank you. I suspect that once you have it in place, you'll find that the code delta is smaller, too.

BTW., don't forget to update `lib/FilterGroups/readme.md` when the code has settled.

PR #329, opened 2018-04-16
Stcom ignore time offset refs UIORG-55

aditya-matukumalli:
Added in the prop `ignoreLocalOffset`. When set to true, this props helps to parse the dates passed in from the server in UTC and display dates like the birthdates constant irrespective of the tenant level timezone setting.


aditya-matukumalli:
@JohnC-80 Thanks. Updated the CHANGELOG

PR #331, opened 2018-04-18
STCOM-199 MCL, Do not consider 0-length cell values for column measurement

JohnC-80:
If a table has a column with many empty cells, it may make conditions a bit cramped for any cells that do have content. This can lead to odd breakage in titles and a general lack of real-estate use.

### Before
column-with-averaged-0s

### After
0-lengths-discarded


rasmuswoelk:
Nice 👍

PR #332, opened 2018-04-18
STCOM-252 Add 'autofocus' prop to form controls

JohnC-80:
Focus management is a key part of accessibility and usability.  In this PR, applying the `autofocus` boolean prop to a form control will cause it to autofocus on its `componentDidMount` lifecycle method, or when a ref to the element is set.

This can be used to apply focus control to new Panes, Layers and Modals.
The components this affects:
* Button
* RadioButton
* Checkbox
* TextField
* Textarea
* Select
* Selection
* Datepicker
* Timepicker
* IconButton
* PaneHeaders (Pane action menu is autofocus-able)

### DIFF
--- a/lib/Button/Button.js
+++ b/lib/Button/Button.js
@@ -22,15 +22,19 @@ const propTypes = {
     PropTypes.arrayOf(PropTypes.node),
   ]),
   buttonRef: PropTypes.func,
+  autofocus: PropTypes.bool,
 };
 
 const defaultProps = {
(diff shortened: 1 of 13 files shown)


rasmuswoelk:
Nice 👍 Looks good but shouldn't the readme's be updated as well?

JohnC-80:
Closing for now... I think this is much simpler to do by passing react's handled `autoFocus` down to the right element. I'll re-approach with that implementation.

PR #342, opened 2018-04-24
Re-added ID for Checkbox input and htmlFor on label

rasmuswoelk:


zburke:
Can you revert this please? 

Unfortunately, we still haven't resolved the underlying issues that led to https://issues.folio.org/browse/UIU-430 (https://issues.folio.org/browse/STCOR-173 and https://issues.folio.org/browse/STRIPES-512) since the long discussion we had when UIU-430 was originally raised (see the full discussion at https://github.com/folio-org/stripes-components/pull/253/files/6782a4b8212fdcb50021beb9a0b77aadb16b2e6b#r175053405). 

STCOR-173 and STRIPES-512 are both low-priority issues at present. If we need to this change in place to move other design goals forward, let's talk to @cateboerema or @skoczko about reprioritizing them and pulling them into a sprint in the near future. 

Sorry for the trouble. I know this has been a long-standing goal.

zburke:
I know this change preserves the `<label><input ></label>` structure that we added in #270, but the specific issue here is that checkboxes in filters on overlays repeat the same element-IDs as those in the main search panel, which causes them not to function.

PR #349, opened 2018-04-27
Create TimeZone and Locale contexts

cherewaty:
## Purpose
Work for https://issues.folio.org/browse/UIORG-55 changed the behavior of `Datepicker`. Instead of emitting an ISO-8601 datetime string in the user's local time zone, it now emits the datetime at UTC.

Resulting bug in eHoldings: https://issues.folio.org/browse/STCOM-259

One of the APIs added to the `Datepicker` included `ignoreLocalOffset`: https://github.com/folio-org/stripes-components/pull/329. `ignoreLocalOffset` could easily create a race condition with the `timezone` peeled off of `stripes.context`.

All of these need tests. We should be able to more easily identify breaking changes.

## Approach
Testing context with the [new React Context API](https://reactjs.org/docs/context.html) is muuuuuch easier than before.

`Datepicker` and `Timepicker` were both tightly coupled to `locale` and `timezone` on the `stripes` context. By refactoring them to use the new Context API, we can make them both independent of `stripes-core` and significantly more testable.

`Datepicker` and `Timepicker` now simply have `locale` and `timeZone` props, that when inside of a `LocaleContext.Provider`/`TimeZoneContext.Provider`, get their value from context. The context can be overriden by the props.

I removed the `ignoreLocalOffset` prop. Instead, a `Datepicker` consumer can pass in the `timeZone` prop set to `UTC`.

### Context providers
Newly introduced: `Locale` and `TimeZone` higher-order components, that include creation of the context provider and consumer for `locale` and `timeZone`.

Modules consuming `stripes-components` can also use the consumer HOCs to receive props from context:
```
import withTimeZone from '@folio/stripes-components/lib/TimeZone';

function ExampleComponent({ timeZone }) => (
  <div>{timeZone}</div>
);

export default withTimeZone(ExampleComponent)
```

## Risk
Accessing `this.context.stripes.locale` and `this.context.stripes.timezone` should continue to work as previously, but we should find a way to deprecate those in favor of using `withLocale` and `withTimeZone`.

## Questions for later
- When in the context of a `redux-form`, `Datepicker` emits an ISO-860 string through `onChange()`, but when not in a `redux-form`, it fires an event whose `target.value` is the locale-formatted date in the input. Should those be more consistent? Should the `Datepicker` even try to reformat as ISO-860 by default?
- Should setting `moment.locale()` happen in a more global place? A lot of the date utility functions defined in `stripes-core` can go away in that case, since they're usually just wrapping `moment()`.

### DIFF
--- a/lib/Datepicker/Datepicker.js
+++ b/lib/Datepicker/Datepicker.js
@@ -7,6 +7,9 @@ import contains from 'dom-helpers/query/contains';
 import debounce from 'lodash/debounce';
 import uniqueId from 'lodash/uniqueId';
 
+import withLocale from '../Locale';
+import withTimeZone from '../TimeZone';
+
 import Button from '../Button';
 import TextField from '../TextField';
 import Calendar from './Calendar';
@@ -28,7 +31,6 @@ const propTypes = {
   onChange: PropTypes.func,
   useFocus: PropTypes.bool,
   hideOnChoose: PropTypes.bool,
-  locale: PropTypes.string,
   readOnly: PropTypes.bool,
   backendDateStandard: PropTypes.string,
   showCalendar: PropTypes.bool,
@@ -39,15 +41,15 @@ const propTypes = {
     PropTypes.string,
   ]),
   passThroughValue: PropTypes.string,
-  ignoreLocalOffset: PropTypes.bool,
   autoFocus: PropTypes.bool,
+  locale: PropTypes.string.isRequired,
+  timeZone: PropTypes.string.isRequired,
 };
 
 const defaultProps = {
   screenReaderMessage: '',
   hideOnChoose: true,
   useFocus: true,
-  locale: 'en',
   backendDateStandard: 'ISO8601',
   autoFocus: false,
   tether: {
@@ -67,12 +69,6 @@ const defaultProps = {
     },
     ],
   },
-};
-
-const contextTypes = {
-  stripes: PropTypes.shape({
-    locale: PropTypes.string.isRequired,
-  }).isRequired,
 };
 
 class Datepicker extends React.Component {
@@ -99,29 +95,26 @@ class Datepicker extends React.Component {
     this.handleFieldChange = this.handleFieldChange.bind(this);
     this.datePickerIsFocused = this.datePickerIsFocused.bind(this);
     this.getPresentedValue = this.getPresentedValue.bind(this);
-    this.timezone = this.context.stripes.timezone;
-    this.parseTimezone = this.props.ignoreLocalOffset ? 'UTC' : this.timezone;
 
-    moment.locale(this.context.stripes.locale);
+    moment.locale(this.props.locale);
     this._dateFormat = moment.localeData()._longDateFormat.L;
 
     // non-null presentedValue will eventually be rendered as the value of the TextField.
     let inputValue = '';
     let presentedValue = null;
 
-
     // if we aren't using redux form...
     if (typeof this.props.input === 'undefined') {
       if (this.props.value === this.props.passThroughValue && typeof this.props.date === 'undefined') {
         presentedValue = this.props.passThroughValue;
-        inputValue = moment.tz(this.parseTimezone).format(this._dateFormat);
+        inputValue = moment.tz(this.props.timeZone).format(this._dateFormat);
       }
       // if we are using redux-form...
     } else if (this.props.input.value === this.props.passThroughValue) {
       presentedValue = this.props.passThroughValue;
-      inputValue = moment.tz(this.parseTimezone).format(this._dateFormat);
+      inputValue = moment.tz(this.props.timeZone).format(this._dateFormat);
     } else if (this.props.input.value !== '') {
-      inputValue = moment.tz(this.props.input.value, this.parseTimezone).format(this._dateFormat);
+      inputValue = moment.tz(this.props.input.value, this.props.timeZone).format(this._dateFormat);
     }
 
     this.state = {
@@ -141,15 +134,15 @@ class Datepicker extends React.Component {
   componentWillReceiveProps(nextProps) {
     let inputValue;
     let presentedValue = null;
-    this.parseTimezone = nextProps.ignoreLocalOffset ? 'UTC' : this.timezone;
+
     if (nextProps.input) {
       if (nextProps.input.value !== '' && nextProps.input.value !== this.props.input.value) {
         const passRE = new RegExp(`^${nextProps.input.value}`, 'i');
         if (nextProps.input.value === this.props.passThroughValue || passRE.test(this.props.passThroughValue)) {
           presentedValue = nextProps.input.value;
-          inputValue = moment.tz(this.parseTimezone).format(this._dateFormat);
+          inputValue = moment.tz(this.props.timeZone).format(this._dateFormat);
         } else {
-          inputValue = moment.tz(nextProps.input.value, this.parseTimezone).format(this._dateFormat);
+          inputValue = moment.tz(nextProps.input.value, this.props.timeZone).format(this._dateFormat);
         }
         this.setState({
           presentedValue,
@@ -164,10 +157,10 @@ class Datepicker extends React.Component {
       }
     } else if (this.props.value !== '' && this.props.date !== nextProps.date) {
       if (nextProps.value === this.props.passThroughValue && nextProps.date === 'undefined') {
-        inputValue = moment.tz(this.parseTimezone).format(this._dateFormat);
+        inputValue = moment.tz(this.props.timeZone).format(this._dateFormat);
         presentedValue = this.props.passThroughValue;
       } else {
-        inputValue = moment.tz(this.parseTimezone).format(this._dateFormat);
+        inputValue = moment.tz(this.props.timeZone).format(this._dateFormat);
       }
       this.setState({
         presentedValue,
@@ -412,13 +405,13 @@ class Datepicker extends React.Component {
       case 'ISO 8601':
       case 'ISO8601':
       case 'ISO-8601':
-        return moment.tz(date, this._dateFormat, true, this.timezone).toISOString();
+        return moment.tz(date, this._dateFormat, true, this.props.timeZone).toISOString();
       case 'RFC 2822':
       case 'RFC2822':
       case 'RFC-2822':
-        return moment.tz(date, this._dateFormat, true, this.timezone).format(DATE_RFC2822);
+        return moment.tz(date, this._dateFormat, true, this.props.timeZone).format(DATE_RFC2822);
       default:
-        return moment.tz(date, this._dateFormat, true, this.timezone).format(this.props.backendDateStandard);
+        return moment.tz(date, this._dateFormat, true, this.props.timeZone).format(this.props.backendDateStandard);
     }
   }
 
@@ -439,7 +432,7 @@ class Datepicker extends React.Component {
       disabled,
       autoFocus,
       tether,
-
+      locale,
     } = this.props;
 
     const screenReaderFormat = this.cleanForScreenReader(this._dateFormat);
@@ -563,7 +556,7 @@ class Datepicker extends React.Component {
               ref={(ref) => { this.picker = ref; }}
               onKeyDown={this.handleKeyDown}
               onBlur={this.hideCalendar}
-              locale={this.context.stripes.locale || this.props.locale}
+              locale={locale}
               excludeDates={excludeDates}
               onNavigation={this.handleDateNavigation}
               id={this.testId}
@@ -576,7 +569,6 @@ class Datepicker extends React.Component {
 }
 
 Datepicker.propTypes = propTypes;
-Datepicker.contextTypes = contextTypes;
 Datepicker.defaultProps = defaultProps;
 
-export default Datepicker;
+export default withLocale(withTimeZone(Datepicker));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 3.0.0 (IN PROGRESS)
 
+* Create `Locale` and `TimeZone` contexts. Fixes STCOM-259.
 * In `<Datepicker>`, added a new `ignoreLocalOffset` prop that ignores the tenant timezone and treats the date as UTC to display the date. Fixes UIORG-55
 * Adjust address read only view. Fixes STCOM-152.
 * `<FilterPaneSearch>` supports `searchableIndexes`, `selectedIndex` and `onChangeIndex` properties. Fixes STCOM-171.
--- a/index.js
+++ b/index.js
@@ -12,6 +12,10 @@ export { default as TextArea } from './lib/TextArea';
 export { default as TextField } from './lib/TextField';
 export { default as Timepicker } from './lib/Timepicker';
 
+/* context providers/consumers */
+export { default as withLocale, LocaleContext } from './lib/Locale';
+export { default as withTimeZone, TimeZoneContext } from './lib/TimeZone';
+
 /* data containers */
 export { default as KeyValue } from './lib/KeyValue';
 export { default as MultiColumnList } from './lib/MultiColumnList';
--- a/lib/Datepicker/readme.md
+++ b/lib/Datepicker/readme.md
@@ -12,23 +12,21 @@ import Datepicker from '@folio/stripes-components/lib/Datepicker';
 ### Props
 Name | type | description | default | required
 --- | --- | --- | --- | ---
-label | string | visible field label | | false
-backendDateStandard | string | parses to/from ISO 8601 standard by default before committing value. | "ISO 8601" | false
-id | string | id for date field - used in the "id" attribute of the text input | | false
-useFocus | bool | if set to false, component relies solely on clicking the calendar icon to toggle appearance of calendar. | true | false
-autoFocus | bool | If this prop is `true`, component will automatically focus on mount | | 
-disabled | bool | if true, field will be disabled for focus or entry. | false | false
-ignoreLocalOffset | bool | if true, ignores the timezone setting and treats the date as UTC to display the date.
-| false | false. See below for more explanation
-readOnly | bool | if true, field will be readonly. 'Calendar' and 'clear' buttons will be omitted. | false | false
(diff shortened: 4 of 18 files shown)


lib/Datepicker/Datepicker.js
```diff
@@ -39,15 +41,15 @@ const propTypes = {
     PropTypes.string,
   ]),
   passThroughValue: PropTypes.string,
-  ignoreLocalOffset: PropTypes.bool,
   autoFocus: PropTypes.bool,
+  locale: PropTypes.string.isRequired,
+  timeZone: PropTypes.string.isRequired,
```
JohnC-80: `timeZone`, while it works adequately with the React Context API, seems a little too open-to-suggestion for Datepicker. A developer will probably only ever have to pass 'UTC' to this prop, right? Any reason why we would want it to be open to **any** developer provided timezone?

cherewaty: Use case for having a overridable `timeZone` prop for `Datepicker`: events. The user might have date and time pickers that they want to convert to another time zone just for the purposes of creating the record.

Example from Google Calendar:

JohnC-80: That seems viable to me.

JohnC-80:
For a hypothetical situation regarding the HOC, there wouldn't be any way for a wrapped component to apply its own component-level propType validation right? It has to use whatever the upper-level context creator uses - is that correct? It would have to be an independant wrapper that applied those validations to its developer-supplied props before overriding them with context- supplied values, right?

cherewaty:
@JohnC-80, if I understand your question correctly, the `Datepicker` still has `PropTypes` defined for `locale` and `timeZone`. If an HOC like `withTimeZone` tries to pass in a `timeZone` that's not a string, the `Datepicker` `PropTypes` validation will fail as usual.

cherewaty:
Last week @skomorokh mentioned a possibility to me: what if we created a `CurrentUserContext` that contained all the attributes currently supplied by `context.stripes`?

One of the advantages of the new context API is explicit subscriptions. `Consumer`s only subscribe to the `Provider`s they care about. Whenever the `Provider` value changes, the `Consumer` re-renders.

That makes it preferable to take the approach used here, where each attribute that used to live on `context.stripes` is now served by its own context `Provider`. Otherwise lots of components would go through unnecessary re-renders, since they'd be subscribing to an entire `CurrentUserContext`, not just the values they specifically need.

Having a `Provider` for every context attribute could make for a particularly verbose `render()` in `stripes-core`'s `Root.js`, but should provide more clarity and better render performance everywhere else.

MikeTaylor:
Folks,

I don't want to see a proliferation of separate contexts, partly for conceptual reasons and partly for the pragmatic reason that I don't want to have to change a bunch of client code. I recognise that the new React context API(*) means we're going to need to stop using the old one when we reach React 17, but the simple upgrade path is just to provide a single context that furnishes the existing Stripes object, exactly as it presently is. Let's do that now; if we later decide we additionally want some more fine-grained contexts within that, we can fault them in as needed.

(*) A pox on the React developers, and indeed _all_ JS developers, for their idiot habit of constantly churning their APIs.

skomorokh:
There is another aspect of this I want to touch on:

One of the benefits we get from having `stripes-components` as its own package is that apps can depend on different versions of it. While we ought rarely make breaking changes to component APIs, it is a nice quality that we can do so and, temporarily, let old modules bring in prior versions. This way we don't have to coordinate an update system-wide.

The part of the new context app I find most disturbing is the reliance on every subsequent import of the context getting a reference to the object created on the first import. Effectively a webpack-mediated global variable. Much like React and Highlander, this means we can only have one. So I think we're best off importing the context from `stripes-core` or some new package that is consistently referred to as a peer dependency and not one where we want to maintain it as a regular dependency and tolerate multiple instances.

Maybe now we can finally establish a coherent practice for such peer deps--does the CI install them? Or always use stripes-cli to test? Or we cross-list them as dev dependencies?

cowboyd:
As I understand it, the goal of stripes components is to be a component library and design system, which means that for components to be _composable_, they need to make as little assumption about their containing environment as possible. Higher coupling = Lower composability = less component, more code util. All things being equal, shouldn't we opt for higher composability in our component library? 

In this instance for example, the concept of a user feels like an application concern, not a component concern. The component, a datepicker, needs a timezone and locale to do its job. Why couple that to a user? 

It's not to say that the coupling won't happen somewhere. Clearly, it must, but it's a question of _where_. Isn't the most logical place to define the coupling between user, timezone and locale the very same place that defines the concept of user in the first place? Why duplicate this coupling across both stripes-core _and_ stripes-components?

Among other things, It seems to me a very plausible use case to want to render a date picker without a user.

> I don't think the verbosity will be restricted to Root.js as the metadata about the current user expands as we add richness to user preferences:

There's nothing saying that a `CurrentUserContext` shouldn't exist, only that it exists as a platform concern, is defined by the platform, and it composes any number of subcontexts.

> partly for conceptual reasons and partly for the pragmatic reason that I don't want to have to change a bunch of client code

If the concern here is backwards compatibility, then I think that's definitely doable that we can have a way forward without breaking existing code.

MikeTaylor:
> As I understand it, the goal of stripes components is to be a component library and design system, which means that for components to be composable, they need to make as little assumption about their containing environment as possible. Higher coupling = Lower composability = less component, more code util.

It's _Stripes_ components. These components run in the context of Stripes. And Stripes gives you the Stripes object.

cowboyd:
> The part of the new context app I find most disturbing is the reliance on every subsequent import of the context getting a reference to the object created on the first import. Effectively a webpack-mediated global variable. Much like React and Highlander, this means we can only have one.

I don't think I understand. Doesn't the same apply to a function or any other constant defined within a module?

```js
// there can be only one (function named createSomething in this module)
export function createSomething() {
//....
}
// there can be only one (constant named CREATE_SOMETHING in this module)
export const CREATE_SOMETHING = "CREATE_SOMETHING";

//  there can be only one (constant named SomethingContext in this module)
export const SomethingContext = React.createContext();
```

It's just a constant that identifies a context, so that it can be consumed and provided.

cherewaty:
So that we're all on the same page about what this would look like, here are some gists.

### Scenario 1
https://gist.github.com/cherewaty/473f8e5e352b38afeea37e8086e780a7
- `CurrentUserContext`
- `withCurrentUser()` HOC that passes in whole `currentUser` object

Basically a direct translation of the current `context.stripes` strategy to the new Context API. End components have to know the structure of the `currentUser` object (it's `stripesShape` all over again). To test a component in isolation, it has to receive a prop whose `currentUser` is the right shape.


### Scenario 2
https://gist.github.com/cherewaty/f60e7294639a2bf9f5139f090ea58c14
- `CurrentUserContext`
- `withCurrentUser()` HOC that spreads props

Slightly better than scenario 1, since end components don't have to know about a `currentUser` object, only now they're being sent props they didn't ask for. Could have strange side effects.


### Scenario 3
https://gist.github.com/cherewaty/3426aac7fcb4286e637858fcc11697c7
- `CurrentUserContext`
- `withTimeZone()` HOC

Slightly better than scenarios 1 and 2, since the end component only receives the `timeZone` prop, but it's still subscribed to the entire `currentUser` context. That will cause a re-render when anything in `currentUser` changes, not just `timeZone`.


### Scenario 4
https://gist.github.com/cherewaty/671afc60ebf250aff6cb18fcf8c2025d
- `TimeZoneContext`
- `withTimeZone()` HOC

What's proposed in this PR. Components subscribe _only_ to the context they need, and can exist independently of any context. `Root.js` will end up with lots of providers, but that can be refactored to keep the list of providers isolated and readable.

skomorokh:
As I brought up a time or two lately, I've come to understand that `stripesShape` is a misuse of proptypes. I think we'd derive more benefit from proptypes if they were used to delineate exactly the props that are actually consumed. If you only need timezone, specify `CurrentUser` as an object with only a `timeZone` property and your test is no harder to build.

And current user metadata is far from all that is in the `stripes` context or will eventually be, I was not proposing moving everything from stripes to it--I don't mind splitting it out into a few contexts (in addition to the user perhaps one for the current app, tenant, Okapi, etc.) But a context for every leaf property seems untenable. It's "lots of providers" to the extent that I anticipate dozens once our configuration is richer. To be fair to Mike, I struggle to justify splitting it at all but do see the benefits both in organisation and communicating something more specific than "stripes". 

With all the HOCs the component tree is already extremely nested and thus less useful for debugging and understanding the app than it might otherwise be. Perhaps the devtools are being improved to better collapse HOCs together?

(Sorry Charles, I have to run, I'll get to your comment on Thursday)

cowboyd:
>And current user metadata is far from all that is in the stripes context or will eventually be, I was not proposing moving everything from stripes to it--I don't mind splitting it out into a few contexts (in addition to the user perhaps one for the current app, tenant, Okapi, etc.) But a context for every leaf property seems untenable.

Why is this untenable? It results components that are simpler, performant by default, and most important, more _composable_. Also as a clear benefit, there's a plain, one-way dependency from stripes-core to stripes-components instead of the circular one we have now. 

In the days when providing and consuming contextual data in React involved a ton of ceremony, rolling our own ad-hoc context layer made more sense. We weren't the only ones though, and the new context API was rolled out precisely as a countermeasure against the accretion of global contexts like the Stripes object that reduce performance, reactivity, and composability.

> With all the HOCs the component tree is already extremely nested and thus less useful for debugging and understanding the app than it might otherwise be. Perhaps the devtools are being improved to better collapse HOCs together?

I was looking at our application yesterday in the react devtools, and our component tree is well-nigh unusable as it is, but it does look like [this discussion has been going on for a very long time](https://github.com/facebook/react-devtools/issues/604) within the React community and that [help is very soon on the way](https://github.com/facebook/react-devtools/pull/997). The fact that there has been so much interest in this feature, and so much involvement from the core team is a strong indicator that this is considered a good practice. Given that it appears that the devtools are moving in this direction and so it shouldn't be a factor in our decision.

JohnC-80:
I have not-so-warm feelings about exposing `hasPerm` and `hasInterface` as props... so that's 2 more we can scratch off - we'll have to figure out something else for those eventually.

MikeTaylor:
> The React folks warned us for a while that the Context API was bound to change.

*Act 1: 2016*
Me: "I might stab you."
You: "Well, I'd rather you didn't."

*Act 2: 2017*
Me: (stabs you)
You: "Ouch!"

*Act 3: later in 2017*
Judge: You are accused of stabbing Mr. Coburn. How do you plead?
Me: guilty, but I did warn him that I might stab him.
Judge: Oh, OK then. Case dismissed.

skomorokh:
I do find the idea of a more honest API where consuming components aren't coupled to unnecessary application detail compelling. The part where I find this untenable is that, by the time we have fleshed out the metadata for the tenant, the application, the user, the app, okapi, etc. we will have several dozen properties in play. Let's say a nice round 100. 

While it feels a bit dangerous to assume, I imagine there isn't really a huge performance impact to 100 wrappers as React is heavily optimized. And if that PR Charles linked to is likely to be merged, we can tidy up the dev tools. Ifs, but not such big ifs as the zeitgeist is blowing this way.

But I'd like to hear more about how we plan to manage providing them. Because at the moment I envision 100 near-identical `FooContext.js` files each with their own `withFoo` export and their own invocation as a wrapper in Root.js.

The maintainability issue there feels like it would take some heroics to tame. Flattening the application detail out of the context props seems comparatively trivial. It seems like the main benefit isn't from each of these being separate contexts (as the rendering issue is moot for the vast majority) but rather from well-specified component APIs enabled by the granularity of having `withFoo()`.

Can't we get that in other ways?
```
import { fromUser } from 'stripes-core/contexts/user';

class Whatever extends component { ... }

export default fromUser(['locale', 'timezone'])(Whatever);

```

MikeTaylor:
All of this feels like a ton of work to yield an effect whose value is very close to zero -- whether slightly positive or slightly negative is not completely clear. Don't we have a million other, more important and urgent, things to be doing? I'm reminded of the zinger at the end of #8 in Henry Spencer's _Ten Commandments for C Programmers_: "thy creativity is better used in solving problems than in creating beautiful new impediments to understanding". That observation must be 30 years old at the very outside, but it's as true now as it ever was.

skomorokh:
@cowboyd what I mean by there can only be one is that there can only be one copy of `FooContext.js`. I like that we can currently use a module that depends on an old version of `stripes-components` and it just brings in a second copy at the old version and more or less works (though maybe the styles look dated or something). 

That makes upgrades a whole lot less painful because we don't have to coordinate new versions of everything at once. And the extra copies are tolerable with async loading. If we store the context in `stripes-components` we lose this. Unless it's really just a static string like your `CREATE_SOMETHING` example and you can consume a context from a different file than the one which produced it, provided it's named the same?

Whereas, if it lives in `stripes-core` where it's primarily provided from, well, we already have to have only one copy of that. It does weird the dependency graph though and I've actively been trying to avoid having things import from core. However, I'm not super stoked about a new package for it either... I guess `stripes-contexts` wouldn't be the end of the world (but it might confuse people to think that it's not okay to have their own contexts within their app tree and otherwise make use of the mechanism when they certainly should be free to).

skomorokh:
@MikeTaylor this is important in the short term for better testability. Tests are important. They really are. In the long term, we need to revisit this anyway because our tools are improving underfoot. Context forms one of the main APIs consumed by Stripes developers and so it's well worth us putting some thought into how we structure it. It also improves the API for the myriad components which consume it by making it clear what they need and thus facilitating reuse.

cherewaty:
@MikeTaylor  "This isn't my priority, why is it yours?" is an unacceptable response to contributions in an open-source community. The project is strengthened by contributors whose diverse priorities converge on a solution that benefits all.

Snarky attempts to shut down constructive collaboration should not, and will not, be tolerated in this community.

MikeTaylor:
@cherewaty It's very considerate of you to tell us what will and will not be tolerated in our own community.

The "solution" that you propose is one that will have implications for all of us, changing code that presently works. I reject your claim that I'm not allowed to object to that.

And with that, I am out of this conversation. You guys can sort it out between you, and do whatever rewriting of the existing code may be necessary as a result.

JohnC-80:
Guys, let's stay on some task here. @MikeTaylor I think there's awareness of the implications there - @cherewaty did note the risk in his description. That's why these PR's exist so that we can talk through these things and find solutions to whichever concerns there are. While it's true that we've got a scope that can drown things in priority, this pattern of context affects how we'll do things going forward. We can't keep adding keys to the old context and behave like the new API isn't a thing we should be concerned about. Ideally, we find positives in producing code with less dependencies on things like the shape of an object that's produced higher in the component tree as well as the niceness of testing components as it's all props now, rather than injected outsider values that have to be shimmed together in some sort of harness in order for testing to even happen properly. Yes, it's a good deal of work to turn the battleship, and it would be nice to find ways to mitigate the workload to hopefully set us up for fewer occurrences of it in the future.

JohnC-80:
@skomorokh Yeah, there's gotta be a better way to do it than having custom HOC's for every key on the contextualized object. It would be nice to have a way for the component to pass through the particular pieces of context it needs in order to receive those as its props... hmmm.

cowboyd:
@skomorokh 

>That makes upgrades a whole lot less painful because we don't have to coordinate new versions of everything at once. And the extra copies are tolerable with async loading. If we store the context in stripes-components we lose this. 

Wouldn't it actually do a _better_ job at achieveing api compatibility? The reason is because the version of the context you need is bundled with the _same version_ of the `stripes-components` library that consumes it. Note that contexts like user, institution, physical locations would not be part of `stripes-components`, because that's not its concern. For example,  Let's say all the sudden a new version of stripes components comes out that moves the timezone context definition somewhere else. Well, since it both declares _and_ consumes the timezone, if you try to upgrade a library that used the old constant, it will break. Precisely the behavior you want.

> Whereas, if it lives in stripes-core where it's primarily provided from, well, we already have to have only one copy of that. It does weird the dependency graph though and I've actively been trying to avoid having things import from core. However, I'm not super stoked about a new package for it either... I guess stripes-contexts wouldn't be the end of the world (but it might confuse people to think that it's not okay to have their own contexts within their app tree and otherwise make use of the mechanism when they certainly should be free to).

I'm 100% behind the idea that the _actual_ composition of the contexts lives in `stripes-core`, just as it does now. The only difference is that `stripes-components` defines only the primitives that it consumes. It says "Hey, if you want to pass me one of my props as a piece of context, just use this class. How (or even _if_) you choose to do that its totally your business".

So for example `UserContext` would be defined by stripes-core because it has the concept of a User (that would be a composition of locale, timezone, etc...). On the other hand, `stripes-components` which does _not_ truck in users, would be remain in blissful ignorance.

It's really the same idea as composition of simple objects. JavaScript gives us Strings, Numbers, Dates and things which we can use to build Users, Addresses, Cars, what have you. The primitives live in one place, and then the app uses them to express the higher level concepts that it works with.

The guiding rule I'd see of where these things would go would be, if you define the concept, supply the context.

JohnC-80:
@cowboyd I think we would want to define these **global** contexts and their components in core or as high as we can to get a single source of the component pairs that a single call to `React.createContext()` generates. The behavior where the `<Provider>-<Consumer>` pair comes from each single call to `React.createContext()` creates concern if we have those pairs generated from multiple versions of `stripes-components` present in the system - `stripes-core` will only be serving its contextual value to a single instance of `<Provider>`.
If two modules exist in the system with differing versions of stripes-components (each setting up its own call to `React.createContext()`, I'm afraid it'll be a race as to which <Provider> gets served a value first to feed its coordinating `<Consumer>`. The `<Consumer>` from the pair on the other version could get left in the dark.

cowboyd:
@skomorokh @JohnC-80 I see what you're saying now. I had not accounted for having multiple versions of stripes-components being used at the same time.

In light of this, I think we should back away from the context portion for the moment, and just focus on the props.

cherewaty:
Closing, since large changes will need to be made based on the conversation moved to https://issues.folio.org/browse/STCOR-208. Thanks @JohnC-80 for moving that over and pushing this forward.

Thanks @skomorokh for pointing out that every module can have its own `stripes-components` version. That's definitely the dealbreaker for this approach - `stripes-core` will need to continue being the one creating context.

PR #359, opened 2018-05-03
folio-1035

funkymalc:
- Support browser-based (i.e karma) tests in Jenkins via 'yarn test'.   Currently supported browser is Chrome.  See top-level 'karma.conf.js'.  Can be extended to support Firefox easily. 
- Publish 'yarn test' report 
- Add configurable parameter 'RunTestOptions' to Jenkinsfile to pass additional options to 'yarn test' to  specify karma parameters.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,3 @@
-@Library ('folio_jenkins_shared_libs@folio-1035') _
 
 buildNPM {
   publishModDescriptor = 'no'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,8 @@
+@Library ('folio_jenkins_shared_libs@folio-1035') _
 
 buildNPM {
   publishModDescriptor = 'no'
-  runLint = 'no'
-  runTest = 'no'
+  runLint = 'yes'
+  runTest = 'yes'
+  runTestOptions = '--karma.singleRun --karma.browsers=ChromeDocker'
 }
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,0 +1,17 @@
+module.exports = (config) => {
+  let configuration = {
+    browsers: ['ChromeDocker'],
+  
+    customLaunchers: {
+      ChromeDocker: {
+        base: 'Chrome',
+        flags: [
+           '--no-sandbox',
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -1,6 +1,7 @@
 
 buildNPM {
   publishModDescriptor = 'no'
-  runLint = 'no'
-  runTest = 'no'
+  runLint = 'yes'
+  runTest = 'yes'
+  runTestOptions = '--karma.singleRun --karma.browsers=ChromeDocker'
```
JohnC-80: Thanks for your effort on this, John... so initially if I attempt to run this locally - it doesn't since I don't have a) Docker installed or b) a docker image with Chrome Headless.
Since we don't want to require those for local testing, I think the solution may be 
a) rename `karma.conf.js` to `karma.ci.conf.js`
b) point the runTestOptions at the config file so that it says 
```karma.ci.conf.js --karma.singleRun --karma.browsers=ChromeDocker```
Give it a shot?
Since locally it spins up Karma through the `stripes-cli` it just uses the stripes-cli's karma config, which spins up the local chrome runner, etc... so everything's there without the config living in stripes-components for the time being. Later if we need our own we'll make one, but nice to differentiate between local and ci versions in the name, methinks.

funkymalc: Thanks,  John C -  

Couple of points/questions: 

- I don't think it's possible to specify an alternate karma config as a command line option like that.  At least, I couldn't get it to work. 

- Even with that top-level karma config "as is",  you should still be able to run the following locally as you did before (without Docker, etc): 

'yarn test --karma.singleRun --karma.browsers=Chrome'
 
- I AM wondering if I should just inject the karma config that has the ChromeDocker settings for CI during the build rather than maintaining in each project's repository.  However, my fear is that if there is an existing, customized karma.conf.js at the top-level of a project (see ui-eholdings,  as an example) this method would overwrite it.      Not sure yet if that's a bad thing necessarily.  Actually,  that may be the way to go.

funkymalc: I'm going to proceed with removing the CI karma.conf.js from the repository and injecting it during the CI build.   Closing this PR.  Will open new PR with changes.

PR #367, opened 2018-05-08
add test harness for context redux-form and i18n

JohnC-80:
Early on we added a test harness that contained all of the stripes context among a great deal of other things that might be needed to test components (and a fair amount that probably wouldn't be needed.) As we move forward testing components with dependencies using pre-16.3 context (react-intl and redux-form) a setup for these does need to be added into a universal, minimal test harness of some kind.


tests/helpers.js
```diff
@@ -23,7 +24,7 @@ function getCleanTestingRoot() {
 
 export function mount(component) {
   return new Promise(resolve => {
-    ReactDOM.render(component, getCleanTestingRoot(), resolve);
+    ReactDOM.render(<Harness>{component}</Harness>, getCleanTestingRoot(), resolve);
```
wwilsman: Maybe we can have a `mountWithContext` helper instead of doing this for every component, even those that do not need it at all.

In the future, when the components that **need** this are _truly_ reusable, we can find the ones that rely on this context and replace `mountWithContext` with the new Provider-Consumer context API.

JohnC-80: Good deal - changes applied. Thanks @wwilsman

PR #368, opened 2018-05-11
STCOM-253 use context instead of injectIntl

rjruizes:
I previously used injectIntl to provide the intl object. Context is a less intrusive way to do the same thing.


cowboyd:
In approach, these two changes are related, and it seems like we need to come to some form of consensus around how we pass data to stripes components before we merge this. 

https://github.com/folio-org/stripes-components/pull/349

I believe based on the conversation around that PR, and also around conversations at WolfCON that we're leaning towards dumber components.

cowboyd:
> Yes, there is a broader issue going forward, but I don't think that needs to delay the present issue, which IIRC Roman needs to fix an actual bug.

I disagree that urgency of a bug fix is a good reason to prematurely merge something into our shared infrastructure that is _very_ likely to be changed in the very near future. That only exposes the code that depends on this change to a higher risk for re-work. 

For example, the issue I referenced before, #349, also pertained to a bug fix, but I'd rather work around it and get the shared components right the first time, then push something through and end up with more shared work for everybody.

@rjruizes If this is urgent, then I'd recommend a simple enough work around within the application or module in which you're operating, and not in the shared infrastructure. You can just implement a component whose sole responsibility is to peel the `intl` object off of the Stripes context, and then pass it along as a prop to the calendar, datepicker object (or whatever else) without requiring a change in those components.

```js
// IntlContext.js
export default class IntlContext extends Component {
  static contextTypes = {
    stripes: PropTypes.shape({
      intl: PropTypes.shape({
        formatMessage: PropTypes.func.isRequired
      }).isRequired,
    }).isRequired,
  };
  render() {
    return this.props.children(this.context.stripes.intl);
  }
}
```

And then in your code:

```js

export default SomeOtherComponent() {
  return (
  <IntlContext>
    {intl =>
       <Calender intl={intl}/>
       <Datpicker intl={intl}/>
    }
  </IntlContext>
}
```

This has the advantage of solving your immediate concern, being mostly forward compatible with the React 16 context api, and also doesn't introduce a dependency in your app to an api that could be deprecated very quickly.

skomorokh:
I don't think the old context API is going away very quickly, but that's definitely the direction we'll eventually need to go and I agree that it's cleaner to have the majority of components simply get exactly what they need from props.

If there is no problem with the existing injectIntl maybe just keep things the way they are as it's letting you keep intl as a prop so it'll fit in with what we're planning.

MikeTaylor:
OK, then I will stay out of this PR and leave it for you to resolve. Thanks.

-- Mike.

rjruizes:
@cowboyd What you’re suggesting is what injectIntl does, wrapping the component. The disadvantage of that is it changes the ref to the wrapper component. The TextField component often uses the ref to apply focus() to the input element.

cowboyd:
To be clear, I also think we should get rid of injectIntl as I think it ultimately suffers from the same problem as being coupled to the stripes context.

> The disadvantage of that is it changes the ref to the wrapper component. The TextField component often uses the ref to apply focus() to the input element. So we have to check all usages of injectIntl elements to make sure we update any refs used.

Thanks for this. It appears that I'm not fully understanding the problem then. Is there any way you can show some code samples of where the conflict arises? Maybe we can use them as a starting point for discussion.

rjruizes:
@cowboyd Here's an example from https://github.com/folio-org/ui-checkout/pull/125:
```
-    const input = this.barcodeEl.getRenderedComponent().input;
+    const input = this.barcodeEl.current.getRenderedComponent().refs.wrappedInstance.input;
```
injectIntl causes the ref to refer to the wrapper, and it must be updated to refer to the wrapped instance.

rjruizes:
I'm going to add a stripes provider to the test harness so the tests pass and this can be merged.

Robdel12:
> The TextField component often uses the ref to apply focus() to the input element.

Is this PR to work around that specific bug? Would using the new `autoFocus` prop help with this? https://github.com/folio-org/stripes-components/pull/341I'm looking through the PR you linked to and I'm having a hard time figuring out how / why the focus is being set. Could you capture how that works in the UI with a GIF? (I tried cloning the repo but there's no docs on how to run the app)

rjruizes:
Originally I used injectIntl to provide the intl object. Then I realized the stripes object was available via context API, and I made this PR. Then I realized intl itself was available via context (doh) and is how the react-intl library retrieves the `intl` object for its own components (`<FormattedMessage>`).

rjruizes:
- The conversation @cowboyd mentioned from #349 was about using separate providers instead of a single `stripes` provider. With my latest changeset, I no longer used `stripes`. Instead, I take advantage of `intl` already being on context, which is how the `react-intl` library retrieves it all across our project.
- lint errors: Unrelated warning. I can fix in a separate PR.
- I just documented this info in our [i18n doc](https://github.com/folio-org/stripes-core/blob/master/doc/i18n.md#intl-object).

cowboyd:
My fundamental concern here though is about moving from a pure props-based API, to a dynamic context-based API. The new context API is not just about being different, it's about the React team's attempt to facilitate better component design by always passing data to components explicitly through props. This makes them more composable and therefore more generally useful since they are (at least at the syntax level), pure functions.

For context, in #349, we're actually stepping away from the idea of adding custom providers inside stripes-components. The discussion there pointed out that this is probably not the best idea. Instead, we're  just going with the idea that prop is prop, and where it is sourced from is an application level concern.

cowboyd:
This isn't to say that `ref` handling is not currently broken. It sounds like it definitely is, I just want to explore if we can fix that while at the same time keeping our components props-oriented and forward compatible.

JohnC-80:
To clean up the ref issue with <TextField> and the components that use it: https://issues.folio.org/browse/STCOM-267

JohnC-80:
As for other parts of this PR... we're tripping over hastiness, here. Seems the broken refs were the cause but the solution was approached differently (opening comment says nothing about ref breakage as symptom.) Which caused some back-sliding into old-context usage. I do hope this can be sorted with some contributor/reviewer guidelines (PR template) in place and some possible tweaking of github settings (changes invalidate approval.) It works for now... but there's still other work left to be done.
An issue surfacing here is how to handle dependencies that lag in upgrading to the new context since it's still new. I'm glad that the presence of the new context doesn't break the functionality of the old. +1 React...

PR #370, opened 2018-05-15
Adjust Selection component to support dynamic data options. Part of STSMACOM-94

mkuklis:
@JohnC-80 this PR includes couple small changes to support https://issues.folio.org/browse/STSMACOM-94.

The locations in STSMACOM-94 are being pulled from the server when institution, campus or library is being changed. I was running into couple issues with `Selection`:

1. The new values were never populated
2. The `Selection` was throwing exceptions when `dataOptions` was empty
3. The filter input would stay read only (not sure why this was happening but I noticed that everything is working well when `TetherComponent` has just one child).

### DIFF
--- a/lib/Selection/SelectList.js
+++ b/lib/Selection/SelectList.js
@@ -135,7 +135,12 @@ class SelectList extends React.Component {
     if (this.props.selected) {
       return `option-${this.props.selectedIndex}-${this.props.selected}`;
     }
-    return `option-0-${this.props.list[0].value}`;
+
+    if (this.props.list.length) {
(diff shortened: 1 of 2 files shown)


mkuklis:
@JohnC-80 good idea. Yeah let's see how many items we get and if we run into any performance issues. The location list is pretty short but we may use `Selection` for longer lists.

MikeTaylor:
Surely `lodash`'s `isEqual` implementation will stop comparing as soon as two elements are different. I don't think we need to go second-guessing this kind of thing. Micro-optimisations are for 1980s C programmers :-)

PR #372, opened 2018-05-16
STCOM-269 Refactor Checkbox with reduxFormField()

cherewaty:
https://issues.folio.org/browse/STCOM-269

## Purpose
I noticed @rasmuswoelk created a test for the Checkbox component a few weeks ago - I thought it would be nice to move those assertions over to a test at `lib/components/Checkbox/tests` with a BigTest interactor. What I discovered in that process was that the `Checkbox` component had several bugs when not used in a [Redux Form](https://redux-form.com/7.3.0/).

Most `stripes-components` that are intended for use in a Redux Form `Field` have code resembling this:
```
handleChange(e) {
   if (this.props.input && this.props.input.onChange) {
     this.props.input.onChange(e);
   }

   if (this.props.onChange) {
     this.props.onChange(e);
   }
 }
 ```
Because the logic dealing with a component's behavior in a Redux Form vs. when it's independent is spread all over the component, it makes debugging behavioral differences very difficult.

## Approach
We can simplify component APIs by moving the handling of Redux Form `Field`'s `input` and `meta` props up a level.

With a `reduxFormField` higher-order component, we can do something like this:
```
import reduxFormField from '@folio/stripes-components/lib/ReduxFormField';

function ExampleComponent({ value, onChange, warningText, errorText }) => (
  <div>{warningText}</div>
);

export default reduxFormField(
  ExampleComponent,
  ({ input, meta }) => ({
    value: input.value,
    onChange: input.onChange,
    warningText: (meta.touched && meta.warning ? meta.warning : ''),
    errorText: (meta.touched && meta.error ? meta.error : '')
  })
);
```

Now a component like the `Checkbox` only needs to deal with a single `onChange()` prop, not both `onChange()` and `this.props.onChange()`.

## Impact
This change shouldn't affect `stripes-components` consumers - there are no intentionally breaking changes. No changes are needed in `stripes-core`, modules, etc.

The `Checkbox` was the proof-of-concept for this idea. Other components can follow as we get to them.

### CODE REVIEW

**Comment by JohnC-80** [comment]
Thanks, @cherewaty, this is great! It will clean up things significantly.  I've wanted a better separation between the redux-form and vanilla versions for a while. An HoC seems totally obvious there (even the redux-form docs kindof point in that direction with the way to make redux-form work with whatever component).

File: lib/Checkbox/Checkbox.js [hunk 1]
```diff
@@ -83,19 +68,14 @@ class Checkbox extends React.Component {
...
 
   // Add feedback class on warning or error
-  getFeedbackClasses() {
-    const meta = this.props.meta;
-    if (typeof meta !== 'object') {
-      return [];
-    }
-
-    const { warning, error, touched } = meta;
+  getFeedbackClasses = () => {
```
JohnC-80: Regarding these methods - why switch from a bound method to an arrow function? I've seen [this article](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1) about the difference there and arrow functions not associating themselves at the Prototype of the class - any concern to be had there?

cherewaty: Interesting. I had no idea of the caveats of the arrow functions vs. binding in the constructor. Since that's not really the aim of this PR, I rolled that part back to its original form.

lib/Checkbox/Checkbox.js
```diff
@@ -166,58 +128,48 @@ class Checkbox extends React.Component {
...
+    } = this.props;
 
-    const { label } = cleanedProps;
     const [, inputCustom] = separateComponentProps(this.props, Checkbox.propTypes);
 
     const ariaProps = {};
-    if (!this.props.name || !this.props.input) {
-      ariaProps['aria-label'] = `checkbox ${this.props.label}`;
+    if (!name) {
+      ariaProps['aria-label'] = `checkbox ${label}`;
```
Robdel12: We shouldn't say the role of the element in the label. 

Screen readers (and other assistive tech) rely on _role_, _name_, & _value_ to give context what an element is and how to interact with it. For this specific example role, name, & value plays out like this:
- Role: Checkbox (inferred from it being an input with `type="checkbox"`)
- Name: `${label}` (If there is an `ariaLabel` on the input this always will override the linked `<label>` element)
- Value: `checked` (is it checked or unchecked?)

lib/Checkbox/Checkbox.js
```diff
@@ -1,40 +1,45 @@
...
-  disabled: PropTypes.bool,
-  autoFocus: PropTypes.bool,
-};
-
 class Checkbox extends React.Component {
+  static propTypes = {
+    label: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.node,
+    ]),
```
Robdel12: Wondering if this should become a required prop. What is the use case for _not_ providing a label to a checkbox? I can't think of any.

JohnC-80: The case might be if it's using `aria-labelledby` for some sort of column-of-checkboxes scenario - then it wouldn't require a label. That case holds a few components back from having labels/id's required.

Robdel12: Hm in that case they should still be labeled but then use a `fieldset` with a `legend` to commonly group them. 

I might be missing the example though :P Agreed it's too big of a change but I wanted to raise the flag here

cherewaty:
Looking forward to applying this pattern to other components as we add test coverage 😁

cherewaty:
@cowboyd:
1. Not at the moment, but I think we want to get there.
2. Maybe? We'll revisit that with `TextField`.

PR #365, opened 2018-05-07
STCOM-253

rjruizes:
https://issues.folio.org/browse/STCOM-253

PR #373, opened 2018-05-16
STCOM-253 additional translations

rjruizes:
https://issues.folio.org/browse/STCOM-253
Follow up to #365, #368

### DIFF
--- a/lib/Accordion/Accordion.js
+++ b/lib/Accordion/Accordion.js
@@ -81,13 +81,13 @@ class Accordion extends React.Component {
     }
   }
 
-  componentWillUpdate(nextProps, nextState) {
+  componentDidUpdate(prevProps, prevState) {
     // prop/state changes to open status set off the animation callbacks.
     let shouldAdjust = false;
     // if using state...
-    if (typeof nextState.isOpen === 'boolean') {
-      if (nextState.isOpen !== this.state.isOpen) {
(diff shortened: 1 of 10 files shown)


cherewaty:
@Robdel12 should all of these `title` attributes should be `ariaLabel`s instead?

Robdel12:
Yeah titles shouldn’t be used anymore. @JohnC-80 how did you want to attack that? Piece by piece as PRs come through? I think even some of these airia labels can disappear too.

JohnC-80:
There's still more work to do in other components in reducing the usage of `title`, but @rjruizes, if you would, go ahead and prefer `ariaLabel` on these changes for the buttons that are using icons. If a button contains text, both `ariaLabel` and `title` are unnecessary - is that in sync with your thinking, @Robdel12 ?
We'll have to figure out something to give us the tool-tip benefit that the `title` attribute provides at some point, but that's in the future...

cherewaty:
@rjruizes thanks for trying to tackle the `componentWillReceiveProps` deprecations. Since that work involves highly volatile changes, could you split it out into a separate PR? That'd mean we can get the lower-risk internationalization work out quicker.

rjruizes:
Yep, will do. I'm going to do one more component just to get the test to run again. The last one false negative.

PR #374, opened 2018-05-17
STCOM-272 `<EditableList>` custom edit controls

JohnC-80:
https://issues.folio.org/browse/STCOM-272
# Purpose
Currently, all editable fields in `<EditableList>` are `<TextFields>` when in edit mode. This update will enable developers to customize the component presented in an item's edit mode.

# Approach
Apply a new prop `fieldComponents` that works in similar fashion to `formatter` where developers can supply the render function for their column's edit mode.

This shouldn't break anything, as `<TextFields>` are still rendered by default. The only affecting change for existing lists: passing an aria-label prop with mapped column name for a more contextual label.

## Example
Rendering a `<Select>` instead of a `<TextField>` for the `color` column:
```
this.fieldComponents = {
      color: ({fieldProps}) => (
        <Field 
        { ...fieldProps }
        component={Select}
        marginBottom0
        dataOptions={[
          {label: 'orange', value: 'orange'},
          {label: 'blue', value: 'blue'},
          {label: 'red', value: 'red'},
        ]}/>
      )
    }

// ... later in the JSX...
    
    <EditableList 
      {/* other EditableList Props */}
      fieldComponents={ this.fieldComponents }
    />
```


doytch:
Nice!

This is probably just a failure of my imagination, but what do we gain by making `fieldComponents`' values be render functions? Would we actually _want_ them to not use `Field`? 

I.e., could we just pass along `component` and a spread `componentProps` to the `<Field>` we're rendering?

JohnC-80:
@doytch I considered the configuration object approach - it's definitely a possibility. I don't have a very strong preference, but I think the JSX approach places it a bit more in-line with the way that 'formatters' already work and it's not appalling to anyone who's already worked with redux-form.

PR #376, opened 2018-05-17
EditableList STCOM-273: Added support for default field values

doytch:
All fields are blank in newly created list items, there's no way to tell the component what you'd like for default values. This is especially important for uneditable fields.

`itemTemplate` wasn't being used anywhere and looks like an artifact from previous refactoring of `EditableList` according to the previous authors of this. So I made it actually do something since I liked the prop name. 😆


JohnC-80:
At one point I believe itemTemplate was going to be used to determine the type of editable control you got - but I'm glad it stayed vacant for more flexible practices.

PR #380, opened 2018-05-21
Added mock context decorator to storybook

rasmuswoelk:
I noticed that some of the components failed to render in storybook after the addition of different context dependable features - such as the intl-context.

image

To fix this I have added a mock context decorator which mimics the context from stripes-core to prevent certain component examples from failing.

This is probably a temporary fix but it gets the job done.


rasmuswoelk:
Hmm.. I guess it makes sense to have these functions come from stripes-core since they will be used anywhere. 

However, this might be another reason for moving to a more customized setup instead of using storybook?

cherewaty:
Yesterday I was reading about context on the `react-intl` docs and they mentioned this: https://github.com/truffls/storybook-addon-intl

Would that storybook add-on simplify this?

rasmuswoelk:
@cherewaty Brilliant. I have updated the config to use storybook-addon-intl instead. It seems to be working nicely.

I removed my mock context for now but leaving the file there - just in case we need it to mock some other context down the line.

PR #381, opened 2018-05-21
STCLI-53 Add code coverage reporting

cherewaty:
## Purpose
Code coverage reports are a handy to way to identify holes in test coverage. I found myself really wanting that tooling as I was writing tests for `TextField` last week.

## Approach
[Istanbul](https://istanbul.js.org/) is a JS code coverage tool. With the use of [babel-plugin-istanbul](https://github.com/istanbuljs/babel-plugin-istanbul) and [karma-coverage](https://github.com/karma-runner/karma-coverage/) (already a dependency of Stripes CLI), we can generate coverage reports for `stripes-components`.

I added a ton of dummy test files so the corresponding components would appropriately register at 0% coverage, instead of just being ignored.

To run tests that will also generate a coverage report, `yarn test --coverage`.

## Next steps
- Record the coverage statistics with CI to see trends over time. I believe the back-end FOLIO modules might already be doing this? [Coveralls](https://coveralls.io/) and [CodeCov](https://codecov.io/) are nice tools for that.
- Add thresholds where the build will fail if coverage drops under a certain percentage. We're a ways off from wanting to do this - there are a few too many moving parts at the moment. With `ui-eholdings`, we've had the threshold at 90% lines covered for a while, and it's been surprisingly unobtrusive. The couple of times it's piped up about dipping below the threshold were very useful at identifying under-tested features.
- Abstract most of the coverage instrumentation setup into Stripes CLI: https://issues.folio.org/browse/STCLI-53

## Change management
This should not cause any changes for `stripes-components` consumers.

## Screenshots
screen shot 2018-05-21 at 7 20 18 am

### Current totals
screen shot 2018-05-21 at 7 20 34 am

### DIFF
--- a/lib/AppIcon/tests/AppIcon-test.js
+++ b/lib/AppIcon/tests/AppIcon-test.js
@@ -1,0 +1,1 @@
+import AppIcon from '../AppIcon'; // eslint-disable-line no-unused-vars
--- a/lib/Avatar/tests/Avatar-test.js
+++ b/lib/Avatar/tests/Avatar-test.js
@@ -1,0 +1,1 @@
+import Avatar from '../Avatar'; // eslint-disable-line no-unused-vars
--- a/lib/Badge/tests/Badge-test.js
+++ b/lib/Badge/tests/Badge-test.js
@@ -1,0 +1,1 @@
+import Badge from '../Badge'; // eslint-disable-line no-unused-vars
--- a/lib/Callout/tests/Callout-test.js
+++ b/lib/Callout/tests/Callout-test.js
@@ -1,0 +1,1 @@
+import Callout from '../Callout'; // eslint-disable-line no-unused-vars
--- a/lib/Datepicker/tests/Datepicker-test.js
+++ b/lib/Datepicker/tests/Datepicker-test.js
@@ -1,0 +1,1 @@
+import Datepicker from '../Datepicker'; // eslint-disable-line no-unused-vars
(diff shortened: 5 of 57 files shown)


JohnC-80:
In case anyone else was wondering why they called it "Istanbul": https://github.com/gotwarlost/istanbul#why-the-funky-name

Was disappointedly not in the recent version's FAQ.

cherewaty:
😂 so `Istanbul` wasn't forked from a project called `constantinople`?

JohnC-80:
They really baited me hard on the TMBG reference.

funkymalc:
@cherewaty @JohnC-80 @ttamj - When run through CI, Jenkins injects a karma.conf.js file at the top level of the project (for all JS projects).   The karma config contains a custom browser launcher for Chrome in Docker as well as a config for the karma junit publisher used to publish karma test reports.     Can the Jenkins karma config and the native project's karma config co-exist somehow when karma is invoked via stripes-cli?   Should/can I incorporate the coverage config for the coverage reporter into the Jenkins karma-config so that it can be easily shared across all relevant projects?   Or should the Jenkins karma config be integrated into stripes-cli somehow?  What so you guys think is the best approach here?

cherewaty:
@funkymalc I think the right answer is that the Jenkins karma config should be integrated into `stripes-cli`. When I run `yarn install` and `yarn test`, I want to be running the same setup whether I'm on my local machine or on Jenkins. Extra configs and reporters that only Jenkins knows about muddy those waters.

@JohnC-80 we don't have to wait on getting the CI setup right to proceed with this PR though :)

cherewaty:
Follow-up: https://issues.folio.org/browse/STCLI-56

PR #384, opened 2018-05-22
Add more TextField test coverage

cherewaty:
## Purpose
Get test coverage for all the parts of `TextField` needed for refactoring to use the `reduxFormField()` higher-order component.

## Approach
I left several lines uncovered re: calculating padding and RTL rendering. We can revisit those later.

A few things I ran into:
- Unlike other icons, the loading spinner didn't accept styles or class name passed to it. Now it does.
- Testing for the `redux-form` async validation behavior was a little tricky, but I got it. 👌 
- No repo in the `folio-org` is using `TextField`'s `focusInput()`. Should we deprecate it?


lib/TextField/tests/TextField-test.js
```diff
@@ -216,6 +303,58 @@ describe('TextField', () => {
...
+        it('renders an error message', () => {
+          expect(textfield.inputError).to.be.true;
+        });
+      });
+    });
+
+    describe('inputting an valid value with asyncValidating', () => {
+      let resolveValidation;
+
+      const AsyncForm = reduxForm({
```
wwilsman: Would this be useful in other areas as well, such as `TestForm`?

This would be nice: `<TestForm async>`. But I'm not sure if it's possible very easily.

Maybe it could be if `TestForm` dynamically created the underlying redux-form instead of just exporting it. This way maybe other variations of redux-form could be tested by providing `TestForm` with different props.

lib/TextField/tests/TextField-test.js
```diff
@@ -216,6 +303,58 @@ describe('TextField', () => {
...
+        await textfield.fillInput('valid');
+      });
+
+      it('displays a loading icon', () => {
+        expect(textfield.hasLoadingIcon).to.be.true;
+      });
+
+      describe('finishing the async validation', () => {
+        beforeEach(async () => {
+          resolveValidation();
```
wwilsman: I was a little quick on the previous comment. I see now we would need to reference this through `TestForm` somehow. So maybe my previous suggestion wouldn't work out after all

lib/TextField/tests/interactor.js
```diff
@@ -32,6 +32,7 @@ function computedStyle(selector) {
 
 export default interactor(class TextFieldInteractor {
   isInput = is('input', inputClass);
+  isFocused = is('input', ':focus');
```
wwilsman: Minor nit about the rest of this interactor: why is some of this using `inputClass` and others just `'input'`? Should we stick to one over the other for clarity?

cherewaty: I prefer just `input`, but could probably be persuaded otherwise. I agree that we might as well make that consistent.

JohnC-80:
Was just about to pull down and approve. I trust ya. Sorry for the wait.

PR #386, opened 2018-05-24
Modal: Added pass-through of enforceFocus prop

doytch:
This is needed for when you have something that should be able to take focus _inside_ a Modal. 

E.g., without this, having a `<Datepicker>` inside a Modal doesn't work at all - the clicks on the Calendar's day buttons never hit their click handler.

### DIFF
--- a/lib/Modal/Modal.js
+++ b/lib/Modal/Modal.js
@@ -38,6 +38,13 @@ const propTypes = {
   dismissible: PropTypes.bool,
   children: PropTypes.node.isRequired,
   showHeader: PropTypes.bool,
+  /**
+   * When `true` The modal will prevent focus from leaving the Modal while open.
+   *
+   * Generally this should never be set to `false` as it makes the Modal less
+   * accessible to assistive technologies, like screen readers.
+   */
+  enforceFocus: PropTypes.bool,
 };
 
 const defaultProps = {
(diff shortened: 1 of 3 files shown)


JohnC-80:
This solves the direct issue, but this is one case where adequate focus containment or possibly different approach to overlay components will be necessary. Writing up a JIRA...

lib/Modal/Modal.js
```diff
@@ -38,6 +38,13 @@ const propTypes = {
   dismissible: PropTypes.bool,
   children: PropTypes.node.isRequired,
   showHeader: PropTypes.bool,
+  /**
+   * When `true` The modal will prevent focus from leaving the Modal while open.
+   *
+   * Generally this should never be set to `false` as it makes the Modal less
+   * accessible to assistive technologies, like screen readers.
+   */
+  enforceFocus: PropTypes.bool,
```
Robdel12: This is awesome! I'm happy this exists because rolling your own focus trap is _rough_. 

While we're here adding focus props, can we also include `restoreFocus`? https://github.com/react-bootstrap/react-overlays/blob/master/src/Modal.js#L172

It returns the focus back to the triggering element.

doytch: You got it!

PR #387, opened 2018-05-25
STCOM-213 Accordion keyboard navigation

JohnC-80:
# Purpose
Keyboard navigation of accordions benefits both user experience (fast page navigation) and accessibilty (for users that cannot use a mouse.)
## Approach
Groups of `<Accordions>` can now easily gain this feature by being wrapped in an `<AccordionSet>`. Previously, `<AccordionSet>` was limited since `<Accordion>`s had to be direct descendants. Through making use of React context, this limitation is no more.
Accordion header controls have been modified to use `role="tab"` This allows for hotkeys to be used for navigation versus their native screen-reader equivalents for reading elements. 
## Adoption in Modules
This change shouldn't affect existing implementations. They didn't even break existing `<Accordion>` tests. A new HoC `withAccordionSet()` allows existing `<Accordion>` props to pass right on through. If no `<AccordionSet>` is present, the `<Accordion>` instance is passed through. All that's required is for the modules to wrap their current `<Accordion>`s in an `<AccordionSet>`, and the keyboard controls will work. Existing `<AccordionSets>` will gain keyboard navigation.
## Keyboard commands
These are based on the set outlined in [w3.org's ARIA-best-practices](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion). They're active when an accordion header is focused.

| key | function |
| -- | -- |
`up` | Navigate to the previous accordion header. If the first accordion header is focused, the focus will jump to the last accordion. |
`down` | Navigate to the next accordion header. If the last accordion header is focused, the focus will jump to the first accordion. |
`home` | Navigate to the first accordion in the set. |
`end` | Navigate to the last accordion in the set. |

## Other changes
* removed dependency on `context.intl` in headers. This contained "expanded" or "collapsed" text before. Instead, this status will be relayed via the `aria-expanded` property and the more-descriptive label text will be announced.
 * added the `hidden` html5 attribute to closed accordions. This allows the tab-able elements within those accordions to be skipped.

## Todo in this PR
- [x] Tests! Just wanted to get some other eyes on this code.


rasmuswoelk:
Very nice. It would be awesome with an example in the storybook 👍

JohnC-80:
I'm pleased to find that it actually does work in the `Basic Usage` page of storybook, but I went ahead and added a specific story for `Keyboard Navigation` as well as a table for the key bindings in the readme.md.

PR #392, opened 2018-05-29
STCOM-277 Deprecate structures directory

cherewaty:
## Purpose
See [STCOM-277](https://issues.folio.org/browse/STCOM-277)

## Approach
Move the components previously found in the `structures` directory up a level, but continue to export them from their previous location (now with a deprecation warning).

## Screenshot
screen shot 2018-05-29 at 9 13 27 am


MikeTaylor:
> Files changed: 72

Eeek!

PR #395, opened 2018-05-29
Added new DateTimePicker component

doytch:
Bit of a helper here that wraps the two components together.

`DateTimePickerForm` exists to manage the connected components via redux-form. It creates `Datepicker` and `Timepicker` components that ignore timezones.

Why not just roll this into one component? Because then I have to not use redux-form because passing an `onChange` to that component (from some `ui-*` code) would result in that callback being called _by redux-form_ with the separate date/time value object instead of a combined one. That kinda breaks the whole point of a DTP component since now the parent component has to do the combining/timezoning.

Instead, `DateTimePicker` creates a `DateTimePickerForm` and handles the changing values, smooshing the date/time together and applying timezone changes. 

Also: bumped stripes-components up a patch version.

### DIFF
--- a/lib/DateTimePicker/DateTimePickerForm.js
+++ b/lib/DateTimePicker/DateTimePickerForm.js
@@ -1,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Field, reduxForm } from 'redux-form';
+import { Row, Col } from '../LayoutGrid';
+import Datepicker from '../Datepicker/Datepicker';
+import Timepicker from '../Timepicker/Timepicker';
+
+class DateTimePicker extends React.Component {
+  static propTypes = {
+    dateProps: PropTypes.object,
+    timeProps: PropTypes.object,
+  }
+
+  render() {
+    const { dateProps, timeProps } = this.props;
+
+    return (
+      <form>
+        <Row>
+          <Col xs={12} sm={6} md={3}>
+            <Field
+              name="date"
+              component={Datepicker}
+              ignoreLocalOffset
+              {...dateProps}
+            />
+          </Col>
+          <Col xs={12} sm={6} md={3}>
+            <Field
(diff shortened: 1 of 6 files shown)


lib/DateTimePicker/DateTimePickerForm.js
```diff
@@ -0,0 +1,44 @@
...
+  static propTypes = {
+    dateProps: PropTypes.object,
+    timeProps: PropTypes.object,
+  }
+
+  render() {
+    const { dateProps, timeProps } = this.props;
+
+    return (
+      <form>
```
JohnC-80: Since it renders its own `<form>` this couldn't currently be used as a field within an existing form. Nested forms are not allowed in HTML5.
Would it be possible to combine these in a way where they could simply be used as the `component` prop to a `redux-form` `<Field>`?

zburke:
As @JohnC-80 said, this can't be used as a field within an existing form. Can you give a use case or two to show where this would be helpful?

doytch:
Oof yeah, good points. I can remove the `<form>` easily and everything still works, but yeah, it looks like when inside a redux-form's `<Field>`, the values aren't being passed along correctly.

The intended use is [UIU-497](https://issues.folio.org/browse/UIU-497). And I guess CheckIn could also use it, it's already using both Datepicker and Timepicker.

Honestly, I've burned more time than I wanted to spend on this as a little helper, wrestling with the datetime stuff and redux-form. I'm honestly more tempted to just ditch this and do it manually the way the CheckIn app is currently working. I'm gonna put this to the side for now and after I'm done with the main work of UIU-497, I'll dive back in and see what it'd take to let this be renderable by a `<Field>`. Looks like Timepicker/Datepicker might be good jumping off points.

JohnC-80:
Seems like a good idea since the Datepicker/Timepicker still have some sway to them.

PR #396, opened 2018-05-29
Check `.npmrc` file into repo

Robdel12:
## What is this?

Since folio uses its own registry & it's not obvious that is happening we should check in an `.npmrc` file. This is so anyone that clones the repo can just `npm i` (or `yarn`) and be off to the races.

This also allows us to make config changes that everyone can just benefit from rather than asking everyone to update their global config.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+@folio:registry=https://repository.folio.org/repository/npm-folioci/


MikeTaylor:
Robert, I think this one may be controversial: it seems to make sense, but I wouldn't be surprised if there's a specific reason why we _don't_ do this at the moment. Can you please hold off merging till we've heard from some others? I'll tag them to review.

Robdel12:
@MikeTaylor absolutely! :D

skomorokh:
I share Wayne's curiosity and concern. Especially as I get the impression we will start using release NPMs again in the near future. I don't even know what changing my `~/.yarnrc` to point at the release registry will do as my system stands since iirc `folio-testing-platform` has a `.npmrc` pointing at CI (which does makes sense for a testing platform).

Similar, yet subtly different, versions of the same NPMs sounds like it would be horrible to debug if @ttamj hadn't integrated this nice clear duplicate warning. But I'd still like to be able to switch between release and CI NPMs / see which one I'm on at the moment without having to think about more than _exactly one_ place where I toggle that.

Ideally we do CI with tags or similar and then we just have one registry. Then we have a single canonical place for `@folio` and can point at it redundantly. And since we have the `@folio` namespace, that sets us up for making that one registry the global NPM registry it arguably always should have been. 

It was rather a while ago but I recall the reasoning being some combination of a) we don't want the SaaS to need NPM to be up (or a teapot) and b) the nexus thing handling our Java artifacts can be an NPM registry too anyhow. Except every SaaS vendor with that concern will set up their own anyway and can do so as a caching proxy. Plus, new instances of the Stripes bundler service may well be spun up from a docker image of a platform already replete with packages. Either way, it doesn't really ring true as a reason not to use the public registry. If we lack other reasons, maybe we can just move there and be done with it?

In the mean time, the idea of a half dozen overlapping npm configurations affecting where my packages come from isn't super thrilling but I guess it's maybe okay if eg. `~/.yarnrc` turns out to hard override registry settings in the nested ones?

MikeTaylor:
Thanks, @wafschneider and @skomorokh in particular, but everyone who's chipped in. I am leaning towards _not_ making this change, as it embeds a bit more magic in the system -- one more thing that Just Works in a really nice way provided you're working in one specific way, but which can cause mysterious and hard-to-diagnose errors the moment you start doing things differently. Yes, as @JohnC-80 says, the need to configure NPM explicitly as part of the setup process is one more element of onboarding; but I think the very explicitness of how we do it now forces it to be an issue in people's minds, which is good.

ttamj:
Related to @skomorokh's comment about being able to switch registries, we unfortunately cannot do so reliably.  With `npm-folioci` having artificially inflated patch numbers, and some `package.json`'s pointing to specific CI builds already, interchangeability with `npm-folio` (or some other vendor's registry) is not going to yield the expected results.  That said, I don't think we should embed a hard dependency on `npm-folioci` in hopes we can untangle this.

PR #398, opened 2018-05-29
Remove create-react-class from dependencies

cherewaty:
## Purpose
`create-react-class` was added as a dependency in https://github.com/folio-org/stripes-components/commit/93c421f2e357e676d2ae7fe8525ae0bf3eb44055

## Approach
`@folio/stripes-react-hotkeys` declares a dependency on `create-react-class`, and `@folio/stripes-components` has no other dependency on `create-react-class`. Therefore it doesn't need to be declared as a dependency of `@folio/stripes-components`.


cherewaty:
Also removed `mousetrap` - same story.

PR #399, opened 2018-05-29
Clean up dependencies

cherewaty:
## Purpose
I was noticing extraneous dependencies (https://github.com/folio-org/stripes-components/pull/398) so I ran [`depcheck`](https://www.npmjs.com/package/depcheck).

## Specifics
- I changed a few paths to be relative so `depcheck` and `eslint` weren't picking up `@folio/stripes-components` as a dependency of itself.
- Most of the changes were removing unused testing infrastructure.
- `stripes-components` now explicitly states its dependency on `stripes-connect` and `redux`. The components with those dependencies should probably be either refactored or moved to `stripes-smart-components`, but that's future work.
- I did not address `Pluggable`'s dependency on `stripes-loader`. Issue created: https://issues.folio.org/browse/STCOM-285


MikeTaylor:
Thanks for sorting these out. Some of the things you changed here were pretty gruesome -- stripes-components depending on itself :-)

Why did babel-polyfill and babel-runtime get _added_ to the dev dependencies?

If query-string is needed as a dependency, how was ESLint not picking that up?

It's annoying that `makeQueryFunction`'s use of `compilePathTemplate` means stripes-connect can't be just a peer-dependency. I wonder whether in the future we can move that utility function out to somewhere lower level. But that shouldn't hold up the present PR.

I don't know how any of the components here use redux and redux-form, but is it possible that they could be given as peer-dependencies rather than regular dependencies?

cherewaty:
- `babel-polyfill` is needed by the `stripes-components` tests
- `@storybook/react` has a peer dependency on `babel-runtime`
- `query-string` was down the dependency chain of both `storybook` and `stripes-cli`; that may have been why ESLint didn't pick it up? (not 100% sure on that one)
- `RepeatableField`, `EditableList`, and `AddressFieldGroup` all have dependencies (not peer) on `redux-form`.

PR #400, opened 2018-05-29
Use own translation strings instead of stripes-core

cherewaty:
`stripes-components` should rely solely on its own translation strings, instead of leaning on `stripes-core`'s.

### DIFF
--- a/lib/EntrySelector/EntrySelector.js
+++ b/lib/EntrySelector/EntrySelector.js
@@ -82,7 +82,7 @@ class EntrySelector extends React.Component {
         onClick={() => this.props.onEdit(item)}
         id="clickable-edit-item"
         ariaLabel="edit"
-        title={this.props.stripes.intl.formatMessage({ id: 'stripes-core.button.edit' })}
+        title={this.props.stripes.intl.formatMessage({ id: 'stripes-components.button.edit' })}
         size="medium"
       />
     );
@@ -117,7 +117,7 @@ class EntrySelector extends React.Component {
     const LastMenu = addMenu || (
       <PaneMenu>
         <Button title={addButtonTitle} onClick={this.props.onAdd} buttonStyle="primary paneHeaderNewButton">
-          {this.props.stripes.intl.formatMessage({ id: 'stripes-core.button.new' })}
+          {this.props.stripes.intl.formatMessage({ id: 'stripes-components.button.new' })}
         </Button>
       </PaneMenu>
     );
--- a/translations/en.json
+++ b/translations/en.json
@@ -47,5 +47,7 @@
   "metaSection.source": "Source: {source}",
   "metaSection.recordCreated": "Record created: {date} {time}",
   "metaSection.recordLastUpdated": "Record last updated: {date} {time}",
-  "tableColumnCount": "Table with {count} columns"
+  "tableColumnCount": "Table with {count} columns",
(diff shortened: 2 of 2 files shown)


MikeTaylor:
I'm not sure about this one. In general I agree it's good that we reduce stripes-components' dependency on stripes-core whenever we easily can. (When the day finally comes that we split stripes-core into stripes-core-core and stripes-core-app, and the dependency graph becomes a simple linear one of core-app -> components -> core-core, this will all make more sense.)

But IIRC the goal of putting certain translations into stripes-core was so that they could be re-used in a uniform way across the system -- which would include use by re-usable components as well as by individual apps. I don't know whether the present translations ({{button.edit}} and {{button.new}}) are in that category, and should therefore remain in stripes-core.

I've added @ibendermonguz as a reviewer, as he is more deeply involved in translation that we are. He should have some relevant insight.

ibendermonguz:
Currently it is used in ui modules as well. e.g.: https://github.com/folio-org/ui-organization/blob/7d2cf336857532a51426a14e877d3901a9df124d/settings/LocationLocations/LocationForm.js#L172

So it seems stripes-core is the right place.

cherewaty:
If the `EntrySelector` is relying on a `stripes-core` translation string, it can't be rendered and tested on its own without failing to find the string, and there's a circular dependency created between `stripes-core` and `stripes-components`.

MikeTaylor:
Yes indeed. This kind of thing is one of the reasons that I have been saying for so very long that we really need to separate out the two quite different things that are currently conflated in stripes-core. There is the actual code -- that without which no Stripes code can run, such as the provision of the Stripes object, the set of loaded modules, and the core translations. And there is the specific application that we are building on top of that. stripes-components properly depends on the former, but not on the latter.

MikeTaylor:
@ibendermonguz That sounds reasonable to me.

But @JohnC-80 if for now you want to merge this anyway, that's fine -- we can always move these things again if we decide to.

MikeTaylor:
Oh, except no: as [Istvan points out](https://github.com/folio-org/stripes-components/pull/400#issuecomment-393217566), we have [other code](https://github.com/folio-org/ui-organization/blob/7d2cf336857532a51426a14e877d3901a9df124d/settings/LocationLocations/LocationForm.js#L172) that also depends on these translations from stripes-core.

So we can't simply remove those translations from stripes-core. Which means that if we add them here in stripes-components, we're making a non-DRY duplicate, with all the potential for drifting apart that that implies. Do we want to do that?

cherewaty:
In my view, preventing a circular dependency between `stripes-core` and `stripes-components` takes priority over avoiding duplication. Duplication is a natural side-effect of a microservice architecture, and we live with it to get the benefits of independent, isolated pieces.

MikeTaylor:
We already have a circular dependency, for this and many other reasons. Fixing that is a longer-term problem. At present we don't _additionally_ have the problem of duplication. I prefer to retain one problem than to live with two.

JohnC-80:
I don't see an issue with a temporary duplication - once these things do get a centralized repo outside of stripes-core that anyone can safely depend on, I imagine they'll get moved there and duplicates cleaned up. I can't imagine `stripes-core` is the permanent home for these, either... the idea of needing to update any core `stripes-module` to get updated translations seems odd to me.

PR #408, opened 2018-06-04
Run ESLint for entire repository

cherewaty:
## Purpose
- In https://github.com/folio-org/stripes-components/pull/399#issuecomment-393144465 we noticed that the `query-string` dependency had not been picked up by ESLint - that's because the `util` directory was not being linted.
- The Stripes ESLint config has matured a bit, so some of the line-disabling no longer had to take place.
- Updates to `eslint-plugin-react` started throwing very chatty warnings about lifecycle hook deprecations. Those are very sensitive refactors that can take place more confidently and responsibly when those components have tests.

## Approach
- ESLint now covers the entire project, not just `lib`.
- Unnecessary line disabling removed.
- Every `eslint-disable-line` now explicitly states which rule to ignore.
- Disable the lifecycle hook warnings for now. There's already a JIRA issue for that: https://issues.folio.org/browse/STCOM-275

### DIFF
--- a/lib/MetaSection/MetaSection.js
+++ b/lib/MetaSection/MetaSection.js
@@ -14,8 +14,7 @@ const propTypes = {
   contentId: PropTypes.string,
 };
 
-class MetaSection extends React.Component { // eslint-disable-line
-
+class MetaSection extends React.Component {
   constructor(props, context) {
     super(props);
     this.context = context;
--- a/lib/Accordion/Accordion.js
+++ b/lib/Accordion/Accordion.js
@@ -226,7 +226,6 @@ class Accordion extends React.Component {
     }
   }
 
-  // eslint-disable-next-line class-methods-use-this
   getContentClass(open) {
     return classNames(
       css.content,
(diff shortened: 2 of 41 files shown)


lib/MetaSection/MetaSection.js
```diff
@@ -14,8 +14,7 @@ const propTypes = {
   contentId: PropTypes.string,
 };
 
-class MetaSection extends React.Component { // eslint-disable-line
-
+class MetaSection extends React.Component {
```
wwilsman: In the other places with `class extends React.Component` we have specified to ignore the `react/no-deprecated` rule.

But here we remove the disable comment completely?

cherewaty: `MetaSection` doesn't appear to have any deprecated lifecycle hooks. I can't tell why that line was disabled.

wwilsman: Gotcha. I thought it was specific to `extends React.Component` since that _is_ the line being disabled on the rest.

PR #411, opened 2018-06-04
Alphabetize propTypes

cherewaty:
## Purpose
As I was working in a component with a lot of props the other day, I accidentally added a prop that already existed. ESLint immediately caught the duplicate prop for me, but I still wanted more organization to the props.

## Approach
`eslint-plugin-react` has exactly what I wanted: [`react/sort-prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-prop-types.md). It enforces alphabetization of `propTypes`.

I probably wouldn't advocate for turning that rule on in `stripes-core` or individual FOLIO modules, but it'll be very useful for a shared library like `stripes-components` that aims to be as self-documenting as possible.

## Future work
It'd be awesome if the prop tables in READMEs were also alphabetized. I didn't attempt to tackle that in this PR, but as I make additions/subtractions to READMEs in the future I'll do it.


MikeTaylor:
Surely there are going to be at least some cases where the developers of components have deliberately authored the props in what they consider to be a semantically meaningful order? (I know that's the case with my own `<SearchAndSort>`, but that's in stripes-smart-components, so that's not directly relevant here.)

cherewaty:
When a component has so many props that any more organization than alphabetization is needed, that's a strong indicator that:
- the component is trying to fit too many use cases, and should be multiplied horizontally
- or the component should be multiplied vertically into several composable components

No existing `stripes-components` appear to have any semantic meaning in their prop organization.

PR #412, opened 2018-06-04
Add timezone prop to Datepicker

cherewaty:
## Purpose
Mimics the approach for `Timepicker` in https://github.com/folio-org/stripes-components/pull/401. Adding a `timezone` prop gets us to testability for the `Datepicker` without dealing with `stripes` context.

## Approach
Avoids using default props for `locale` and `timezone`, so the logic is more clear. When we get to the point of implementing a React 17-friendly strategy for context, we should revisit that implementation.

The most exciting part of this PR: TESTS! Datepicker and Timepicker now have minimal tests.

## Future work
`ignoreLocalOffset` is no longer really needed, but I left it in for now.


cherewaty:
I agree that `state.parseTimezone` seems a little off. I'll be revisiting `Datepicker` pretty soon to use the `reduxFormField` higher-order component, so that'll be a good time to tackle that refactor.

PR #414, opened 2018-06-07
STCOM-270: Relocate language files

ibendermonguz:

PR #419, opened 2018-06-08
Fix translations path for storybook

cherewaty:
## Purpose
Storybook wasn't runnable, because the translations files moved. Follow-up to https://github.com/folio-org/stripes-components/pull/414

## Future
We should add a test to CI just to make sure storybook builds.


rasmuswoelk:
Merging this in.. can't use storybook without it ;)

PR #421, opened 2018-06-12
STCOM-289: Added autocomplete prop to TextField and turned it off in Datepicker

doytch:
This turns off the autocompletion for the Datepicker (STCOM-289) but also gives us more granularity over autofills in the future [by specifying a particular string such as `street-address`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#attr-autocomplete).

PS: I think it's pretty dumb that React makes you write `autoComplete` instead of `autocomplete` to avoid throwing an error but 🤷‍♂️.

### DIFF
--- a/lib/TextField/TextField.js
+++ b/lib/TextField/TextField.js
@@ -12,6 +12,7 @@ import omitProps from '../../util/omitProps';
 export default class TextField extends Component {
   static propTypes = {
     ariaLabel: PropTypes.string,
+    autoComplete: PropTypes.string,
     autoFocus: PropTypes.bool,
     clearFieldId: PropTypes.string,
     /**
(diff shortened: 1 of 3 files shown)


lib/TextField/TextField.js
```diff
@@ -273,6 +274,7 @@ export default class TextField extends Component {
         {...inputCustom}
         aria-required={required}
         aria-label={this.props.ariaLabel}
+        autoComplete={this.props.autoComplete}
```
cherewaty: Should we add `autoComplete` to `defaultProps`?

doytch: I thought about it but figured I'd leave the prop `undefined` for default behaviour. That also lets a developer use the inherited version of `autocomplete` that's defined on a `<form>` without having to override it in every `<TextField>`

doytch: Here's the relevant section from the WHATWG spec:

"If the autocomplete attribute is omitted, the default value corresponding to the state of the element's form owner's autocomplete attribute is used instead (either "on" or "off"). If there is no form owner, then the value "on" is used."

PR #426, opened 2018-06-13
Use injectIntl() instead of context.intl

cherewaty:
## Purpose
Now that we know it's on the fast train to deprecation, we want to avoid using the old React context API.

Related to https://github.com/folio-org/stripes-core/pull/321

## Approach
Use `react-intl`'s `FormattedDate`, `FormattedNumber`, and `injectIntl` instead of `context.intl` or `stripes.context.intl`. When `react-intl` moves to the new context API under the hood, those abstractions should still work.
- Although destructuring the `intl` and then shortening `formatMessage()` to `formatMsg()` saved characters, it hurt searchability in the repo. Now it's consistently `intl.formatMessage()`.
- The storybook `withMockContext.js` wasn't needed anymore.


rjruizes:
The disadvantages of InjectIntl:
1. the ref changes; you have to refer to the wrappedInstance (see https://github.com/folio-org/ui-checkout/pull/125/files)
2. the DOM is cluttered with wrappers

Since the legacy context API deprecated, I thought we would transition the the new context API. Keep a single <IntlProvider> at the root level rather than wrapping every stripes component with <InjectIntl>.

cherewaty:
`injectIntl()` shouldn't add any DOM on its own, even though it clutters the React component tree. `FormattedMessage`, `FormattedDate`, etc. will add `<span>`s, but that shouldn't be a big deal.

The ref problem with `injectIntl()`: definitely a thing. I wonder if `react-intl` could use a pull request that uses `forwardRef` in `injectIntl()` to fix that problem. Would definitely be a breaking change to the library. I'll look into it.

Is the `TextField` the only component we're worried about modules accessing refs to?

cherewaty:
New commit added that forks `injectIntl()` with `forwardRef()`. It should maintain existing ref behavior (see the `TextField-test` for an example).

I'll make a PR to `react-intl` to see if we can get that nicer ref forwarding behavior upstream, but this fork should work nicely for us in the meantime.

JohnC-80:
I doubt TextField is the only component that people want to get refs to - prettymuch any form control will be needed if the focus is being managed somehow by the application. #422 from yesterday actually did break some functionality, but I feel like it's due to the instance being accessed in a non-ideal way (seen occurrence of actually going in to grab the `input` instance property) - @cherewaty you mentioned deprecating getInput() due to lack of use - I couldn't help but wonder how else people were getting the input element if they needed it.

cherewaty:
We'll hold off on this PR until we have the `ref` issues exposed by https://github.com/folio-org/stripes-components/pull/422 sorted out.

PR #427, opened 2018-06-14
Icon: Added "status" prop used for fill color

doytch:
It's currently not possible to `fill` an `<Icon>`'s SVG without hardcoding a color in your JS and passing that as a prop. This is because we currently don't have a way of getting CSS variables into JS. 

Until we do, I added a `status` prop on Icon that satisfies the main goal, which is colouring an Icon based on our warn/error/success colours.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@
 * Storybook resource added. Run `yarn storybook` to see component demos and documentation.
 * Pass a ref through to TextArea for access by a parent. Refs STSMACOM-4.
 * Add ability to pass custom filter to `<Selection>` component. Fixes STCOM-251.
+* Add ability to colour `<Icon>` using `status` prop.
 
 ## [1.9.0](https://github.com/folio-org/stripes-components/tree/v1.9.0) (2017-10-13)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v1.8.0...v1.9.0)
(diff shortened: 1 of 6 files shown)


JohnC-80:
I like the idea - I want @rasmuswoelk to have a look prior to approval.

rasmuswoelk:
Looks good 👍 

We preserved the global className "stripes__icon" for updating the icon color from external CSS but I think it's good to have the status-prop as well to avoid duplicate CSS in some circumstances.

Can you add examples of status colors in the storybook example file as well?

doytch:
Good point, added a separate story for it since it doesn't play nice alongside `color` as a knob.

rasmuswoelk:
@doytch Ah ok. Actually, I think we should remove the color-knob. We would want developers to change the colors of the icons either with CSS or by passing the status prop instead of using fill.

Do you want me to remove it or can you remove the knob as you update with the new status-prop?

doytch:
I can remove it, but your rationale leads me to think that we should then go ahead and remove the `color` prop entirely from `<Icon>`. Do you agree? 

In any case, I'll respin the storybook changes I just pushed to instead update the knob with setting the `status` and merge this. If we _did_ deprecate usage of `color`, that'd obviously _technically_ be a breaking change even though no one's using it from my quick search of the codebase.

rasmuswoelk:
@doytch Correct. It's probably best to remove it entirely. And yeah it would be a breaking change but as you mention; no apps or components use the color-prop anyway so I think we can go ahead and remove it without any trouble.

PR #438, opened 2018-06-20
STCOM-290 Fix TextField clear button

cherewaty:
## Purpose
The clear button on `TextField` is broken. Related to https://issues.folio.org/browse/STCOM-290.

## Approach
Adopts the pattern established in https://github.com/folio-org/stripes-components/pull/372 with the `reduxFormField` higher-order component.

`reduxFormField` needed to start using `forwardRef` so the `TextField`'s refs would be available to parent components.

The clear button was previously in the tab order when a field was marked invalid, but not in the tab order in any other case. Now it's consistenly not tabbable.

I split out the test assertions specific to using `TextField` with `redux-form`.

Also added comments for every `TextField` prop.

## Warnings
`forwardRef` works with `redux-form` `Field`s, but currently throws some prop type warnings:
```
ERROR: 'Warning: Failed prop type: Invalid prop `component` supplied to `Field`.
ERROR: 'Warning: Failed prop type: Invalid prop `component` supplied to `ConnectedField`.
```

I thought I addressed that in https://github.com/erikras/redux-form/pull/4041, but it looks like `forwardRef` doesn't even return a `node`, just an object. There's a lot of recent chatter in the React repos about that - I'll see if a recommended pattern emerges.

## Versioning
Adding new props should trigger a minor version bump.

## Testing
- Manually tested with `storybook`.
- Manually tested with VoiceOver on.
- Ran `ui-eholdings` automated tests with these changes to `stripes-components`.

## Next Steps
- https://issues.folio.org/browse/STCOM-290 raises the question of whether our custom clear button should exist at all.
- https://issues.folio.org/browse/STCOM-287 might be related to the padding calculations going on.
- Add a README.
- Add even more tests for `TextField`.

## Learning
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

### Before
before

### After
after


lib/TextField/TextField.js
```diff
@@ -252,126 +349,102 @@ export default class TextField extends Component {
...
       <input
-        ref={this.input}
-        className={this.getInputStyle()}
-        onFocus={this.onFocus}
-        onBlur={this.onBlur}
         {...inputProps}
         {...inputCustom}
-        aria-required={required}
         aria-label={this.props.ariaLabel}
+        aria-required={required}
```
wwilsman: Shouldn't this input also get a `required` attribute?

lib/TextField/TextField.js
```diff
@@ -252,126 +349,102 @@ export default class TextField extends Component {
...
+          ariaLabel={intl.formatMessage({ id: 'stripes-components.fieldIsValid' })}
+          icon="validation-check"
+          iconClassName={css.successIcon}
+          id={`icon-${this.testId}-validation-success`}
+        />
+      );
+    }
+
+    if (validationEnabled && this.props.error) {
+      validation = validationEnabled && (
```
wwilsman: This second `validationEnabled` seems redundant under `if (validationEnabled)`

cherewaty:
After my latest rebase, I re-tested against `ui-eholdings`. Looks like this breaks some of `Datepicker`'s `onBlur` behavior. Marked as WIP until I get that fixed.

cherewaty:
The `Datepicker` blur bugs were from other work (and have since been resolved). I've retested this against `ui-eholdings`, and it should be good to go. cc @JohnC-80

PR #439, opened 2018-06-21
Auto close Pane Action Menu on NavItem click

rasmuswoelk:


cherewaty:
cc @elrickvm @sadatay @wwilsman

JohnC-80:
This is okay, but still lacks on flexibility for handling the use case where a developer wants to control whether or not a menu item closes the Dropdown.
I think if we had 2 props: `actionMenuOpen`(bool) and a `toggleActionMenu` callback, apps could adequately control the menu from their own state if they need to. 
Optional state/props could be handled with `getDerivedStateFromProps` - something like
```
static getDerivedStateFromProps(props, state) {
    const nextState = null;
    if (typeof props.actionMenuOpen !== 'undefined') {
      nextState.actionMenuOpen = props.actionMenuOpen;
    }

    return nextState;
  }
```
Eventually, there's going to be more control needed for this if modules want checkboxes and whathaveyou in the pane dropdowns.

sadatay:
while this fixes our immediate issue i agree fully with @JohnC-80 that a more robust solution (as described) would be beneficial in the future

elrickvm:
@JohnC-80 and @sadatay I concur.

rasmuswoelk:
I agree that a more robust solution is needed here. Once we have some more details about the more complex Pane Action Menu we can begin to plan out how to implement this feature properly. We should properly move this functionality out of the PaneHeader itself and into its own component(s).

Thanks for the approval, @JohnC-80. I'll merge it in for now.

PR #441, opened 2018-06-21
Clean up injectIntl

cherewaty:
## Purpose
`injectIntl` needed a little more time in the oven. Follow-up to https://github.com/folio-org/stripes-components/pull/426

Being thrown:
```
ERROR: 'Warning: Failed prop type: Invalid prop `forwardedRef` of type `function` supplied to `InjectIntl`, expected a single ReactElement.
```
and
```
ERROR: 'Warning: React does not recognize the `forwardedRef` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `forwardedref` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

## Approach
- Be more permissive of `forwardedRef` prop type.
- Don't pass along `forwardedRef` to children.
- Give a `displayName` for `injectIntl`.

### DIFF
--- a/lib/InjectIntl/InjectIntl.js
+++ b/lib/InjectIntl/InjectIntl.js
@@ -4,12 +4,19 @@ import { intlShape } from 'react-intl';
 
 export default function injectIntl(WrappedComponent, options = {}) {
   const { withRef = false } = options;
+  const componentName = WrappedComponent.displayName || WrappedComponent.name || 'Component';
 
   class InjectIntl extends Component {
     static WrappedComponent = WrappedComponent;
+    static displayName = `InjectIntl(${componentName})`;
 
     static propTypes = {
-      forwardedRef: PropTypes.element
+      forwardedRef: PropTypes.oneOfType([
+        PropTypes.node,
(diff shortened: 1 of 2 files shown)


cherewaty:
@cowboyd I took a bit of a lazy approach - the HOC gets run through several scenarios by virtue of being called by other components' tests. The errors thrown didn't cause any existing tests to fail, since all the functionality still worked.

MikeTaylor:
It's late in the day to be asking this, but ...

Do we know why this is called `injectIntl` rather than `withIntl` (as in for example `withRouter`)? Is this just something that the wider React community hasn't yet settled on a convention for?

cherewaty:
I'm guessing `react-intl` just settled on that API before `withIntl()` would've been the community standard. I'll go back this week and make it more clear that the version of `injectIntl()` introduced in this PR is only intended to be a temporary fork. I'll be making a PR to `react-intl` to see if we can get the much cleaner `forwardRef()` upstream (instead of having to use `getWrappedInstance()`).

PR #446, opened 2018-06-22
Modify makeQueryFunction to support param namespacing. Part of STCOM-300

mkuklis:
The purpose of this PR is to introduce a way for namespacing params used by `makeQueryFunction`.
The implementation follows the document written by @MikeTaylor here:

https://github.com/folio-org/stripes-components/blob/master/util/parameterizing-makeQueryFunction.md

There is another PR against `stripes-smart-components` which modifies `<SearchAndSort>` component in order to set/namespace the URL query parameters: 

https://github.com/folio-org/stripes-smart-components/pull/201/files

An example of the usage of this new functionality can be found in `ui-users`:

https://github.com/folio-org/ui-users/blob/STCOM-300/Users.js#L37-L59

### DIFF
--- a/util/nsQueryFunctions.js
+++ b/util/nsQueryFunctions.js
@@ -7,7 +7,7 @@ export function getNsKey(key, params) {
   return (isString(params)) ? `${params}.${key}` : (params[key] || key);
 }
 
-// Adds namespace / prefix to keys for given query
+// Adds namespace / prefix to keys for given values object
 //
 // example:
 // values = { query: "test", filters: 'active', userId: 1 }, params = 'users'
@@ -18,7 +18,7 @@ export function mapNsKeys(values, params) {
   return mapKeys(values, (value, key) => getNsKey(key, params));
 }
 
-// Removes namespace / prefix from keys for given values
+// Removes namespace / prefix from keys for given values object
 //
 // example:
 // values = { "users.query" : "test", "users.filters": "active" }, params = 'users'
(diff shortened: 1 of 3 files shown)


util/nsQueryFunctions.js
```diff
@@ -0,0 +1,45 @@
...
+  if (!params || !PARAM_FILTER[key]) return key;
+  return (isString(params)) ? `${params}.${key}` : (params[key] || key);
+}
+
+// Adds namespace / prefix to keys for given values object
+//
+// example:
+// values = { query: "test", filters: 'active', userId: 1 }, params = 'users'
+// result:
+// { "users.query" : "test", "users.filters": "active", userId: 1 }
```
zburke: Should the userId key be prefixed too? I was expecting
```
// { "users.query": "test", "users.filters": "active", "users.userId": 1 }
```

mkuklis: Thanks for reviewing this @zburke. I left the `userId` on purpose there to show that we only deal with the search and sort params and others will be left untouched.

zburke: Ah, ok; so it only names-spaces the keys in `PARAM_FILTER`. That wasn't immediately clear because `PARAM_FILTER` isn't mentioned in `mapNsKeys`, though I see it in `getNsKey`. Maybe add a comment on the constant to explain how it's used in the mapping functions, or a general "this is how things work" header comment?

mkuklis: Good idea @zburke! Also I'm not sure about the naming conventions for those functions. Not sure they are clear enough. What do you think?

zburke: The functions here are clearly and consistently named. I generally shy away from abbreviations because they can make the codebase a little harder to search ("namespace" appears only in the comments here, never in a function-name) but I don't have strong feelings about it.

mkuklis:
@JohnC-80 sounds good.  @zburke I added couple more comments and changed the `PARAM_FILTER` to `PARAM_WHITELIST`.

PR #448, opened 2018-06-25
Add Testing Documentation

wwilsman:
## Purpose

[[UIEH-436] Stripes Component Testing Guide](https://issues.folio.org/browse/UIEH-436)

To help people more easily start writing tests, we should have a guide / tutorial showing how it's done.

## Approach

I opted to write a guide instead of a step-by-step tutorial to also explain _why_ we should write tests in addition to the how-to.

In the actual how-to part, I preface with a brief explanation of convergences and why we separate side-effects. Instead of step-by-step on writing tests, I point to existing tests (the button tests) to study from and copy it's patterns.

I then explain each of the main patterns, from the `mount` helper and separating side-effects, to the actual interactor. Instead of explaining interactors in-depth and the various properties, I give a gist then link to newly released interactor guides (which are much more in-depth). I also dedicated a short section to the `selectorFromClassnameString` helper to explain why and when it's needed.

Finally, I left the part about testing context last, since we should discourage it. I still gave an example of the `mountWithContext` helper in case it's absolutely necessary, but once we fully migrate over to the new context API, we should likewise change / remove this section.

I also cleaned up the button tests since I used it as an example. I noticed we were letting the interactor default to the document body, and then having each property target the button individually. It's better if the whole interactor is scoped to the button, and we only need to create properties that don't already exist (like `text` and `click`). Making the button interactor in this way allows it to be reusable and we can target specific buttons using a selector with the constructor (`new ButtonInteractor('#button-2'`).

### Learning

- There are some good resources on interactors now with the new [BigTest Interactor Guides](https://bigtestjs.io/guides/interactors/introduction/). Hopefully linking to these in this testing document will help inform others in the future.


wwilsman:
[Rendered Document](https://github.com/folio-org/stripes-components/blob/ww/testing-in-stripes-components-doc/docs/Testing.md)

docs/Testing.md
```diff
@@ -0,0 +1,248 @@
...
+Tests in stripes components are written using
+[Mocha](https://mochajs.org/) via
+[`@bigtest/mocha`](https://github.com/bigtestjs/mocha). This package
+makes code within `it` blocks _convergent_, which means that if any
+assertions within fail, they will run again repeatedly until they pass
+or a timeout elapses. This makes testing asynchronous rendering easier
+since we don't have to know exactly _when_ a component, or any updates
+to a component, render into the DOM. As long as it renders within the
+timeout, the test will eventually pass.
+
```
cowboyd: Convergences can be counterintuitive to folks encountering them for the first time, and so I'd spend a sentence or two introducing them as a concept first.

> Tests in Stripes components a written using [Mocha](https://mochajs.org/) and [BigTest](https://bigtestjs.io) (via [`@bigtest/mocha`](https://github.com/bigtestjs/mocha)). At the core of BigTest is the concept of *convergences* which make testing asynchronous applications both robust and fast. They accomplish this by freeing us from the burden of worrying about _when_ a component, or any updates to a component, will render into the DOM. As long as it renders within the timeout, the test will eventually pass. To read more about how this strategy works, and why it is so effective, checkout the [documentation on convergences]

docs/Testing.md
```diff
@@ -0,0 +1,248 @@
...
+Stripes Components uses
+[`@folio/stripes-cli`](https://github.com/folio-org/stripes-cli) to
+launch a browser and run our tests using
+[Karma](http://karma-runner.github.io). The tests themselves live
+alongside each component in their own `tests` subdirectory, and are
+written using [Mocha](https://mochajs.org/),
+[Chai](https://github.com/chaijs/chai), and
+[BigTest](https://bigtestjs.io/).
+
```
JohnC-80: I feel like we can skip some of the "this is why testing is important" part of this - or probably move it to some of the introductory documentation, like the testing section of the top-level Readme.md

docs/Testing.md
JohnC-80: I agree about convergences - and even I'd go as far as to say they may be a little deep for the thick of this document. I think deeper info can either be left up to links out to BigTest's docs or possibly a section on "Why convergence-based tests are so fast" at the end to resolve any deeper curiousity.

docs/Testing.md
```diff
@@ -0,0 +1,248 @@
...
+  describe('clicking the button', () => {
+    beforeEach(async () => {
+      await button.click();
+    });
+
+    it('calls the onClick handler', () => {
+      expect(clicked).to.be.true;
+    });
+  });
+});
```
JohnC-80: Would it be a more useful flow to include comments within this code snip that describe each of the pieces rather than placing it after the fact? 
When I see descriptions after lengthy code portions I find myself scrolling back and forth to find what it was talking about. Many readers of this will be digesting this anatomy of a `*-test.js` file for the first time and it's tough to retain that shape to apply the understanding later.

docs/Testing.md
```diff
@@ -0,0 +1,248 @@
...
+
+An `Interactor` is a composable [page
+object](https://martinfowler.com/bliki/PageObject.html) we use
+throughout our tests. This adds a layer of durability to our suite
+because when things like classnames or the markup inevitably change,
+we should only need to update our interactor as oppose to updating
+each of our tests. Interactors can also be composed by each other, so
+if another component uses a button, that component's interactor can
+use the `ButtonInteractor` too.
+
```
JohnC-80: I like this description of Interactors and the benefits they give. While readers of this doc may have experience with mocha tests, the shape of this file might be totally alien to them - so, again, code comments?
From personal experience, as a new BigTest-er I find myself wondering about the best workflow strategy between setting up the interactor and writing out the assertions. I suppose we'd need interactors to cover any portion of the DOM that a component updates - eventually it'd be cool to grow this out to recipes/snippets for testing often repeated aspects of the DOM.

wwilsman:
@JohnC-80 I just noticed that there is a top-level TESTING.md. Should this content go there, or remain in the `docs` folder (and move the top-level document content to this document)

wwilsman:
Added comments to the big code snippet and changed some wording based on feedback.

Since I noticed the `TESTING.md`, I deleted that and moved some of it's content into the new testing documentation. I also updated the README with some of the introductory content and pointed to this new document.

docs/Testing.md
```diff
@@ -0,0 +1,281 @@
...
+helper will clean up and teardown any previously mounted component,
+then mount our new component into the DOM for us. We use the
+`async`/`await` syntax to wait for React to mount our component so the
+tests don't start running too early. Since this helper cleans up
+previous components on _the next_ call, this allows us to investigate
+and debug our tests after they run. You can then use Mocha's `it.only`
+to isolate a specific test to debug.
+
+Since `@bigtest/mocha` makes our tests convergent, and could possibly
+run them multiple times, _they only contain assertions_. You'll also
```
Robdel12: Maybe be a little more explicit here and say "they **must** only contain assertions"? We've seen this trip up people many times already

wwilsman: I thought it read weird since we're talking about an observation of the tests, but the actual sentence structure doesn't reference said observation (like "**in** our tests"). So I added and kept it since it is still technically correct.

rasmuswoelk:
This is great 👍It's a very helpful document already.

PR #449, opened 2018-06-25
STCOR-208 Use new context API: withStripes()

rjruizes:
https://issues.folio.org/browse/STCOR-208

- This PR converts components from the legacy context API to the new context API
- `withStripes()` is a HoC to wrap components with a `StripesContext.Consumer`
- The Providers were set in https://github.com/folio-org/stripes-core/pull/330


cherewaty:
Please get review(s) before merging. Resulting bug: https://github.com/folio-org/stripes-components/pull/450

PR #450, opened 2018-06-25
Fix withStripes bugs

cherewaty:
## Purpose
In a storybook context, `Datepicker` and `Timepicker` (https://github.com/folio-org/stripes-components/pull/443) wouldn't receive the `intl` context. The `intl` prop would just get eaten before it made its way down.

screen shot 2018-06-25 at 4 36 54 pm

## Approach
Flip the order of the higher-order components.

### DIFF
--- a/lib/Datepicker/Datepicker.js
+++ b/lib/Datepicker/Datepicker.js
@@ -612,4 +612,4 @@ class Datepicker extends React.Component {
 Datepicker.propTypes = propTypes;
 Datepicker.defaultProps = defaultProps;
 
-export default injectIntl(withStripes(Datepicker));
+export default withStripes(injectIntl(Datepicker));
--- a/lib/InjectIntl/InjectIntl.js
+++ b/lib/InjectIntl/InjectIntl.js
@@ -39,9 +39,14 @@ export default function injectIntl(WrappedComponent, options = {}) {
     }
   }
 
-  return forwardRef((props, ref) => {
-    return (
-      <InjectIntl forwardedRef={ref} {...props} />
-    );
-  });
+  function forward(props, ref) {
+    return <InjectIntl forwardedRef={ref} {...props} />;
+  }
+
+  forward.displayName = componentName;
+
+  if (withRef) {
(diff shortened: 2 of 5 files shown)


rjruizes:
I tried this way initially, but flipped it because injectIntl’s withRef component had no displayName for the withStripes component to get.

cherewaty:
Two commits added:
- setting a `displayName` for `forwardRef`and only using `forwardRef` when `withRef` is true; now there aren't any scenarios in `stripes-components` where `forwardRef` and `withStripes` are direct parent/children
- using `withStripes` in `Settings` - looks like it was just missed, I caught that by running this branch against the eHoldings test suite

JohnC-80:
@cherewaty, @rjruizes Thank you guys for both of your efforts here! Sorry I haven't been able to dig in and formally stamp these.

JohnC-80:
@rjruizes, @cherewaty  So taking a look at what's going on here and in #449 ... would it be best for the withStripes HoC to live in stripes-components and import the context consumer from stripes-core, so that we only have a single reference to stripes-core context to (hopefully) clean up in the future?

rjruizes:
@JohnC-80 You're suggesting we put `StripesContext` in stripes-core and `withStripes` in stripes-components? Can you clarify the advantage? I don't get the need for separation; but I can see withStripes being either a "core" thing or a "component" thing.

I still need to add withStripes to the other modules using legacy context API for stripes:
- stripes-smart-components
- ui-checkout
- ui-organization
- ui-requests

cherewaty:
I think `stripes-core` is the right home for `withStripes()`.

If we adopt the strategy presented in https://issues.folio.org/browse/STCOM-298, `stripes-components` shouldn't need to use `withStripes()` at all. `withStripes()` creates a dependency on `stripes-core` that we don't want `stripes-components` to have.

This dependency chain for using `withStripes`:
`ui-checkout -> stripes-core`
is preferable to:
`ui-checkout -> stripes-components -> stripes-core`

PR #455, opened 2018-06-26
`<Selection>` accessibility bugfix, tests, etc 

JohnC-80:
# Adjustments in this PR...
* `<SelectList>` list overlays UI (like Datepicker and Dropdown) rather than blocking/breaking/reflowing out.
* Refs updated to use new Ref API (React.createRef()).
* `hidden` attribute used on `<SelectList>` to hide tabbable elements for better accessibility (issue in IE11)
* attribute `focusable="false"` added to `<Icon>`'s rendered svg's (otherwise they get focus in IE11)
* adjusted roles to resemble that of [ARIA 1.1 Collapsible Listbox example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html)
* Added validation styles/aria attributes
* Fixed bug in `<SRStatus>` with dangling timeout causing the "cannot update state on unmounted component" error in tests.
* Added numerous tests for Single Select mode (regular and redux-form)

Here's a screencap of screen-reader performance in JAWS/IE11: https://streamable.com/xfdhu

## Still to-do:
* [x] 'translate' all strings

## Now testing with a real browser!
Thanks to @funkymalc we're now running CI tests with normal Chrome rather than the headless version.


cherewaty:
@JohnC-80 I'm just as stumped as you as to why the tests are failing on Jenkins. They all pass for me locally. I'm a little worried that they're running in Headless Chrome in Jenkins, of an unknown vintage (it's just reporting 0.0.0). It'd be nice to get CI running a real browser.

JohnC-80:
Agreed. Thanks @cherewaty  I'll reach out to @funkymalc to see what we can do to make that happen.

PR #459, opened 2018-06-26
Implement basic Pane rendering test

cherewaty:
Renders a `Pane` inside a `Paneset`.

### DIFF
--- a/lib/Pane/tests/Pane-test.js
+++ b/lib/Pane/tests/Pane-test.js
@@ -1,1 +1,25 @@
-import Pane from '../Pane'; // eslint-disable-line no-unused-vars
+import React from 'react';
+import { describe, beforeEach, it } from '@bigtest/mocha';
+import { expect } from 'chai';
+
+import { mountWithContext } from '../../../tests/helpers';
+
+import Paneset from '../../Paneset';
+import Pane from '../Pane';
+import PaneInteractor from './interactor';
+
+describe('Pane', () => {
+  const pane = new PaneInteractor();
+
+  beforeEach(async () => {
+    await mountWithContext(
+      <Paneset>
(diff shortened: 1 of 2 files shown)


lib/Pane/tests/interactor.js
```diff
@@ -0,0 +1,9 @@
+import {
+  interactor,
+  isPresent
+} from '@bigtest/interactor';
+import css from '../Pane.css';
+
+export default interactor(class PaneInteractor {
+  panePresent = isPresent(`.${css.pane}`);
```
wwilsman: So we can make this interactor more reusable, we can do this:

``` javascript
export default interactor(class PaneInteractor {
  static defaultScope = `.${css.pane}`
});
```

And use the default `isPresent` property

``` javascript
expect(pane.isPresent).to.be.true
```

This way when we eventually test multiple panes, we can pass IDs (or data-test attributes) to scope multiple `PaneInteractor`s to their respective pane components

cherewaty: 💯 That's an excellent feature, Mr. @wwilsman.

lib/Pane/tests/Pane-test.js
```diff
@@ -1 +1,25 @@
-import Pane from '../Pane'; // eslint-disable-line no-unused-vars
+import React from 'react';
+import { describe, beforeEach, it } from '@bigtest/mocha';
+import { expect } from 'chai';
+
+import { mountWithContext } from '../../../tests/helpers';
```
wwilsman: Is the context truly needed for a pane / paneset? 😢

cherewaty: Yes, because of the `closeItem` `aria-label` on a `PaneHeader` :(

PR #465, opened 2018-06-28
Set maximum line length to 120

cherewaty:
## Purpose
Very long lines of code can be challenging to read and maintain. We had some lines in `stripes-components` that were particularly unwieldy.

## Approach
Set a maximum length of 120 characters for code using `eslint`'s `maxlen`. The default for `maxlen` is 80 characters, but that's just a holdover from punch cards (_wow, that had some staying power_). 120 is a more modern guideline that should fit into GitHub's code views across platforms without horizontal scrolling.

## Resources
- https://eslint.org/docs/rules/max-len
- https://stackoverflow.com/questions/22207920/what-is-githubs-character-limit-or-line-length-for-viewing-files-on-github
- How to set vertical rule in VSCode: https://stackoverflow.com/questions/29968499/vertical-rulers-in-visual-studio-code
- How to set vertical rule in Atom: https://crunchify.com/atom-preferred-line-length-issue/


doytch:
I like this, but what's the reasoning behind doing this here and not in `eslint-config-stripes`?

cherewaty:
@doytch only limiting the impact. I'd be fine with it in `eslint-config-stripes`, but it didn't seem worth the noise it would cause at the moment.

PR #469, opened 2018-06-30
Refactor ModalFooter tests

rasmuswoelk:
Some minor refactoring of the ModalFotoer tests. Now using the Button-interactor for buttons in the ModalFooter-interactor using the `scoped` method.

I also added "label" to the Button interactor.


rasmuswoelk:
@wwilsman Ah. Good point. But yeah - in this case, I guess it can't hurt to have an alias for the label.

PR #470, opened 2018-07-01
Added test coverage for Modal

rasmuswoelk:
I added some tests for the Modal. I think the coverage is pretty good but let me know if you guys have ideas for more tests for the Modal.

Here's the output of the tests:

image

### DIFF
--- a/lib/Modal/tests/Modal-test.js
+++ b/lib/Modal/tests/Modal-test.js
@@ -11,7 +11,7 @@ import { mountWithContext } from '../../../tests/helpers';
 import Modal from '../Modal';
 import ModalInteractor from './interactor';
 
-describe.only('Modal', () => {
+describe('Modal', () => {
   const modal = new ModalInteractor();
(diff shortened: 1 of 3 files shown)


lib/Modal/tests/Modal-test.js
```diff
@@ -1 +1,95 @@
...
+    it('Modal should be present', () => {
+      expect(modal.isPresent).to.be.true;
+    });
+    it('Modal backdrop should be present', () => {
+      expect(modal.backdrop.isPresent).to.be.true;
+    });
+  });
+
+  describe('When the "showHeader"-prop is set to false', () => {
+    beforeEach(async () => { await mountWithContext(<Modal open label={label} showHeader={false}>Test</Modal>); });
```
wwilsman: This was a little hard to digest at first, probably better across multiple lines

rasmuswoelk: Good point. I'll fix it 👍

PR #471, opened 2018-07-01
ConfirmationModal updates, tests and example

rasmuswoelk:
## Changes
**ConfirmationModal**
- Updated to use the new ModalFooter component
- Added example in Storybook under the "Modal"-tab
- Changed the test ID to be based on the ID prop if provided and a unique ID if it isn't
- Added some simple tests (I will add some more once #470 is merged in and I can use the Modal-interactor)

**Modal**
- Added ID's for sub-elements based on the Modal ID for testing purposes (label, footer, header and close-button)
- Changed label to use Headline with a tag of h1

## Notes
I have changed the shape of the test ID for the ConfirmationModal which of course can cause tests to fail. However, searching the code base showed no signs of tests being dependent on these ID's so these changes should not break anything. 

I have tried running the `ui-testing`-tests to be sure but I'm having issues with these tests currently and a lot of the tests fails when they shouldn't be failing. 

We are in the progress of fixing these issues but until that happens I would appreciate if someone could run the tests for me to make sure that nothing breaks before this gets merged in.

## Test results
image

## Before
image

## After
image


lib/ConfirmationModal/tests/interactor.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * ConfirmationModal interactor
+ */
+
+import { interactor, scoped } from '@bigtest/interactor';
+import ButtonInteractor from '../../Button/tests/interactor';
+
+export default interactor(class ModalFooterInteractor {
+  confirmButton = scoped('#clickable-confirmation-modal-test-confirm', ButtonInteractor);
```
cherewaty: This seems like a good case for `data-test` attributes instead of ids.

You can set the attribute `data-test-confimation-modal-test-confirm` on the element and select it with `'[data-test-confimation-modal-test-confirm]'`. The `stripes` production builds will strip those attributes out, and it prevents situations where you could have multiple things with the same id.

rasmuswoelk: I didn't know that it was an option. Very convenient!

For this case I would have to pass it in the object for the buttons in the ModalFooter like in the screenshot below. To prevent getting linting errors for "Inconsistently quoted property found (quote-props)." I need to add quotes to each property in the object. It's not very pretty :( 

Do you know of a cleaner way to do it?

image

I see your point of ID's potentially popping up more than once. However, in this case the of the ConfirmationModal I think it's very unlikely that it will ever be rendered twice in the same view.

For me it doesn't make a big difference so I'll let you decice. Stick with ID's or switch to data attributes?

cherewaty: We could maybe switch `primaryButton` and `secondaryButton` to accept nodes/elements instead of being data objects? That's probably how I would approach that, but we can tackle that some other time :)

For now, the quoted keys is probably worth it to use a `data-test-` attribute instead of an id.

rasmuswoelk: Hmm. Yeah, we could do that. My reasoning for this approach was that I wanted to make sure that the ModalFooter has a "fixed" design so that the primary/secondary buttons are always in the correct order and also to make it quick and easy to define a ModalFooter without having to import the Button component etc. It's still an option to pass your own custom ModalFooter to the Modal if you want.

But it might be a better solution to pass nodes/elements to the primary/secondary props 🤔I just feel that it leaves room for potential inconsistencies in the modal design pattern.

rasmuswoelk: I am added the data-test attributes for now.

PR #472, opened 2018-07-02
Added tests for Icon

rasmuswoelk:
I added the following tests for the Icon-component:

1. Test if an Icon with no "icon"-prop passed renders the default icon
2. Test all sizes (small, medium and large) to see if the correct classNames are applied and the rendered sizes are matching the CSS
3. Loop over all icons to make sure that the SVG includes the passed iconClassName

**Some of the test log:**

image


lib/Icon/tests/interactor.js
```diff
@@ -0,0 +1,19 @@
...
+const iconClassSelector = selectorFromClassnameString(`.${css.icon}`);
+
+export default interactor(class IconInteractor {
+  static defaultScope = iconClassSelector;
+  width = property('offsetWidth');
+  height = property('offsetHeight');
+  className = attribute('class');
+  svg = scoped('svg', interactor(class SvgInteractor {
+    className = attribute('class');
+  }));
```
wwilsman: If this interactor won't be used elsewhere, a plain object of properties also works as the second argument to `scoped`!

```
svg = scoped('svg', {
  className: attribute('class')
});
```

lib/Icon/tests/Icon-test.js
```diff
@@ -1 +1,94 @@
...
+      beforeEach(async () => {
+        await mount(<Icon icon="clearX" size={size} />);
+      });
+
+      it(`It should render an icon with a className of "${sizeClassName}"`, () => {
+        expect(icon.className).to.include(sizeClassName);
+      });
+
+      it(`It should have a width of ${sizes[size].width}px`, () => {
+        expect(sizes[size].width).to.equal(icon.width);
```
wwilsman: I think this assertion, and the one below, we should reverse the subjects so that we're asserting against the `icon` as oppose to asserting against `sizes`.

rasmuswoelk:
@wwilsman Great idea with an addition to the IconInteractor that allows for checking for specific icons. That could definitely be useful in other component tests.

lib/Icon/tests/Icon-test.js
rasmuswoelk: Thanks for your feedback! I'm not entirely sure what you mean here. Can you give elaborate?

wwilsman: If this produces an error, the error will be about `sizes[size].width` as oppose to the error being about the `icon.width`.

```
expect(icon.width).to.equal(sizes[size].width)
```

would produce an error about the `icon.width` not being the expected size.

rasmuswoelk: Got it 😀👍🏻

PR #480, opened 2018-07-05
UICHKOUT-437 Modal: Catch pointer events

doytch:
#476 started rendering Modals into the `overlayContainer`, but the `overlayContainer` has `pointer-events: none` set on it so all mouse events get sent to what's _behind_ the Modal.

This PR turns pointer-events back on for Modals.


doytch:
Good call, @cherewaty. Done!

PR #481, opened 2018-07-06
STCOM-305: Add link to user record in MetaSection

doytch:
`MetaSection` can now take an Okapi user object as well as a string for the `createdBy` and `updatedBy` props. This object contains the `personal` object from which it will construct the name itself, and the user's `id` with which it will construct a `Link` to that record if the logged-in user has the appropriate permissions.


taras:
@doytch ping ^^

doytch:
That would also be a possibility. The downside then is that you repeat the perm check, name construction and `Link` building with hardcoded strings in every usage. I find that much more likely to get out of sync and difficult to maintain. What if we change the perm requirement? What if we customise/change how names are shown? What if the users endpoint changes? All those changes then require us to go grepping through every module to fix it individually.

Theoretically you could then move those jobs somewhere else into a util function but then it's just blown apart across the filesystem rather than all in one place.

PR #483, opened 2018-07-09
STCOM-284 Remove structures deprecation

cherewaty:
No `folio-org` repos are pointing to components previously found in `lib/structures` anymore.

Resolves https://issues.folio.org/browse/STCOM-284.

Major version bump to 3.0.0. I'll create corresponding tag too on merging.


JohnC-80:
Happy to go 3.0 (once we get the changelog conflict sorted)

PR #487, opened 2018-07-11
Add tests to Datepicker components

taras:
## Purpose

The goal of this PR is to increase confidence in the Datepicker/Calendar/DayButton components by writing tests for them. 

## Approach

I gathered required requirements about this component from @JohnC-80 and @cherewaty. 

**Features that Datepicker should support**
- uses a standard text input - value can be typed as well as picked using the calendar
- picker disappears with a click outside or blur of the input
- presents in localized format (based on locale to determine input and output)
- custom format can be applied
- excluded dates
- a11y - screen reader compatibility (announces actions, selections, values)
- a11y - keyboard navigation of dates
- accepts colors from a stripes theme/works with stripes-components postcss workflow/css-modules
- supports some version of the “passThroughValue” to keep repetitious entries to a minimum. Use-case: checkin - applying today’s date to multiple items within a check-in.  (would have to discuss alternatives with UX design staff)
- validation styles/classes in-line with other stripes-components
- Optional: outputs date in iso format (output value) (redux: format for input, normalize for output) (assumption is that all folio apps use full ISO, eholdings does not because of ebsco)
 
Then I created a list of all props that this component accepts and went through and ensured that all props either have tests or are marked as deprecated. Here is a list of all props that have been covered by the tests.

**Commonly used props**
- [x] intl
- [x] backendDateStandard
- [x] timeZone
- [x] dateFormat
- [x] locale

**HTML Attributes**
- [x] value
- [x] disabled
- [x] label
- [x] id
- [x] required
- [x] onChange
- [x] readOnly

**Deprecated**
- [x] excludeDates
- [x] ignoreLocalOffset

**Not Deprecated in this PR (but probably should)**
- [x] passThroughValue - when value matches specified value, just pass the value through

## Learning

- [x] Create an issue of accessibility issues in this component - [STCOM-315](https://issues.folio.org/browse/STCOM-315)
- [x] `prop-types-extra` package has [deprecated](https://www.npmjs.com/package/prop-types-extra#deprecatedvalidator-reason) prop type.
- [x] we might want to consider using [react-dates](http://airbnb.io/react-dates/?selectedKind=DateRangePicker%20%28DRP%29&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) instead of fixing this component or writing something from scratch using [downshift](https://github.com/paypal/downshift)

## Screenshots

screen shot 2018-07-18 at 9 19 59 pm

### DIFF
--- a/lib/Datepicker/Datepicker.js
+++ b/lib/Datepicker/Datepicker.js
@@ -22,11 +22,7 @@ const propTypes = {
   date: PropTypes.object,
   dateFormat: PropTypes.string,
   disabled: PropTypes.bool,
-  excludeDates: PropTypes.oneOfType([
-    PropTypes.array,
-    PropTypes.object,
-    PropTypes.string,
-  ]),
+  exclude: PropTypes.func,
   hideOnChoose: PropTypes.bool,
   id: PropTypes.string,
   ignoreLocalOffset: PropTypes.bool,
@@ -449,7 +445,7 @@ class Datepicker extends React.Component {
     if (/2822/.test(backendDateStandard)) {
       return parsed.format(DATE_RFC2822);
     }
-    
+
     return parsed.format(this.props.backendDateStandard);
   }
 
@@ -466,7 +462,7 @@ class Datepicker extends React.Component {
       screenReaderMessage,
       label, id, readOnly,
       required,
-      excludeDates,
+      exclude,
       disabled,
       autoFocus,
       tether,
@@ -590,7 +586,7 @@ class Datepicker extends React.Component {
               onKeyDown={this.handleKeyDown}
               onBlur={this.hideCalendar}
               locale={this.locale}
-              excludeDates={excludeDates}
+              exclude={exclude}
               onNavigation={this.handleDateNavigation}
               id={this.testId}
             />
--- a/lib/Datepicker/Datepicker.js
+++ b/lib/Datepicker/Datepicker.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { intlShape } from 'react-intl';
 import PropTypes from 'prop-types';
+import { deprecated } from 'prop-types-extra';
 import TetherComponent from 'react-tether';
 import RootCloseWrapper from 'react-overlays/lib/RootCloseWrapper';
 import moment from 'moment-timezone';
@@ -8,7 +9,6 @@ import contains from 'dom-helpers/query/contains';
 import debounce from 'lodash/debounce';
 import uniqueId from 'lodash/uniqueId';
 import { withStripes } from '@folio/stripes-core/src/StripesContext';
-import deprecated from 'deprecated-prop-type';
 
 import injectIntl from '../InjectIntl';
 import Button from '../Button';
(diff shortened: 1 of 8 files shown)


lib/Datepicker/tests/Datepicker-test.js
```diff
@@ -96,76 +129,204 @@ describe('Datepicker', () => {
...
-  describe('coupled to redux form', () => {
-    describe('selecting a date', () => {
-      let dateOutput;
+  describe('passThroughValue', () => {
+    let onBlur;
+    let onChange;
 
+    beforeEach(async () => {
+      onBlur = mock.fn();
+      onChange = mock.fn();
```
cherewaty: The jest mocks feel a little unit-test-y vs. BigTest-y, but their use makes sense.

lib/TextField/stories/BasicUsage.js
```diff
@@ -5,14 +5,22 @@
...
-      required
     />
+    <form>
+      <TextField
+        label="Required field"
+        placeholder="Placeholder Text"
+        required
+      />
+      <Button type="submit">Submit</Button>
+    </form>
```
cherewaty: What's the additional `<form>` and `<Button>` illustrating on the `<TextField>` story?

lib/Datepicker/Datepicker.js
```diff
@@ -317,20 +321,24 @@ class Datepicker extends React.Component {
...
   }
 
-  handleSetDate(e, date, stringDate) {
+  handleSetDate = (e, date, stringDate) => {
     if (e.type === 'click' || e.type === 'keydown') {
       e.preventDefault();
 
       if (this.props.onChange) { this.props.onChange(e); }
-      this.textfield.current.getInput().focus();
+      // this.textfield.current.getInput().focus();
```
cherewaty: Leave as-is, delete, or uncomment?

lib/Datepicker/tests/Datepicker-Redux-test.js
```diff
@@ -0,0 +1,572 @@
...
+    let onChange;
+
+    beforeEach(async () => {
+      onChange = mock.fn();
+
+      await mountWithContext(
+        <Datepicker
+          input={{ onChange }}
+          meta={{}}
+        />
```
cherewaty: I've preferred testing `redux-form` like `<Field component={Datepicker} onChange={onChange} />`.

lib/TextField/stories/BasicUsage.js
taras: It was to test the required functionality. I wanted to see what it takes to get a validation error. It's essentially submit that causes the validation error to be displayed when an input is required.

lib/Datepicker/Datepicker.js
taras: I'm going to to it back in. It's a left over from accessibility work that I started but it'll need to happen in a separate PR.

lib/Datepicker/tests/Datepicker-Redux-test.js
taras: Ok, I applied `<Field>` to all redux tests

lib/Datepicker/Datepicker.js
taras: Removed

PR #491, opened 2018-07-13
Upgrade bigtest packages

cherewaty:
Some of the `Button` tests started throwing errors:
```
Error: convergent assertion encountered a async function or promise; since convergent assertions can run multiple times, you should avoid introducing side-effects inside of them
```

Fixed those 👌


cherewaty:
@taras this might help you with with getting better errors from failed tests

PR #490, opened 2018-07-12
preliminary icon tests

zburke:

PR #492, opened 2018-07-16
Revert "preliminary icon tests"

zburke:
Reverts folio-org/stripes-components#490 since #472 was already pending and approved for like fifty years.


rasmuswoelk:
Awesome, @zburke 🙌🏼

Some of your tests covered stuff that I hadn't covered so I'll copy those ones into my branch and merge them in. Cool? 🙂

PR #493, opened 2018-07-16
Update translation strings as of 20180716T200628

folio-translations:


translations/stripes-components/en.json
```diff
@@ -1,58 +1,51 @@
...
+    "showLessAddresses": "Show less",
+    "showMoreAddresses": "Show more addresses",
+    "showOrHideDatepicker": "Show\/hide datepicker",
+    "showOrHideTimepicker": "Show\/hide timepicker",
+    "stateProviceOrRegion": "State\/Prov.\/Region",
+    "submit": "Submit",
+    "tableColumnCount": "Table with {count} columns",
+    "validatingInProcess": "Validating...",
+    "zipOrPostalCode": "Zip\/Postal Code"
+}
```
JohnC-80: Appears we're missing some of the translations. Looks like the ones for `selection.*` are absent from the new set. The difference in lines tipped me...

JohnC-80: As well as `button.*`

translations/stripes-components/en.json
```diff
@@ -1,58 +1,51 @@
...
+    "metaSection.source": "Source: {source}",
+    "primaryAddress": "Primary address",
+    "removeFields": "remove fields for {item} {num}",
+    "removeSelection": "Remove selection",
+    "saveChangesToThisItem": "Save changes to this item",
+    "screenReaderLabel": "{label} format is {screenReaderFormat} use arrow keys to navigate dates. Enter key to confirm. {screenReaderMessage}",
+    "selectDay": "Select {dayOfWeek}, {date}",
+    "showLessAddresses": "Show less",
+    "showMoreAddresses": "Show more addresses",
+    "showOrHideDatepicker": "Show\/hide datepicker",
```
zburke: Why are the forward slashes escaped?

translations/stripes-components/en.json
```diff
@@ -1,58 +1,51 @@
...
+    "showOrHideTimepicker": "Show\/hide timepicker",
+    "stateProviceOrRegion": "State\/Prov.\/Region",
```
zburke: Typo in this key, but that's not new. Can we add the correct key, fix our apps, and then remove this?

translations/stripes-components/en.json
dltj: _sigh_: there is an export option for this:

> Unescape forward slashes: Turn on to skip escaping forward slashes.

The default is to escape forward slashes.  I've fixed the export API call parameters.

dltj: Wow -- the description of the parameter in the API call is even more mentally twisting:

> Enable to leave forward slashes unescaped.

translations/stripes-components/en.json
dltj: Thanks, John -- I found that the automated notification to Lokalise on merge-to-master is _not_ automatically set up.  It needs to be set up for each GitHub repository.  I'll go through and do that.  In the meantime, I'm going to do a manual fetch from GitHub and create a new pull request on stripes-components.

folio-translations:
New attempt at #499

PR #498, opened 2018-07-17
Change order of ModalFooter buttons

cherewaty:
Uses `flex-direction` to show the primary button:
- first in markup
- first on small screens
- last on wider screens

I also replaced `gutter` with `controlMarginBottom`, which reduced the vertical spacing between buttons from 15px to 12px. Fits in better with other form elements.

### Narrow screen
localhost_9001__selectedkind modal selectedstory modalfooter addons 1 stories 1 panelright 0 addonpanel react_storybook 2freadme 2fpanel iphone 5_se

### Wider screen
screen shot 2018-07-17 at 6 15 18 pm

### Wider screen, RTL
screen shot 2018-07-17 at 6 13 05 pm

### DIFF
--- a/lib/Modal/Modal.css
+++ b/lib/Modal/Modal.css
@@ -34,9 +34,8 @@
   border-top: 1px solid var(--color-border-p2);
   padding: var(--gutter);
   display: flex;
-  justify-content: flex-end;
   align-items: center;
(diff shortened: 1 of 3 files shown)


rasmuswoelk:
Nice 👍

PR #501, opened 2018-07-18
change size of a small icon

sduvvuri-ebsco:
- In order to use this Icon in ui-eholdings app, we want size "small" to be 14x14 since that matches with font size. 
- We are using this Icon component as part of an external link component, which displays an external link inline with this icon.
- Open question is whether changing this size affects other apps in a bad way, and if we want to introduce another size, possibly x-small that represents 14x14 and leave small as 16x16.

### DIFF
--- a/lib/Icon/Icon.css
+++ b/lib/Icon/Icon.css
@@ -40,8 +40,9 @@
  */
 
 .small {
-  height: 16px;
-  min-width: 16px;
+  height: 14px;
+  width: 14px;
+  min-width: 14px;
 }
 
 .medium {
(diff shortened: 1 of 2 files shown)


MikeTaylor:
@filipjakobsen I imagine you will have thoughts on globally changing the size of all "small icons" in FOLIO for the convenience of a single app :-)

cherewaty:
@MikeTaylor Filip specifically asked for that icon to be 14x14, so we're just trying to resolve the discrepancy here. It seemed odd to have both 14x14 and 16x16 icons.

MikeTaylor:
That's weird! I guess we'll have to wait for his clarification.

rasmuswoelk:
The icons should be 14x14px 👍 Approved.

PR #506, opened 2018-07-20
Settings: Expose nav pane width

doytch:
[UIORG-84](https://issues.folio.org/browse/UIORG-84) is a bit of an umbrella issue for all the scenarios where the pane sizing on Settings page is a bit off. 

PR #486 handled some of these scenarios but only solves the problem for Settings pages that use an EntrySelector. Pages that don't include the Locations pages that are specifically called out in UIORG-84, along with any page that uses `<ControlledVocab>`. 

Those three-pane pages are made cleaner by shrinking the nav pane that contains the links to a module's settings pages. I've exposed the width of that pane via `<Settings>.navPaneWidth`.

**Feedback requested about whether `navPaneWidth`'s default should be changed to something like 30% instead of keeping the same behaviour that we currently have (`fill`)**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change history for stripes-components
 
 ## 4.0.0 (IN PROGRESS)
+* Expose `Settings` nav pane width via `navPaneWidth` prop.
 
 ## [3.0.0](https://github.com/folio-org/stripes-components/tree/v3.0.0)
 
--- a/lib/Settings/Settings.js
+++ b/lib/Settings/Settings.js
@@ -83,7 +83,7 @@ class Settings extends React.Component {
     return (
       <Paneset nested defaultWidth="80%">
         <Pane
-          defaultWidth="fill"
+          defaultWidth={props.navPaneWidth}
           paneTitle={props.paneTitle || 'Module Settings'}
           firstMenu={(
             <Link to="/settings" className={css.settingsBackButton}>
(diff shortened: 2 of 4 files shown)


doytch:
Readme updated and set the default to `30%`.

You're right @JohnC-80, it does mean that the little pane title lower border doesn't stretch across, but it's not really noticeable. Also, I think it looks better now that the title ("Organization" for example), doesn't start way to the right of the nav links and then snap back left when a settings page is selected.

cherewaty:
The `defaultWidth` prop of a `Pane` sets the `flex-basis`, but there's no way to set its `flex-shrink` or `flex-grow`.

Seems like if we had a way to set `flex-grow`, we could solve this the way the CSS flexbox implementation intends. The final pane would have something like `flex: 1 0 320px` to set a minimum width of 320px and allow the pane to grow as necessary. Example: https://github.com/folio-org/ui-eholdings/blob/f2bcf3d7e57d975472e9f3d05d5f4a3dda54d560/src/components/results-pane/results-pane.css
☝️ that gap in the use of `flexbox` is the reason `ui-eholdings` doesn't use `<Pane>`s in its search layout.

JohnC-80:
Panes used in conjunction with Panesets allow for defaultWidths of "fill" to divide leftover space. Pane resizing has always been in the cards for FOLIO apps. If it weren't for the necessity of user control there, I feel it would be fine to allow free reign over a Pane's width/flexbox capabilities - with the resizable aspect in mind, the best setting props might do is apply initial values - the flexbox settings at the pane level go out of the window once pane width becomes a user preference. Despite this imposed limit, flexbox still handles a great deal for us in its automatic rtl support. For certain situations where resize is not a necessity (like settings navigation menus) I can see specific components that do make better use of flexbox capabilities.

PR #507, opened 2018-07-20
New component: IconWithText

rasmuswoelk:
IconWithText simply renders an icon (Icon or AppIcon — depending on which props you pass to it) with an inline label. Additional props can help ensure proper spacing and you can adjust the positioning of the icon.

This provides a consistent way of displaying inline icons with a label across the platform. It's RTL-compatible and can be used inside other existing components (such as the Button, MCL columns etc.).

The reasoning behind the creation of this component can be found in various prototypes such as the Inventory-prototype (http://ux.folio.org/prototype/en/inventory) which heavily relies on displaying icons with labels in a consistent way (properly inlined text/icon, correct spacing etc.). A future user setting for hiding labels to save space (e.g. in the MCL) is (afaik) on the drawing table which this component also could be extended for to easily allow for this.

Check out the readme [here](https://github.com/samhaeng/stripes-components/blob/7557f6b08f38e37c05ff1ab819aac9a47379e8a0/lib/IconWithText/readme.md) and a basic usage example [here](https://github.com/samhaeng/stripes-components/blob/7557f6b08f38e37c05ff1ab819aac9a47379e8a0/lib/IconWithText/examples/BasicUsage.js).

## Example
image

## Tests
image

### DIFF
--- a/lib/AppIcon/tests/interactor.js
+++ b/lib/AppIcon/tests/interactor.js
@@ -8,6 +8,8 @@ import css from '../AppIcon.css';
 
 const appIconClassSelector = selectorFromClassnameString(css.appIcon);
 
-export default interactor(class AppIconInteractor {
+@interactor class AppIconInteractor {
   static defaultScope = appIconClassSelector;
-});
+}
+
+export default AppIconInteractor;
--- a/lib/AppIcon/tests/interactor.js
+++ b/lib/AppIcon/tests/interactor.js
@@ -1,0 +1,13 @@
+/**
+ * AppIcon interactor
+ */
+
+import { interactor } from '@bigtest/interactor';
+import { selectorFromClassnameString } from '../../../tests/helpers';
+import css from '../AppIcon.css';
+
+const appIconClassSelector = selectorFromClassnameString(css.appIcon);
+
+export default interactor(class AppIconInteractor {
+  static defaultScope = appIconClassSelector;
+});
--- a/lib/IconWithText/tests/IconWithText-test.js
+++ b/lib/IconWithText/tests/IconWithText-test.js
@@ -1,0 +1,31 @@
+/**
+ * IconWithText tests
+ */
+
+import React from 'react';
+
+import { it, describe, beforeEach } from '@bigtest/mocha';
+import { expect } from 'chai';
+import { mount } from '../../../tests/helpers';
+
+import IconWithText from '../IconWithText';
+import IconWithTextInteractor from './interactor';
+
+describe.only('IconWithText', () => {
+  const iconWithText = new IconWithTextInteractor();
+  const text = 'Hello World';
+  const standardIcon = 'closeX';
+
+  beforeEach(async () => {
+    await mount(
+      <IconWithText
+        text={text}
+        icon={standardIcon}
+      />
+    );
+  });
+
+  it(`Should render with a text of "${text}"`, () => {
+    expect(iconWithText.text).to.equal(text);
+  });
+});
--- a/lib/IconWithText/tests/IconWithText-test.js
+++ b/lib/IconWithText/tests/IconWithText-test.js
@@ -11,7 +11,7 @@ import { mount } from '../../../tests/helpers';
 import IconWithText from '../IconWithText';
 import IconWithTextInteractor from './interactor';
 
-describe.only('IconWithText', () => {
+describe('IconWithText', () => {
   const iconWithText = new IconWithTextInteractor();
   const text = 'Hello World';
   const standardIcon = 'closeX';
--- a/lib/IconWithText/tests/IconWithText-test.js
+++ b/lib/IconWithText/tests/IconWithText-test.js
@@ -25,7 +25,32 @@ describe.only('IconWithText', () => {
     );
   });
 
+  /**
+   * Icon and default tests
+   */
   it(`Should render with a text of "${text}"`, () => {
     expect(iconWithText.text).to.equal(text);
+  });
+
+  it('Should render an <Icon /> if a valid icon string is passed to the "icon"-prop', () => {
+    expect(iconWithText.regularIcon.isPresent).to.be.true;
+  });
+
+  /**
+   * App Icon
+   */
+  describe('Passing a string to both the "app"-prop', () => {
+    beforeEach(async () => {
+      await mount(
+        <IconWithText
+          text={text}
+          app="users"
+        />
+      );
+    });
+
+    it('Should render an <AppIcon />', () => {
+      expect(iconWithText.appIcon.isPresent).to.be.true;
+    });
   });
 });
--- a/lib/IconWithText/tests/IconWithText-test.js
+++ b/lib/IconWithText/tests/IconWithText-test.js
@@ -33,7 +33,7 @@ describe('IconWithText', () => {
   it(`Should render with an ID of "${id}"`, () => {
     expect(iconWithText.id).to.equal(id);
   });
-  
+
   it(`Should render with a text of "${text}"`, () => {
     expect(iconWithText.text).to.equal(text);
   });
@@ -51,12 +51,45 @@ describe('IconWithText', () => {
         <IconWithText
           text={text}
           app="users"
+          padding="end"
         />
       );
     });
 
     it('Should render an <AppIcon />', () => {
       expect(iconWithText.appIcon.isPresent).to.be.true;
+    });
+  });
+
+  /**
+   * Padding
+   */
+  describe('When passing a string to the padding-prop', () => {
+    it('Should have classname of "padding-start" when padding="start"', () => {
+      mount(
+        <IconWithText
+          padding="start"
+        />
+      );
+      expect(iconWithText.hasPaddingStartClass).to.be.true;
+    });
+
+    it('Should have classname of "padding-end" when padding="end"', () => {
+      mount(
+        <IconWithText
+          padding="end"
+        />
+      );
+      expect(iconWithText.hasPaddingEndClass).to.be.true;
+    });
+
+    it('Should have classname of "padding-both" when padding="both"', () => {
+      mount(
+        <IconWithText
+          padding="both"
+        />
+      );
+      expect(iconWithText.hasPaddingBothClass).to.be.true;
     });
   });
 });
--- a/lib/IconWithText/tests/IconWithText-test.js
+++ b/lib/IconWithText/tests/IconWithText-test.js
@@ -11,7 +11,7 @@ import { mount } from '../../../tests/helpers';
 import IconWithText from '../IconWithText';
 import IconWithTextInteractor from './interactor';
 
-describe('IconWithText', () => {
+describe.only('IconWithText', () => {
   const iconWithText = new IconWithTextInteractor();
   const text = 'Hello World';
   const id = 'my-test-id';
@@ -30,15 +30,15 @@ describe('IconWithText', () => {
   /**
    * Icon and default tests
    */
-  it(`Should render with an ID of "${id}"`, () => {
+  it(`Renders with an ID of "${id}"`, () => {
     expect(iconWithText.id).to.equal(id);
   });
 
-  it(`Should render with a text of "${text}"`, () => {
+  it(`Renders with a text of "${text}"`, () => {
     expect(iconWithText.text).to.equal(text);
   });
 
-  it('Should render an <Icon /> if a valid icon string is passed to the "icon"-prop', () => {
+  it('Renders an <Icon /> if a valid icon string is passed to the "icon"-prop', () => {
     expect(iconWithText.regularIcon.isPresent).to.be.true;
   });
 
@@ -56,7 +56,7 @@ describe('IconWithText', () => {
       );
     });
 
-    it('Should render an <AppIcon />', () => {
+    it('Renders an <AppIcon />', () => {
       expect(iconWithText.appIcon.isPresent).to.be.true;
     });
   });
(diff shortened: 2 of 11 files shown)


lib/AppIcon/tests/interactor.js
```diff
@@ -0,0 +1,13 @@
...
+ * AppIcon interactor
+ */
+
+import { interactor } from '@bigtest/interactor';
+import { selectorFromClassnameString } from '../../../tests/helpers';
+import css from '../AppIcon.css';
+
+const appIconClassSelector = selectorFromClassnameString(css.appIcon);
+
+export default interactor(class AppIconInteractor {
```
cowboyd: We can use decorators here, right?

export default @interactor class AppIconInteractor {
}

lib/IconWithText/tests/IconWithText-test.js
```diff
@@ -0,0 +1,62 @@
...
+        icon={standardIcon}
+        id={id}
+      />
+    );
+  });
+
+  /**
+   * Icon and default tests
+   */
+  it(`Should render with an ID of "${id}"`, () => {
```
cowboyd: Stylistic, but typically, we opt for a stronger statement than "should":

it renders with an id of X
it renders with a text of T
it renders an AppIcon

taras:
Hi @rasmuswoelk, it's cool to see BigTest being used :) I wanted to ask you a bit more about the motivation for these components. 

Usually, I reach for a common component when it takes longer to implement what I want than it is to find a component. For example, select components be difficult so I would look in a library before trying to implement it myself.

For things that can be done with a few lines of code, it's usually easier to just write them. This might be a different story if the UI was so complex that it was easier to position an Icon & Text using a component rather than CSS. For example, in React Native it might be necessary to create a component that positions icons and text together. 

## Alternatives

1. How would implement this without a component?
2. Have you considered a component that could be used to layout things inline regardless of what they are? For example, `<Inline><Icon icon="bookmark"/> Bookmark</Inline>` or `<Inline>Bookmark <Icon icon="bookmark"/></Inline>` can provide same functionality as 

```jsx
<IconWithText
    text="Bookmark"
    icon="bookmark"
    iconPosition="end"
    paddingEnd
/>
```

Look forward to hearing your thoughts.

lib/AppIcon/tests/interactor.js
rasmuswoelk: I like the decorator syntax better. Only thing though: If I do it like you propose it will give me a linting error of `AppIconInteractor is declared but never used` so I had to move the `export default` to the bottom of the file to avoid this warning.

lib/IconWithText/tests/IconWithText-test.js
rasmuswoelk: Good point. I have updated the wording a bit.

rasmuswoelk:
Hi @taras 

Thank you for your comments. It's great with this kind of feedback!

BigTest is really awesome. I enjoy writing tests using the re-usable iterators.

The motivation for creating this component came from looking at a lot of different prototypes/designs that features icons with an associated label. Since it will be a very common design pattern across the system, we decided to make a specific component for this purpose. 

You could implement the same functionality without the component but it would require a little bit of CSS to make it look right. I like your idea of having a more universal component to do inlining and we probably could use a utility component like that as well.

For this case.. Personally, I like to have specific components for common patterns like these because I know they will pop up a lot of places in the app and having a specific component ensures full control of the styling thereby preventing inconsistencies from occurring while being easy and straight-forward to implement (for those who doesn't like CSS). Some of the things to consider are the spacing between the label and the icon as well as the RTL support (margins/padding stay the same after being flipped so this needs to be dealt with also).

Also, as I mention in my PR description, it's very likely that we will have a user-setting in the future that allows for hiding either labels or icons in some places in the app (to save space). Here it will come in handy to have a more specific component like this that could be updated to support this functionality.

I hope that answered your questions. If not, then let me know!

taras:
Hi @rasmuswoelk,

> The motivation for creating this component came from looking at a lot of different prototypes/designs that features icons with an associated label. Since it will be a very common design pattern across the system, we decided to make a specific component for this purpose.

The motivation seems valid, but I'm trying to figure out if this particular component is the right approach to addressing this particular requirement.

> You could implement the same functionality without the component but it would require a little bit of CSS to make it look right. 

What specifically would this CSS look like?

> For this case.. Personally, I like to have specific components for common patterns like these because I know they will pop up a lot of places in the app and having a specific component ensures full control of the styling thereby preventing inconsistencies from occurring while being easy and straight-forward to implement (for those who doesn't like CSS).

The challenge with seeing components as a solution is that sometimes they create an inaccurate sense of progress. When you expose the API, it feels like you made progress but you might not aware of long tail consequences of adding the API surface. 

For example, you might address inlining and vertical position of elements in this component but it doesn't mean that it'll work in most cases that people need to use it in. When issues around these components come up, they'll be issues that a developer will experience based on a context specific implementation. 

In that case, they're unlikely to reach out and file an issue in Jira for this component. They'll just work around the issue in their code. The result is that your abstraction is not being used because it might not be comprehensive enough but there is sense that progress was made.

When making components, especially once that can be easily written by hand, you might be running the risk of creating an abstraction were one is not necessary or the understanding of the problem is not ready to be generalized.

> Also, as I mention in my PR description, it's very likely that we will have a user-setting in the future that allows for hiding either labels or icons in some places in the app (to save space). Here it will come in handy to have a more specific component like this that could be updated to support this functionality.

Regardless what happens in the future, showing icons will likely be a context specific thing. I don't think you can just disable icons globally. If anything, you might use the icon instead of text in touch environments. Visibility like this is usually based on a Media Query which is usually implemented with CSS. If you're doing this through state, then it *has* to be context specific because the prop needs to be bound to some state . 

As I said, I'm not saying that this use case is invalid, but the abstraction has to have a strong reason to exist. You might find this article called [RYE: Repeat Yourself Enough](https://frontside.io/blog/2009/06/29/rye-repeat-yourself-enough.html) interesting. It talks to finding the right time to create abstractions.

If you did not have ability to merge into this repo and you had to convince someone of the need for this. How would you do it? You would need to propose why this solution is the right solution to this particular problem. So convince us without relying on hypothetical future situations. If you need to rely on these to defend the implementation then this PR should wait until those use cases are flashed out.

rasmuswoelk:
Hi @taras 

Just like the component itself, the CSS is fairly simple. It's mostly about the correct spacing between the icon and the label as well as the optional spacing on the left/right sides of the component. The latter makes it easier to add proper spacing between multiple IconWithText's but can be useful in other situations where spacing is needed.

Using an AppIcon requires a bit different margins than using a regular Icon because most icons have some white space inside the SVG itself.

If we take the preview pane in Inventory as an example. It currently looks like this:

image

In this case we would need to add some margins to ensure proper spacing. Let's assume we have a component to inline things.

The markup would look like this:
```jsx
 <Inline>
   <AppIcon app="inventory" iconKey="instance" className={css.myIcon} />
   <span className={css.myLabel}>Instance Record</span>
   <AppIcon app="inventory" iconKey="text" className={css.myIcon} />
   <span className={css.myLabel}>Text</span>
 </Inline>
```

We need some CSS to add margin between the icon and the label. We use `em` to make sure that it adjusts to the parent font-size.

```css
.myIcon {
  margin: 0 0 0 0.4em;
}

/* Don't forget RTL. We need to flip margins */
[dir=rtl] .myIcon {
  margin: 0 0.4em 0 0;
}
```

We also need some margin between the label and the next icon. So we add a bit of CSS for this as well.

```css
.myLabel {
  margin: 0 0 0 0.8em;
}

[dir=rtl] .myLabel {
  margin: 0 0.8em 0 0;
}
```

It's not a lot of CSS we need to make this look how it's supposed to look. 

image

Now we just have to repeat this in all parts of the app that display icons in this way. You can click around this prototype and see how the pattern repeats: 
http://ux.folio.org/universal-navigation/uxi-3.

Or see these prototypes for more use-cases:
http://ux.folio.org/prototype/en/inventory
http://ux.folio.org/prototype/en/calendar

These use-cases mostly revolves around displaying AppIcon's with labels. Other use-cases that definitely will arise: Displaying icons inside buttons, list items, menus, KeyValue's etc.

It's definitely an option to put this CSS somewhere in stripes-components and document this as a pattern. Since we don't use global class names, every developer would of course need to import the class names from some CSS file and apply them accordingly. This is the only alternative I can think of right now. Maybe you have a better idea?

As for convincing why this component is needed.. Well, I feel that I already made my points in previous comments. The primary point being consistency in the design. And then I guess convenience for developers. Given the lazy nature of developers I feel that a component like this provides a way of displaying this particular —  but very much repeated —  design patteren in a simple and non-tedious way. Just like any other stripes components you are of course free to make your own custom solution if this component doesn't fit your exact needs.

The feature of being able to hide the label or the icon is a future thing, yes, but not hypothetical. Whether or not it would be solved using a component like this or not.. well, I don't know. It was meerely a suggestion of how this component also could be useful in the future. I could imagine that there would be some accessibility-related things to consider when hiding either label or icon that could be handled behind the scenes in a component like this.

I would like to note that I consulted @JohnC-80 before I started working on this component and we came to the conclusion that a component like this would be convenient to have in our toolbox.

JohnC-80:
@rasmuswoelk  thank you for your patience as we discuss this before the commitment here is solid. Every day it's a new detail to truly get things right.
From an FOLIO app developer perspective, is there something we can do better to make the specific use-cases more obvious? Rather, is there something we must do to keep developers from misusing it? Are there cases where an icon and text are needed where we wouldn't use this component for any reason? For example, say we *did* do some under-the-hood hiding of the label - that strikes me as a very risky thing if the component has been used to display important info as text - so those instances are the "misuse cases".

I don't think devs are lazy as much as they've got timelines to meet, so they can't always dig into CSS either due to the bandwidth it takes to learn or the bandwidth it takes to tweak styles - either way, the convenience a component provides has to be truly pure - no unmentioned gotcha's with necessary flexibilities intact. Say they supply a custom image for an icon that may require some tweaking of the margin instead of a built-in one - is there a way to allow that?

@taras is there a principle of component/library design that we can put our finger on that's truly missing here?

taras:
@rasmuswoelk thank you for taking the time to answer my questions. 

In retrospect, I didn't give this issue enough focus to clearly articular what problem I see with it.  There seemed something off, but I couldn't quiet put my finger on it. I spent some time looking at it now and I'm left with the impression that we can probably get away with what we already have without having to introduce another component.

Icon & AppIcon are already inline elements, so they don't need anything special to work as inline elements. To turn an Icon or an AppIcon into a IconWithText, all you have to do is pass text as children to  the component. The API change that we need to make is to allow Icon & AppIcon to render children props. 

```js
<AppIcon icon="inventory">Inventory</AppIcon>
```

The spacing can be control with actual space character or with css that's part of the App/Icon components. We can add flags to the components to add spacing on right and left.

```js
<AppIcon icon="inventory" left-margin right-margin> Inventory</AppIcon>
```

If we want to get clever, we can allow `left-margin ` & `right-margin` to accept an optional value which can be used to configure the style attribute of the icon. 

```js
<AppIcon icon="inventory" left-margin="1em" right-margin> Inventory</AppIcon>
```

This will likely be flexible enough for us without introducing another component.

> @taras is there a principle of component/library design that we can put our finger on that's truly missing here?

I'm not sure if I know a name for it but we need to think thrice before creating component because they become a liability that we need to maintain. If something can be accomplished with composition, then we should aim to do that rather than creating another component to handle that case.

rasmuswoelk:
Hi @taras 

Sorry for not getting back to you sooner.

I kinda agree that we can utilize the children-prop or add an optional label-prop to the Icon- and AppIcon-component. It would achieve the same result.

I think my initial thinking for creating a new component for this purpose was to not add more props to the existing components to try to keep them as simple and easy to understand as possible.

As you mention we would then also have to add the spacing props to both components. I would not recommend using inline styling for this purpose (too specific and difficult to overwrite using regular CSS) because we want to be in full control of the styling.

And in my opinion we also need to add the `iconPlacement`-prop to make it possible to change the positioning of the icon/label in an RTL compliant way. This prop would of course not make sense if there's no label added. Alternatively, we could call it "labelPlacement" instead to make it clear that it's related to the label-prop.

If we can agree on this approach instead and @JohnC-80 approves, we can scrap this component and I can add the props and functionality to the existing AppIcon- and Icon-components. It's the same for me — I'm really just interested in being able to implement labels to icons in a proper way across the system as soon as possible.

JohnC-80:
I like the approach of rendering children. I feel like in this case, the children would be in charge of their own spacing in relation to the Icon, so they could potentially use a `<Layout>` component to provide some sort of utility style if the spacing does need some adjustment.

rasmuswoelk:
@JohnC-80 Alright. I'll submit a PR that adds the option of children to Icon and AppIcon.

However, the AppIcon already has an option to pass an image as a child. I can make a label-prop instead in this case but should I make it a label-prop for both components? (For consistency)

JohnC-80:
Hmm... any idea on what the frequency of use is for AppIcon children?

rasmuswoelk:
@JohnC-80 Come to think of it.. I only think it's used in the examples. And that's mostly because the stripes core context isn't available in Storybook.

So maybe I should just remove that logic and use it for the label instead?

JohnC-80:
At this stage, I feel that's the best path. If you wanted to substitute the current `children` functionality on AppIcon with an additional prop - `external-icon`? that might work... but really I'm not sure if we have a strong use-case for that functionality at all, just speculations...

rasmuswoelk:
@JohnC-80 I agree. I don't think it's needed currently and most likely it won't be necessary. We can always add it if it becomes a necessity.

rasmuswoelk:
We went in a different direction - adding a label option to both Icon and AppIcon instead.

PR #513, opened 2018-07-24
Change button relationship margins

cherewaty:
## Purpose
This CSS in `button.css`...
```
&:first-child {
  margin-left: 0;
}
```
... meant that any button that's a first child of its parent (even if it's by itself) would have its margin overridden.

The specific case I ran into: I wanted to put a button on the `firstMenu` of a `<PaneHeader>` with a `buttonClass` that defined some margins. 

`:first-child` was more specific than my class, so I couldn't set a left margin:

localhost_3000_eholdings_titles_new ipad

## Approach
Rely on relationships between buttons instead.

Set left and right margins of `<Button>`s to zero by default, then buttons that are next to each other get 4px between them (in both LTR and RTL languages).


cherewaty:
I want good spacing primitives instead too :) We'll get there!

PR #514, opened 2018-07-29
Add printer icon

cchiama:
Add printer icon


MikeTaylor:
Why do we define all the SVG icons in a single JavaScript file, instead of putting each in an SVG file?

JohnC-80:
@mike We currently render the SVG's in-line and apply a few aspects dynamically (style, classnames) to control the appearance - which is where the inline-rendering benefits us over other methods (`` tag or css `background-image`). As that list does continue to grow, we may look at more dynamic ways of handling it than storing all of them as keys on a single object.

JohnC-80:
Thank you @cchiama !
Would you mind cleaning up those lint errors that are showing up?
You can run the linter locally by running `yarn lint` from the command line in your `stripes-components` folder. It looks like it just needs a space before those closing brackets in the JSX.

MikeTaylor:
@JohnC-80 Are you saying that there are CSS style tweaks you can make to inline SVG that you can't  can't make when pulling in a separate SVG file?!

JohnC-80:
@MikeTaylor That's correct. When an SVG is rendered within one of those elements that doesn't place its code directly into the DOM, there's not much you can do to affect what appears inside of that SVG's markup.

cchiama:
I nerd other icons for marcati module, because our state application have more complexity, so the the porting is not easy @MikeTaylor @JohnC-80

JohnC-80:
@cchiama  I need one more thing - could you please electronically sign the Contributor License Agreement? You just need to follow the link in this comment: https://github.com/folio-org/stripes-components/pull/514#issuecomment-408662814 I spoke to the design guys who handle our icons and they're on board with merging this... we just have to take care of that formality. Thanks!

PR #518, opened 2018-07-30
WIP Add segmented control test

tkrose:


lib/SegmentedControl/tests/SegmentedControl-test.js
```diff
@@ -1 +1,149 @@
...
+import { expect } from 'chai';
+import Button from '../../Button';
+
+import { mountWithContext } from '../../../tests/helpers';
+
+import SegmentedControl from '../SegmentedControl';
+import SegmentedControlInteractor from './interactor';
+
+
+describe.only('SegmentedControl', () => {
```
cherewaty: `.only()` is handy for isolating the test locally, but we don't want it in `master` - then no other tests would run!

cherewaty:
@tkrose - just a heads up: I merged a PR that deleted the previously blank version of `SegmentedControl-test.js`; when you rebase this, make sure your version of `SegmentedControl-test.js` gets re-created

taras:
@tkrose this is not a WIP anymore right? If so, then we can remove the WIP from the title.

JohnC-80:
I had to look up the significance of the "shipit" squirrel emoji. TIL there's doesn't have to be a reason for everything.

PR #519, opened 2018-07-30
STCOM263-MultiSelection New component

JohnC-80:
# MultiSelection Component
Pairing with Selection, we also have scenarios where a user may have to select multiple values from (potentially large) collections of system entities.
Standard `<select>[multiple]`'s are burdensome for development, accessibility, and user experience since they don't display any up-front feedback for selected values nor do they provide a comfortable keyboard-only experience. Further, extra steps may have to be taken in code to adequately capture the real 'value' of the input.

## Approach
Make use of [Downshift](https://github.com/paypal/downshift) - a great utility for creating components of this type that provide built-in ways to easily apply numerous a11y features such as `aria-attributes` and keyboard navigation. 

### Tether is optional
Potential pattern we can use with our other overlay components to push them to the more-accessible side of things.

Putting this up as a WIP to expose the code to the community and get some extra eyes on it for review.

WIP-GIF:
multiselectioncomponent

### test coverage
some shortcomings due to lack of testing for small screen widths.

image

## Documentation
[The readme](https://github.com/folio-org/stripes-components/blob/stcom-263-multiselection/lib/MultiSelection/readme.md)
## To-do
* [x] basic functionality
* [x] validation messages
* [x] a11y announcements of labels, actions, and values
* [x] add `downshift` dep
* [x] responsive mode
* [x] tether-opt-in
* [x] tests
* [x] i18n
* [x] docs!
* [x] storybook stories


JohnC-80:
@mkuklis with that specific feature in mind, it includes an 'actions' prop for options that will render to the list and give developers the ability to call a particular function via the action's `onSelect` property. You can see some detail about that here: https://github.com/folio-org/stripes-components/blob/stcom-263-multiselection/lib/MultiSelection/readme.md#actions

mkuklis:
@JohnC-80 thank you for pointing me to it!

PR #520, opened 2018-07-31
Add contentClass prop to Accordion

cherewaty:
## Purpose
In `ui-eholdings` we're trying to place a list using `impagination` inside a component (example: 
https://folio.frontside.io/eholdings/providers/19).

The CSS and JS we're currently leaning on to deal with scrolling places a `min-height` of `1` and `flex` `1` on the list. Placing it within an accordion makes that tricky.

## Approach
Introduce a `contentClass` prop to the `<Accordion>` the default accordion `content` CSS can be overridden.


JohnC-80:
This seems like one of those floodgate use-at-your-own-risk kind of props and I'm wondering if its best to apply it to the accordion's content element or perhaps an inner container since the accordion itself controls styling on the content div.

cherewaty:
You're right, this was a dangerous solution. Replaced with https://github.com/folio-org/stripes-components/pull/522

PR #530, opened 2018-08-08
Display delete icon on intial render of the field row, refs UIORG-98

aditya-matukumalli:


cherewaty:
I like this simplification 😁

PR #531, opened 2018-08-09
Update translation strings as of 20180809T173930

folio-translations:


folio-translations:
Removing old pull request.

PR #534, opened 2018-08-13
🔒Updated disabled styles

rasmuswoelk:
This PR adds missing disabled styles as well as fixing some of the existing ones to achieve uniform disabled styles across all form components.

On @filipjakobsen's request, I have added a lock icon which has been added to the label on all _disabled_ form fields. The added lock-icon and updated disabled styles can hopefully help emphasize that a field is disabled.

Check out the changes on the storybook draft site: http://ux.folio.org/storybook-draft/

## Here's how it looks:

image

## Overview of changes
- Added new icon: lock
- Updated the global disabled styles for input, select and textarea
- Added lock icon to all form components (TextArea, Select, TextField, Timepicker, Datepicker, Checkbox, RadioButton) in the label area. Visible if the disabled-prop is true
- Added disabled examples in storybook for several of the above components
- Added disabled state for RadioButton and Checkbox. Disabling interaction styles if disabled is true
- Removed buttons etc. from Timepicker and Datepicker if the fields are disabled. This is the same as when the fields are set to "read only"

Interesting read on screen readers and 'read only':
https://tink.uk/screen-reader-support-for-disabled-read-only-form-fields/


cherewaty:
I'm curious about including the lock icon - I don't think I've seen that as a standard before for disabled fields. Sometimes we disable fields and buttons while a form is submitting, but a lock wouldn't really make sense popping up on fields for just a second or two while the server request finishes.

Should the lock just be a recommended pattern to add to labels, instead of being applied to all disabled field?

JohnC-80:
True.. this might be a discrepency between 'read-only' and 'disabled'... @filipjakobsen @rasmuswoelk ?

filipjakobsen:
If that is the distinction in front end usage between locked and read-only, then I think it makes sense to apply the lock icon style to "read-only" fields rather than "locked" fields. We just need to make it clear how people ought to use the two, then.

rasmuswoelk:
Good point, @cherewaty. As @filipjakobsen mentions - can we add the lock to read-only fields?

If we go in this direction, I would like to know if the disabled styles should be added to read-only fields then?

filipjakobsen:
Yes, let's reuse all aspects of the styling for read-only 👌

rasmuswoelk:
Okay, so the <select> doesn't support "read-only". Also, the <Selection> component is a <button> which doesn't support read-only either.

In these cases, we could just make them "disabled" when the "readOnly" prop is applied and then show the lock in the label area? I'm thinking this would make sense in terms of consistency.

What do you guys think? @cherewaty @JohnC-80

JohnC-80:
Hmm... so disabling them would functionally put them out of the tab order while readOnly inputs remain there. `<select>`s are weird animals when it comes to this and it looks like devs resort to hacks for it... hidden inputs to capture state (previous values already in state) -or- setting 'disabled' attributes on all non-selected options. Not sure what's not evil here...

JohnC-80:
Rather than leaning solely on the pseudostates to apply these styles, why not also have `.disabled` and `.readonly ` class names that could be applied so that the styles could work regardless of what attributes the elements support?
I think we can deal with read-only functionality (applying the styles here is fine) of selects specifically in a future PR since this would require more tinkering within the components themselves to make it functional... further, the usage of 'disabled' vs 'readOnly' is definitely good material for a guide to forms in Stripes.

rasmuswoelk:
@JohnC-80 

I have updated the code and moved the lock-icon to the read-only state of all form elements. I have also added the .disabled and the .readOnly class names that apply the same styles as `:disabled` in form.css.

Since the Select, Checkbox And RadioButton doesn't have the read-only option natively I have only applied the disabled/read-only styles to these components. This, of course, means that the fields aren't really "read-only" currently - they only appear like this and they will render the lock-icon.

Should we do anything further on these components? I'm not entirely sure what the approach would be. E.g. we could disable onChange handlers etc. to make sure that they truly are "read-only". I'm not sure if this is a good idea though. 🤔

Alternatively, it could be (and probably should anyway) documented what the "read-only"-state is and how to use it. Developers could then manually make sure that the above-mentioned components are read-only in their own code.

Do you have an opinion on what to do here?

JohnC-80:
Hmm... interestingly enough, the readOnly select actually seems to be working as I imagine it ought to in the storybook... ( it's focusable, but I can't select anything) For checkboxes, I think we might have to resort to not rendering the actual control itself, but something that a) visually represents the value and b) is focusable, announcing its value and that it's read-only.

rasmuswoelk:
@JohnC-80 

In Storybook the read-only Select isn't set up correctly with a proper onChange handler which is why you can't select anything. 

So in a real form, you would be able to pick a value. Should we prevent this functionality (being able to choose a value) in the component itself _or_ should it be described as a pattern somewhere?

For the Checkboxes and Radio Buttons: Ok, sounds good. I'll look into what we can do.

JohnC-80:
Ok.. so for the 'disable-all-non-selected-options' path, it will need slightly more complicated logic in the render ( probably a method call for that 'disabled' attribute on each option) and additional logic to add an appropriately discernable label to relay the 'read-only' status....

@rasmuswoelk your comment makes me think of react documentation satire : "For a read-only select apply the permanent value in the 'value' prop, but DON'T apply an onChange handler. It's not Javascript,  It's the way React works!" Not a legitimate strategy, unfortunately, as the accessibility gap of informing a user that it's 'read-only' still has to be fulfilled.

lib/RadioButton/RadioButton.js
```diff
@@ -167,9 +184,23 @@ class RadioButton extends React.Component {
             ref={(ref) => { this.input = ref; }}
             autoFocus={autoFocus}
             value={value}
+            readOnly={readOnly}
           />
           <span className={css.labelPseudo} />
-          { label && <span className={css.labelText}>{label}</span>}
+          { (label || readOnly) &&
```
JohnC-80: I believe we can simplify this and lose the odd aria-conflict with `aria-labelledby` + logic reaching into an `aria-hidden` span...
JAWS and NVDA will both read whatever text content is present within the `<label>` tag, so the lock icon and its text (with the visually-hiding `sr-only` class) could be rendered there, something like 
```javascript
 <span className={css.labelText}>
   {label}
   {readOnly && (
      <Icon size="small" icon="lock">
         <span className="sr-only">
            <FormattedMessage id="stripes-components.readonly" />
         </span>
       </Icon>
    )}
  </span>
```

rasmuswoelk: I like this much better.

Checkbox and RadioButton is now updated.

PR #537, opened 2018-08-14
Added open/close transition to Modal

rasmuswoelk:
I added a very subtle open/close transition to the Modal-component.

Check out the transition here: http://ux.folio.org/storybook-draft/?selectedKind=Modal

aug-14-2018 15-48-29


rasmuswoelk:
Thanks, @cherewaty! I will be adding more subtle transitions to various components in the future!

PR #538, opened 2018-08-14
Upgrade to eslint-config-stripes 3

cherewaty:
I was able to run `yarn eslint --fix` for most of the changes.

Notable new rules I saw pop up:
- [react/destructuring-assignment](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md): I ignored this one in `.eslintrc` for now; it's worthwhile, but will take enough work that it should be its own separate PR.
- [react/no-access-state-in-setstate](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md)
- [lines-between-class-members](https://eslint.org/docs/rules/lines-between-class-members)
- [operator-linebreak](https://eslint.org/docs/rules/operator-linebreak#enforce-consistent-linebreak-style-for-operators-operator-linebreak)
- [react/jsx-tag-spacing](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-tag-spacing.md)

`react/no-deprecated` now fires on the relevant line of the deprecated hook, instead at the beginning of the class. That's a nice improvement.


lib/Accordion/headers/FilterAccordionHeader.js
```diff
@@ -78,12 +78,12 @@ class FilterAccordionHeader extends React.Component {
...
             <div className={css.labelArea}>
-              <div className={css.filterSetLabel} >{label}</div>
+              <div className={css.filterSetLabel}>{label}</div>
             </div>
           </button>
         </Headline>
-        { clearButtonVisible ?
-          <IconButton
+        { clearButtonVisible
+          ? <IconButton
```
JohnC-80: First, thanks @cherewaty these will clean things up significantly... the only one that feels weird to me is the operator-linebreak rule applied to ternary operators.... While the operator is grouped with its true outcome, it seems clearer when grouped with the conditional... but that's just, like, my opinion, man! @zburke @mkuklis how do you guys feel about this spacing?

JohnC-80: It appears that the rule can have overrides set for cases like this targeting certain operators. I'm not sure it feels right to have this be consistent across the board, but I'm curious as to other folks opinions on this. https://eslint.org/docs/rules/operator-linebreak#options

MikeTaylor:
To my mind, this is a classic example of the kind of judgement much better made by a human than by a computer: enforcing a rigid style only creates busy-work while making the code _less_ readable.

> "Design and programming are human activities; forget that and all is lost" -- Bjarne Stroustrup.

lib/Accordion/headers/FilterAccordionHeader.js
cherewaty: I don't have a strong opinion on that rule. I tend to use parentheses most of the time anyway to have more clear indentation:
```
{someBoolean && (
  <div>
    Hello
  </div>
)}
```

JohnC-80:
Thankfully we, the humans, do have the ability to assign local overrides to these rules to support what seems to work best. The goal of this isn't necessarily to let the machine dictate how you write, it's to let it guide the code you write for the benefit of others. I've grown to appreciate linting as it continuously does make you think about the legibility of your code and general comfort of those who might be working with it in the future... The things that it focuses on may seem quite nitpicky but it's within the little mechanics of writing code that do make some difference - and consistency of style is nice within any codebase.

lib/Accordion/headers/FilterAccordionHeader.js
JohnC-80: I'm leaning towards losing the rule. With additional nuances such as yours, I don't think it's worth enforcing in any shape.

cherewaty: New set of commits pushed with `operator-linebreak` turned off and `lines-between-class-members` set to ignore single-line class members.

PR #539, opened 2018-08-15
Minor <Settings> update

rasmuswoelk:
~~Added some line breaks to add a bit of spacing between NavListSections.~~

~~It's probably better to style these with CSS instad of using `<br />`'s. However, since the sections in the Settings-pane in stripes-core currently uses this method for spacing I decided to do the same here to match the spacing.~~

Added some margin to each NavListSection of the Settings-component.

## Before
image

## After
image


MikeTaylor:
Agreed -- let's not compound the earlier error by establishing it as The Way We Do This.

rasmuswoelk:
Good point 🙂I'll change it to CSS and submit a PR for stripes-core as well 👍🏻

PR #542, opened 2018-08-16
UIU-459: export csv

rjruizes:
https://issues.folio.org/browse/UIU-459

- `exportToCsv(csv)` is in its own file for transforming the JSON to CSV, optionally omitting columns, and triggering the download
- MultiColumnList was originally just the table. I added a sibling that precedes it, ActionbarContainer, which caused the whitespace diff seen on Github. This container has two children: the optional prop `actionbar` and the "Export to CSV" `<Button>`
- increment patch version so it can be consumed by `ui-users`

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -892,7 +892,7 @@ class MCLRenderer extends React.Component { // eslint-disable-line react/no-depr
             onClick={() => { exportToCsv(contentData, csvExcludeKeys); }}
             style={{ margin: '5px 15px 0 -10px' }}
           >
-            <FormattedMessage id='stripes-components.exportToCsv' />
+            <FormattedMessage id="stripes-components.exportToCsv" />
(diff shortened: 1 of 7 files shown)


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -875,63 +881,78 @@ class MCLRenderer extends React.Component { // eslint-disable-line react/no-depr
...
+          {this.props.actionBar}
+          {this.props.csvShowButton && (
+          <Button
+            id="export-csv"
+            onClick={() => { exportToCsv(contentData, csvExcludeKeys); }}
+            style={{ margin: '5px 15px 0 -10px' }}
           >
-            {renderedHeaders}
-          </div>
+            Export to CSV
```
cherewaty: Should we drop this string in the translations file?

rjruizes: Good catch, thank you

PR #550, opened 2018-08-22
Date range primitive

JohnC-80:
# New component: `<DateRangeWrapper>`

## Purpose
A primitive wrapper component for assembling date range structures with 'start' and 'end' fields. Layout and mark-up could vary for this combination of fields - so that aspect is kept variable with this approach.

## Approach
Uses render-props to supply `exclude` props for dates outside of the range.
Has props on the main component to override/augment existing functionality to handle edge use-cases.

[Check out the readme](https://github.com/folio-org/stripes-components/blob/date-range-primitive/lib/DateRangeWrapper/readme.md)

daterangewrapper

## Tests are there
image


lib/DateRangeWrapper/DateRangeWrapper.js
```diff
@@ -0,0 +1,98 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const callAll = (...funcs) => (...args) => funcs.forEach(fn => fn && fn(...args));
```
mkuklis: @JohnC-80 this one is pretty cool 👍

mkuklis: Everything looks great @JohnC-80

cherewaty:
This looks pretty good. I'm going to try it out in eHoldings before giving it a full thumbs up.

lib/DateRangeWrapper/DateRangeWrapper.js
JohnC-80: That's a trick from Kent Dodd's. Gotta love elegant one liners. :smiley_cat:

lib/DateRangeWrapper/readme.md
```diff
@@ -0,0 +1,96 @@
...
+// the only child is a function, mentioned later as the 'child function'
+<DateRangeWrapper>
+  {({
+    getStartInputProps,
+    getEndInputProps,
+    endDateExclude,
+    startDateExclude,
+  }) => (
+    <div>
+      <Datepicker
```
cherewaty: We probably want to deprecate the `input` prop of `Datepicker` when we start to use `reduxFormField()` with it.

Is this possible instead?
```
<Field
  component={Datepicker}
  label="Start date"
  exclude={startDateExclude}
  {...getStartInputProps}
/>
```

...then later we can update the documentation to use a bare `Datepicker` without `redux-form`?

JohnC-80: With Redux-form/Field, that's the way to do it. I'll include an additional 'with redux-form' example.

PR #554, opened 2018-08-24
Update pt_BR translation

trmurakami:


JohnC-80:
@trmurakami Thank you for the contributed translations! Looks like we need to get the Contributor License Agreement from you. You need to associate the email address used for you commit with your github account. Once that's done,  simply follow the link in the CLA comment and sign it with your github credentials - this is a one-time process. Thanks again!

PR #555, opened 2018-08-24
Perform case insensitive sorting when displaying locations, refs FOLI…

aditya-matukumalli:
…O-1416


MikeTaylor:
Well, hang on: this doesn't just change the sort-order to case-insensitive for location -- it changes it for _everything_ that uses `<EntrySelector>`. Are we sure we want that? More importantly, are we sure the POs and SMEs want it?

PR #557, opened 2018-08-27
Added an icon-prop to the NavListItem

rasmuswoelk:
Passing an icon as a strong to the NavListItem will render an `<Icon>` next to the label.

image

image

### DIFF
--- a/lib/NavList/readme.md
+++ b/lib/NavList/readme.md
@@ -5,7 +5,7 @@ Renders a vertical navigation list in conjunction with NavListSection
 ## Basic Usage
 NavList uses `<NavListSection>` to add one or more sections to your navigation.
 
-You can use the `<NavListItem>` component to add links/buttons to your NavListSection. By using the "to"-prop it will render a react-router `<Link>`-component, using the "href"-prop will render a regular <a> and the "onClick"-prop will render a button. The component will accept any valid props that these HTML elements accepts.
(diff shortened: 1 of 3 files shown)


cherewaty:
I have doubts that we want to design icons for every possible thing that could appear in that menu, but this implementation looks good 👍

JohnC-80:
I have similar concerns, @cherewaty .
@rasmuswoelk Would `<NavListItems>` without icons in the same list still be completely left-aligned?

rasmuswoelk:
@JohnC-80 @cherewaty 

After submitting it I did think about whether or not this was the correct approach. You can achieve the same thing by passing an `<Icon>` to the an list item.

The reasoning behind having it as prop was to make it easier for devs to add an icon since I expect that it will be a fairly repeated design pattern across the system (judging from prototypes).

It's of course completely optional and any ListItems without icons are untouched by this change.

JohnC-80:
I think a design situation that's to-be-considered is this, when not ALL of the menu items have an icon associated with them: there's both an alignment issue and a design issue: 
Example: an icon missing, menu-item goes full-left-align, out of alignment with the rest.
image

A possible way, but would add default padding or some sort of spacer to all items if they didn't have an icon (possibly not ideal either)
image

@filipjakobsen @rasmuswoelk

rasmuswoelk:
@JohnC-80 Hmm. I guess if you don't have icons for all items in a list, you would not use icons in any of the items. And vice versa.

But since this can be achieved by simply passing an `<Icon>` to a `<NavListItem>` currently I will close this issue.

PR #561, opened 2018-08-28
STCOM-88 Update <RadioButton> and <RadioButtonGroup> to work independently of Redux Form

cherewaty:
## Purpose
Resolves [STCOM-88](https://issues.folio.org/browse/STCOM-88), [STCOM-322](https://issues.folio.org/browse/STCOM-322), and [STCOM-323](https://issues.folio.org/browse/STCOM-323).

## Approach
- Use `reduxFormField()` for consistent logic with or without Redux Form.
- Made the props identical between `<Checkbox>` and `<RadioButton>`.
- Checked against the eHoldings acceptance tests. Found and fixed several regressions that way.
- Updated stories and readme.

## Versioning
Contains significant prop changes to `<RadioButton>` and `<RadioButtonGroup>`. I bumped to the next major version in `package.json` and the changelog. Consumers of those two components with Redux Form shouldn't expect any breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change history for stripes-components
 
-## 3.1.0 (IN PROGRESS)
+## 4.1.0 (IN PROGRESS)
+
+## [4.0.0](https://github.com/folio-org/stripes-components/tree/v4.0.0) (2018-08-28)
+[Full Changelog](https://github.com/folio-org/stripes-components/compare/v3.0.0...v4.0.0)
 
 * Expose `Settings` nav pane width via `navPaneWidth` prop.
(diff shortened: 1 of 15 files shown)


JohnC-80:
@cherewaty At the moment, a major version bump in stripes components means that modules across the platform have to stay in-line with the major version of stripes-core's stripes-components dependency... The React Context issue ([STCOM-335](https://issues.folio.org/browse/STCOM-335)) shows up if there are multiple major versions present within the platform. A possible idea - announce this change now via console warning in `<RB>` and `<RBG>` and set up a v4 branch of breaking changes (this, and I'm sure there will be more once some of the stories have been completed) We might have a number of these breaking changes within close proximity - could we possibly strategically announce and group them?

cherewaty:
Closing, will open against https://github.com/folio-org/stripes-components/pull/566 instead.

PR #563, opened 2018-08-29
Deprecate passThroughValue

cherewaty:
Removing `passThroughValue` will toss out a lot of complexity from `<Datepicker>` and `<Timepicker>`.  `ui-checkin` is the only `folio-org` repo using `passThroughValue`, and I'm closing in on an alternative design solution. Probably a good time to start throwing deprecation warnings.


JohnC-80:
We worked on deprecating it before. Things to consider:
It is a submit-time issue, and *not* an issue of this component setting a value via a named string.
If it's left within the component, and a form stays open for an extended amount of time (say, overnight) the value that's set at the form's mount time will be what gets used - so 'today' will no longer be 'today'...
Although check-in is the one instance of this, I'm surprised if it's the only one that would ever come up.
There will have to be design discussion going back to the way that check-in works for this.
I don't think we should put out deprecation warnings until the alternative has a more solid determination - in this case, changes in the design of check-in could be executed and that use-case can be discussed before we go through deprecating and breaking things...

cherewaty:
Alternative solution to `passThroughValue`: https://github.com/folio-org/ui-checkin/pull/58

PR #564, opened 2018-08-30
fix icon plus size and add document  icon

cchiama:
fix icon plus size and add document  icon


rasmuswoelk:
I think we'll rename any icons without whitespace to have a suffix of `inline` which indicates that it should be used inline with text. E.g. `icon-plus-inline` (without whitespace) and `icon-plus` (with whitespace).

We would need to update documentation and replace icons in various places while doing so to avoid "breaking" anything. 

So for now, could we just add the new document icon and leave the re-sizing of the plus-icon out of this PR?

rasmuswoelk:
@cchiama

cchiama:
Hi @rasmuswoelk

rasmuswoelk:
@cchiama Looks good!

I was wondering if "document" would be a more fitting icon name than "report"?

be-engineering:
> @cchiama Looks good!
> 
> I was wondering if "document" would be a more fitting icon name than "report"?

I agree with you @rasmuswoelk

cchiama:
I agree with you @rasmuswoelk

cherewaty:
With https://github.com/folio-org/stripes-components/pull/674 merged...
- we can remove `down-caret-inline` and `up-caret-inline`
- resize `document` to 32x32

... and this'll be shippable 🚀

cchiama:
Hi @rasmuswoelk i made a unique change, on document icon and i cannot made change on the icon plus, as request in above comments. The PR can be merge if you want.

cherewaty:
Sorry to keep moving the target @cchiama - we've made a lot of changes here recently. All the other icons now have a `viewBox` of `"0 0 32 32"`.

cchiama:
fix EOL lint error @rasmuswoelk.

rasmuswoelk:
@cchiama 

I think the last thing that needs to be updated is the viewBox. As @cherewaty mentioned above the viewBox should be `viewBox="0 0 32 32"`

It should be mergeable after this small change 👍

cherewaty:
Looks like you resized the `viewBox` but didn't scale the shape itself, so it's a lot smaller than other icons: 
screen shot 2018-12-03 at 1 14 27 pm

cherewaty:
Resolved with https://github.com/folio-org/stripes-components/pull/760

PR #565, opened 2018-08-30
fix external-link icon size and unify svg viewBox with the other icons

cchiama:
fix external-link size and unify svg viewBox


rasmuswoelk:
This icon is meant to be used inline with text - which is why it's a bit different from the rest of the icons. We might need to have two versions of this icon - one with whitespace and one without.

What are your thoughts about this, @filipjakobsen?

rasmuswoelk:
Hi @cchiama 

You can just leave the change that you have made to the `external-link` icon here.

But can you add the old version to the icons file under the name `external-link-inline`? 

The inline icon can be used inline with text and the other icon can be used for buttons etc.

cchiama:
I

cchiama:
This PR can be closed, because is the same of #564 , related icon size. And as @cherewaty said, the best option is not created inline icon, so i made a simple change in a previus PR. Please merge or close this.

PR #578, opened 2018-09-04
Check in yarn.lock

cherewaty:
Checks in `yarn.lock`, so we don't have surprise dependency changes preventing us from running tests.

Adds `yarn.lock` to `npmignore`, so it won't get published with the package.


MikeTaylor:
Hang on, though ...

@skomorokh @zburke @ttamj @JohnC-80 I seem to remember us having a long discussion a while ago about whether or not we wanted to check our `yarn.lock` files into git, and concluding -- at least at that time -- that we did not. Annoyingly I can't remember any of the details. Can any of you?

PR #583, opened 2018-09-06
MultiSelection bugfix

JohnC-80:
Fixed issues with updating values. Values are now stored in external state. We can add an "uncontrolled" wrapper if the case warrants.

Code for actions is now in-line with documentation. (`filterText` vs `filterValue`);
API remains the same (and now performs closer to expectations)


mkuklis:
@JohnC-80 I just gave it a try and everything works as expected. Thank you again for fixing this!

PR #586, opened 2018-09-06
STCOM-88 Update <RadioButton> and <RadioButtonGroup> to work independently of Redux Form

cherewaty:
## Purpose
Resolves [STCOM-88](https://issues.folio.org/browse/STCOM-88), [STCOM-322](https://issues.folio.org/browse/STCOM-322), and [STCOM-323](https://issues.folio.org/browse/STCOM-323).

Replaces https://github.com/folio-org/stripes-components/pull/567.

## Approach
- Use `reduxFormField()` for consistent logic with or without Redux Form.
- Made the props identical between `<Checkbox>` and `<RadioButton>`.
- Checked against the eHoldings acceptance tests. Found and fixed several regressions that way.
- Updated stories and readme.

## Versioning
Contains significant prop changes to `<RadioButton>` and `<RadioButtonGroup>`. I bumped to the next major version in `package.json` and the changelog. Consumers of those two components with Redux Form shouldn't expect any breaking changes.

## TODO
- [x] Add more documentation to readme

### DIFF
--- a/lib/RadioButton/RadioButton.js
+++ b/lib/RadioButton/RadioButton.js
@@ -148,7 +148,6 @@ class RadioButton extends React.Component {
     if (!name) {
       ariaProps['aria-label'] = `${label}`;
     }
-    const { ...inputAriaProps } = ariaProps;
 
     return (
       <div className={this.getRootStyle()}>
@@ -156,7 +155,7 @@ class RadioButton extends React.Component {
           <input
             id={id}
             type="radio"
-            {...inputAriaProps}
+            {...ariaProps}
             checked={checked}
             {...inputCustom}
             name={name}
--- a/lib/RadioButton/RadioButton.js
+++ b/lib/RadioButton/RadioButton.js
@@ -1,41 +1,119 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+
+import reduxFormField from '../ReduxFormField';
 import separateComponentProps from '../../util/separateComponentProps';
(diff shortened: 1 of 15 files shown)


lib/RadioButton/RadioButton.js
```diff
@@ -1,42 +1,120 @@
...
+    }
+
+    return classes;
+  }
+
+  // Get feedback element (if there's any feedback to show)
+  getFeedbackElement() {
+    const hasFeedback = !!this.getFeedbackClasses().length;
+    if (!hasFeedback) {
+      return false;
```
wwilsman: Since this is meant to maybe return a node, we should return `null` here so any type checking is still accurate.

lib/RadioButton/RadioButton.js
```diff
@@ -47,59 +125,68 @@ class RadioButton extends React.Component {
...
-    const [, inputProps] = separateComponentProps(this.props, RadioButton.propTypes);
+    const [, inputCustom] = separateComponentProps(this.props, RadioButton.propTypes);
 
-    if (input) {
-      Object.assign(inputProps, input);
+    const ariaProps = {};
+    if (!name) {
+      ariaProps['aria-label'] = `${label}`;
     }
+    const { ...inputAriaProps } = ariaProps;
```
wwilsman: This is effectively a shallow copy of an object we just created, any reason for this?

cherewaty: Probably just vestigial. Removed.

PR #588, opened 2018-09-10
Add initial value

etalmacsi:
Set the initial value of the textfield with a props

### DIFF
--- a/lib/Datepicker/Datepicker.js
+++ b/lib/Datepicker/Datepicker.js
@@ -121,7 +121,14 @@ class Datepicker extends React.Component {
     }
 
     // non-null presentedValue will eventually be rendered as the value of the TextField.
-    let inputValue = '';
+    let inputValue;
+
+    if(this.props.initialValues){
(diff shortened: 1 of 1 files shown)


cherewaty:
@etalmacsi we should probably make some documentation additions to make this more clear, but are you using [Redux Form](https://redux-form.com/) with the `Datepicker`?

If yes, you should be able to set `initialValues` on the form, not on the individual field:
```
import React, { Component } from 'react';
import { reduxForm, Field } from 'redux-form';
import { Datepicker } from '@folio/stripes-components';

class FormWithCalendar extends Component {
  static propTypes = {
    onSubmit: PropTypes.func.isRequired
  }

  render() {
    return (
      <form onSubmit={onSubmit}>
        <Field
          name="myDateField"
          component={Datepicker}
        />
      </form>
    );
  }
}

export default reduxForm({
  form: 'myFormWithADatepicker',
  initialValues: { myDateField: '2018-09-10' }
})(FormWithCalendar)
```

PR #589, opened 2018-09-10
Updates to <Layout> component

rasmuswoelk:
The `<Layout>` component is a very useful utility component that allows developers to apply various pre-defined styles to parts of an UI.

Perfect for small adjustments here and there as well as re-using existing CSS styles inside current CSS files by utilizing PostCSS's composition feature — all of which can save the final CSS bundle from becoming bloated in the future.

**Changes:**
- Added readme
- Added tests for a 100% coverage
- Made sure that you can pass along custom class names instead of only mapping to predefined class names
- Split style up into smaller files for a better overview
- Added example to storybook that allows for filtering through class names

For a better understanding of what each class names do I added a file called "classNamePropertiesMapping.js" that maps each class name to the specific styles that it applies. This, of course, means that you have to update this file when adding or updating new class names. The neat thing about it is that you can search both CSS attributes and class names to find the specific class name you need.

You can check it out here: https://ux.folio.org/storybook-draft/?selectedKind=Layout

I tried finding a more automatic way of doing this but I never found a solid solution. If anyone knows of an elegant way of grabbing CSS attributes by className from a CSS file and exporting them to JS/JSON (or some other way to import them into my JS) then please let me know!

**Note:** I did not remove any of the existing class names to avoid stuff breaking. Some of the existing ones might need to be removed later on. I did add a few more helper classes and I will be adding more later.


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "editor.tabSize": 2,
+    "editor.detectIndentation": false,
+}
```
JohnC-80: We should add an entry for `.vscode/` to the `.gitignore` Otherwise I really like this.

rasmuswoelk: Done 👍

PR #592, opened 2018-09-12
Removed forced text capitalization from MCH column headers

rasmuswoelk:
Removed forced text capitalization from MCH column headers so that sentence casing can be controlled elsewhere - e.g. in translation files etc. The goal is to achieve consistent sentence casing across the FOLIO system.

In some apps (e.g. Inventory - example below) the header labels are currently all lowercase. This will need to be changed of course - maybe while doing so, we could change them into using translated strings instead?

image

Issue: https://issues.folio.org/browse/STCOM-337


JohnC-80:
I think we should include additional instruction in the `<MCL>` documentation on using the `columnMapping` prop to achieve this and note the design consistency point of sentence case for column headings. It would give PO's a reference point and possibly a link that they can pass along to developers, and create some extra visibility for this aspect of things to developers themselves.

rasmuswoelk:
@JohnC-80 Got it. I'll update the readme 👍

PR #595, opened 2018-09-12
Improve component deprecation messages

cherewaty:
They were a bit too brief before. These should provide more clarity.

### DIFF
--- a/lib/AddressFieldGroup/AddressEdit/AddressEditList.js
+++ b/lib/AddressFieldGroup/AddressEdit/AddressEditList.js
@@ -26,7 +26,11 @@ class AddressEditList extends React.Component {
   constructor(props) {
     super(props);
 
-    console.warn('Warning: Update path to stripes-smart-components/lib/AddressFieldGroup/AddressEdit/AddressEditList');
+    console.warn(
+      '\nWarning: <AddressEditList> is deprecated and will be removed in the\n' +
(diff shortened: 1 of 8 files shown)


lib/AddressFieldGroup/AddressEdit/AddressEditList.js
```diff
@@ -26,7 +26,11 @@ class AddressEditList extends React.Component {
   constructor(props) {
     super(props);
 
-    console.warn('Warning: Update path to stripes-smart-components/lib/AddressFieldGroup/AddressEdit/AddressEditList');
+    console.warn(
+      '\nWarning: <AddressEditList> is deprecated and will be removed in the\n' +
+        'next major version of @folio/stripes-components.\n\n' +
+        '<AddressEditList> is now available from @folio/stripes-smart-components.\n'
+    );
```
JohnC-80: I think we could knock these down to a Deprecation Warning 'label', the `<ComponentName>`, and the brief migration instructions - or - a link to migration instructions living in a markdown file. Ex: 
```
Deprecation Warning: <AddressEditList> - Update import path to "stripes-smart-components/lib/AddressFieldGroup/AddressEdit/AddressEditList" - see http://github.com/folio-org/stripes-components/MIGRATIONS.md#v3-to-v4 for more details.
```
We're going to want to keep those instructions around for migrations that take place outside of the window of these console warnings anyhow.

PR #598, opened 2018-09-13
STCOM-286 - Fix overlapping TextField icons

wwilsman:
## Purpose

[STCOM-286](https://issues.folio.org/browse/STCOM-286) - `TextField` can occasionally get in a state where it does not have adequate padding for the search icon at the beginning of the field.

While unable to reproduce this case specifically, I was able to reproduce a related issue in storybook:

1. Open storybook
2. Navigate to the `TextField` basic usage and scroll down to the start/end control example
3. Refresh the page and scroll down to the example again
4. Observe that the final example is now double the width of the viewport due to excessive padding.

#### Explanation:

Storybook injects CSS for a module on demand. When this page loads for the first time, there are no styles associated with icon sizes, so the icons default to 100% width.

`TextField` is mounted and calculates the ridiculous padding.

CSS is loaded, icons are resized, but the `TextField` padding remains.

I suspect this is similar to what is happening when the icon overlaps the text, except when the padding is calculated, the icon has no size at all. Resulting in the overlap once styles are applied.

## Approach

Remove the padding calculations and use flexbox instead.

Flexbox allows us to layout the control components next to the input field while keeping everything within the proper width and expanding the input to be as wide as it can be accounting for any controls. All without having to do any manual calculations, re-rendering the component, or interacting with the DOM at all.

In order to achieve this, the styles needed to be moved to the input group that contains the controls and input elements. The input is then made transparent so the group styles are visible from behind. 

This has the caveat of disabled and focus styles not being able to encompass the entire input group. To counteract this, `isDisabled` and `isFocused` classes were added to the shared form styles and applied to the input group.

The previous blur handler only applied when clicking clear. It has been moved to the input group, and logic was added to the blur handler to ensure it is only blurred when focus is not within the group itself, and that focus is only called when not already focused. The clear handler was also adjusted to utilize `setState`'s callback instead of relying on a timeout of `5`.

## Screenshots

**Before:**

before

**After:**

after

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,6 +65,8 @@ jobs:
       - run:
           name: Build storybook bundle
           command: yarn storybook-build
+      - store_artifacts:
+          path: ./.out
 
   test-chrome:
     docker:
@@ -76,18 +78,15 @@ jobs:
           name: Run tests
           command:
             yarn test
+            --coverage
             --karma.singleRun
             --karma.browsers=Chrome
-            --karma.reporters mocha junit BrowserStack
+            --karma.reporters mocha junit
             --karma.junitReporter.outputDir artifacts/junit/karma
       - store_test_results:
           path: ./artifacts/junit
       - store_artifacts:
-          name: Store junit test report
-          path: ./artifacts/junit
-      - store_artifacts:
-          name: Store code coverage report
-          path: ./artifacts/coverage/lcov.txt
+          path: ./artifacts
 
   test-firefox:
     docker:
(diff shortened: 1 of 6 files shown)


rasmuswoelk:
I'm a big fan of removing the "calculate-and-apply-padding-with-javascript" functionality. Using flexbox is much more smooth and less prone to cause weird visual bugs.

However, we still need to apply the focus to the outer element and not the input itself. Using `:focus-within` is not an option (yet) in my opinion because of the incompatibilities in IE/Edge.

I guess the only way to make it work would be to add a CSS class using JS for now. It's still less "hacky" than the previous solution and we get the desired visual result in all browsers.

In `sharedStyles/form.css` we have the shared styling for inputs. We could add `.is-focused` alongside `:focused` and then apply it with JS on focus.

wwilsman:
@rasmuswoelk That was my thoughts exactly. I didn't do it at first because I wanted to hear your thoughts on this pure CSS solution.

wwilsman:
@cherewaty Can I get another review? I realized that my blur handler change broke the clear button, but there were no clear button tests that caught it

cherewaty:
👍 I love anytime we can cut a `setTimeout`

PR #599, opened 2018-09-17
Added src, title & alt props to AppIcon + retina detection

rasmuswoelk:
In order to fix this issue: https://issues.folio.org/browse/STCOR-253, we need to be able to overwrite the alt and title attributes on the AppIcon. In this update, I changed the AppIcon so that it's possible to pass alt, src, and title as props without breaking the existing functionality.

I also updated the if statement for icons that are retrieved from the stripes context, that has a size of "small" to check if the user's screen is a non-retina screen. We only want to render the 14px PNG version of the icon for users on non-retina screens because the SVG version can get a bit blurry on non-retina screens.

Since the screen resolution (afaik) can't be tested consistently I decided to ignore this specific block of code from tests so that the test coverage is still 100%.

### DIFF
--- a/lib/AppIcon/AppIcon.js
+++ b/lib/AppIcon/AppIcon.js
@@ -15,7 +15,7 @@ import css from './AppIcon.css';
 const AppIcon = ({ size, icon, alt, title, src, style, children, className, tag, app, iconKey, stripes }) => {
   const getIcon = () => {
     let appIconProps;
-    
+
     /**
      * Icon from context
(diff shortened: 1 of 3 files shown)


rasmuswoelk:
@JohnC-80 I'll merge for now but I'll gladly extract the retina detection into file afterward. I'm assuming this would go in the `utils`-folder?

I'm curious - do you have any examples of components that might need retina detection?

PR #600, opened 2018-09-17
Allow 'iconAriaHidden' on <AppIcon> and updated PaneHeader

rasmuswoelk:
Today I noticed that the icon in the PaneHeader was a bit misaligned and after some brief investigation I found that it had been wrapped in an additional `<span>` with `aria-hidden="true"`.

To avoid having the extra span I decided to add an "ariaHidden" prop to the AppIcon but then I realized that the AppIcon can have a label so it does not make sense to make the whole component aria-hidden.

Instead, I added the prop "iconAriaHidden" which only applies to the icon itself - leaving the label (if available) to still be readable by screen readers.

### DIFF
--- a/lib/AppIcon/AppIcon.js
+++ b/lib/AppIcon/AppIcon.js
@@ -12,7 +12,7 @@ import classNames from 'classnames';
 import { withStripes } from '@folio/stripes-core/src/StripesContext';
 import css from './AppIcon.css';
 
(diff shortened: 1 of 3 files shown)


JohnC-80:
I think it would be a good idea also for `ariaHidden = true` to be the default setting for this - these icons are usually within a button or link that's already focusable or has its own discernable label (possibly the icon's child.)

PR #602, opened 2018-09-18
Update stripes-form dependency to 1.0

mkuklis:


mkuklis:
Yes waiting for this one https://github.com/folio-org/stripes-core/pull/423

cherewaty:
`stripes-core@2.13.0` is published

PR #608, opened 2018-09-19
Decouple form element styles

cherewaty:
Should help to alleviate the specificity problems we've been experiencing with the `form.css` styles.

### DIFF
--- a/lib/sharedStyles/form.css
+++ b/lib/sharedStyles/form.css
@@ -4,11 +4,6 @@
 
 @import "../variables";
 
-:root {
-  --input-horizontal-padding: 4px;
-  --input-vertical-padding: 8px;
-}
-
 /**
  * Input Group
  */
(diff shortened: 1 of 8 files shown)


lib/sharedStyles/form.css
```diff
@@ -99,33 +94,6 @@
   }
 }
 
-/**
- * Specific <input> style
- */
-input.input,
-.input input {
```
cherewaty: I'm really confused as to what this selector was doing.

wwilsman: It was initially just `input.input`, I'm assuming for forced specificity reasons. I added the `.input input` when I moved the `.input` class to be on the wrapper.

cherewaty:
I haven't tested with the `stripes-core` login form yet. I'm going to take another pass at this before merging - likely with some class name changes too (`input` -> `formControl`?)

rasmuswoelk:
@cherewaty It does not fix the login screen but I just submitted a PR that will.

And I agree; we should change the class to `formControl` since the class is not only being applied to `<input>`'s.

cherewaty:
Added commit that renames `input` to `formControl`.

PR #611, opened 2018-09-20
STCOM-348 Change global base font size from 14px to 16px

cherewaty:
https://issues.folio.org/browse/STCOM-348

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Update `stripes-form` dependency to v1.0.0
 * Fix paneset CSS behavior on narrow screens
+* Change global base font size from 14px to 16px
 
 ## [3.1.0](https://github.com/folio-org/stripes-components/tree/v3.1.0) (2018-09-13)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v3.0.0...v3.1.0)
--- a/lib/global.css
+++ b/lib/global.css
@@ -34,7 +34,7 @@ html {
     メイリオ    Meiryo,
     "ＭＳ Ｐゴシック",
     sans-serif;
-  font-size: 14px;
+  font-size: 16px;
   color: var(--color-text);
   letter-spacing: 0.04em;
   -webkit-font-smoothing: antialiased;
--- a/util/isStatefulComponent.js
+++ b/util/isStatefulComponent.js
@@ -1,6 +1,8 @@
-const isStatefulComponent = component => typeof component.type !== 'string' && component.type.prototype && component.type.prototype.render;
+const isStatefulComponent =
+  component => typeof component.type !== 'string' && component.type.prototype && component.type.prototype.render;
 
-const isStatefulComponentType = componentType => typeof componentType !== 'string' && componentType.prototype && componentType.prototype.render;
+const isStatefulComponentType =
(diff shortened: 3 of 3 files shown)


JohnC-80:
Seems like we just turned this down to `14px`... but if we want to support the range of user-agent font-sizes and stay above 9px, that might be necessary. @filipjakobsen should chip in also.

filipjakobsen:
Jeffrey’s point on touch devices and zoom is important. But we have other styles that also need to kick in on touch (all touchable elements should change their minimum height and width to 44px from 24px). This, together with the fact that A) having different font sizes in play for basic elements will make the interface look more disorganized; and B) The font size change ought to be accompanied by some other global styling updates; means I would suggest waiting with this change.

cherewaty:
There's no reliable way to detect if the user is on a touch screen (and what about a device where the user is flipping back and forth between mouch and touch?). Because of that, no component library I've ever seen attempts to resize its inputs. The form controls stay a consistent size no matter what the screen width or device capabilities are.

How have they dealt with the touch target issue instead?

### Form controls

#### Material Design
https://material-components.github.io/material-components-web-catalog/#/component/text-field
Label is cleverly integrated into the Text Field component, so touch target height is 56px.

#### Bootstrap
http://getbootstrap.com/docs/4.1/components/forms/
Input is 38px tall, but clicking on a corresponding label using `for=""` focuses the input without any JS magic. That means the touch target is effectively 63px tall.

#### Foundation
https://foundation.zurb.com/sites/docs/forms.html
Same thing with labels.

### Buttons
They've all made compromises with buttons. That 44px target height can be enormous in some layouts.

#### Material Design
https://material-components.github.io/material-components-web-catalog/#/component/button
36px tall

#### Bootstrap
38px tall, larger sizes available for some layouts

#### Foundation
https://foundation.zurb.com/sites/docs/button.html
40px tall, larger sizes available for some layouts

A change with this many ripples should happen as soon as possible, so we can address side effects. It will become more and more difficult to fix as development continues.

cherewaty:
Related label/focus issue: https://issues.folio.org/browse/STCOM-349

filipjakobsen:
Thanks for your comment, @cherewaty.

The major difference between those design systems, and the FOLIO design system, is that they are general-purpose design systems, whereas FOLIO is a highly specialized system for high-density, primarily large screen, heavy-use, productivity interfaces for libraries. As such, I would find it unfortunate to have to make e.g. checkbox lists, radio button lists, results lists, simple lists, navigational lists, app lists, content lists and other elements significantly bigger in order to at all times support touch, when the reality (currently) is that most people will be navigating this with a keyboard and a mouse. We should support responsiveness, browser zoom, touch, etc., but in order to use the screen real estate in the best possible way on laptops and on desktop screens, I feel like we need to make a distinction between the two scenarios, and that the design should reflect that. We have discussed this several times before, and there are things that could be done to support this in practice, without having to rely on auto-detection of touch.

I think that including the label in the touch zone for input fields, etc. is a good thing, and I think we in almost all places would be able to include an invisible hotspot around e.g. buttons, to make their effective touch target touch friendly (just like we do it for app icons in the Universal Header; and for Pane Header Icons in the Pane Header).

I am mainly concerned with enlarging list items of all kinds, as well as with using inconsistent font sizing across elements, in the system.

PR #616, opened 2018-09-24
STCOM-274 Fix paneset CSS behavior on narrow screens

cherewaty:
Paneset no longer simply disappears on smaller screens. Still lots of related bugs to work out before https://issues.folio.org/browse/STCOM-274 is marked as done.

"After" screenshot also indicates poor grid usage.

### Before
localhost_3000_users_view_d7af6445-28c6-58de-bb33-f9d5307825f4_query e sort name iphone 5_se

### After
localhost_3000_settings_users_perms iphone 5_se

### DIFF
--- a/lib/Paneset/Paneset.css
+++ b/lib/Paneset/Paneset.css
@@ -4,20 +4,25 @@
   width: 100%;
   height: 100%;
   position: absolute;
-  display: flex;
-  flex-direction: row;
-  align-items: stretch;
-  justify-content: flex-start;
+  display: block;
   will-change: transform;
   overflow: hidden;
 
   & .paneset,
   &.nested {
     position: relative;
-    flex-grow: 2;
   }
 
   &.static {
     position: static;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


lib/Paneset/Paneset.css
```diff
@@ -21,10 +21,3 @@
     position: static;
   }
 }
-
```
JohnC-80: still seeing some odd behavior in nested panesets (settings > users > permission sets....)
image

I ended up moving `display: flex` and all of the flex-related properties back into a --mediumUp query on Paneset.
```
@media (--mediumUp) {
  .paneset {
    display: flex;
    flex-direction: row;
    align-items: stretch;
    justify-content: flex-start;
  }
}
```

At small screen widths, there isn't any need for these to use flexbox since it's really one-pane-at-a-time situation.

cherewaty: Good catch. Commit added to address that. After:

PR #606, opened 2018-09-19
Add more headline sizes

cherewaty:
## Purpose
https://issues.folio.org/browse/STCOM-342
I want to be able to use a `legend` tag with the `<Headline>` component to get consistent styles.

## Approach
Removed global heading and legend styles, then set the new `<Headline>` sizes to browser default proportions:
https://www.w3.org/TR/html5/rendering.html#sections-and-headings

The existing sizes remain unchanged.

### Before
before

### After
after

PR #621, opened 2018-09-27
Adjust headline sizes

cherewaty:
## Purpose
Replaces https://github.com/folio-org/stripes-components/pull/606

## Approach
- Removed global heading styles, then set the new `<Headline>` sizes to browser default proportions:
https://www.w3.org/TR/html5/rendering.html#sections-and-headings
- Added `base` option that creates

### Before
screen shot 2018-09-27 at 2 58 34 pm

### After
screen shot 2018-09-27 at 2 56 19 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Fix paneset CSS behavior on narrow screens
 * Change global base font size from 14px to 16px
 * Expose legend as tag for `<Headline>`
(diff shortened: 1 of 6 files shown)


cherewaty:
Carrying forward the approvals I got on the initially larger changes I introduced in https://github.com/folio-org/stripes-components/pull/606 😁

PR #622, opened 2018-09-27
STCOM-341 Layer focus management

JohnC-80:
# Purpose
Layers, a11y-wise, should adopt the behavior of dialogs -
* focusing their element
* announcing a label
* trapping focus.

## Approach
Focuses the layer's element on open, unless focus is already within it via autoFocus or other means... will warn for an aria-label on the layer if it's missing.
Focus trapping is accomplished through a utility function. With default parameters, this function makes exceptions for the `#OverlayContainer` div and for `tether` elements. This is very similar to the 'enforceFocus' functionality of `react-overlay`'s modal, but that has proven to have some unforgiving, unpredictable timing (see https://issues.folio.org/browse/UICHKOUT-442 for one such story)
Here, focus management relies on an `afterClose()` callback to let the application control where focus should go once the Layer is dismissed.


JohnC-80:
My thoughts are the same about a refactor, but propping up the a11y for now...

PR #631, opened 2018-10-01
STCOM-355 Add padding to accordion content

cherewaty:
Resolves https://issues.folio.org/browse/STCOM-355

screen shot 2018-10-01 at 12 44 24 pm

Filter groups were already good, but needed some CSS changes to override the new defaults:

screen shot 2018-10-01 at 12 44 33 pm


cherewaty:
Commit added to use `margin` instead of `padding`.

PR #633, opened 2018-10-02
UIMPROF-25: Handle notification issue.

VictorSoroka1:


VictorSoroka1:
Callout component does not appear if "Change password" page visited straight away by URL `/settings/myprofile/password`. According to a conversation with Mark Deutsch, this could be caused by the rerender of the `OverlayContainer`. @doytch, @JohnC-80 could you please take a look?

doytch:
Victor and I were looking at this earlier today and it's weird.

1. Load a page using a callout like `/settings/users/groups`. I put in a breakpoint at the start of `Callout`'s `render()` and refreshed.
2. When it breaks there the first time, `this.calloutContainer` is undefined so we go and `getElementById` and fetch it. Before continuing from the debugger, I went ahead and threw a new attribute onto the overlayContainer element via the Elements pane in the debugger.
3. Continue from the breakpoint and when we hit the breakpoint again, I go into the Elements pane and note that the attribute I put on the overlayContainer is gone.

So I think that we're re-rendering the overlayContainer completely causing the Callout to be holding on to an outdated DOM element to render the Portal against.

Victor said this was fairly new behaviour so I went digging and the only thing that I saw that piqued my interest [was this change](https://github.com/folio-org/stripes-core/commit/bedc80d28b48854d65e538934e70f627da56fd4b) by @cherewaty that I figured could cause a whole rerender. I took out that `key` setting locally and even without this PR, everything worked properly again.

doytch:
So while this PR correctly addresses the regression introduced but I'm hesitant to just merge and go on with our lives. I think we should address the regression as well, since I'm worried that not only the `OverlayContainer` is affected by this re-rendering due to the key.

PR #634, opened 2018-10-02
STCOM-356 Remove child.type checks

cherewaty:
`react-hot-loader` prevents accurate `type` evaluations of children: https://issues.folio.org/browse/STCOM-356

By disabling these checks, these components are easier to misuse, but if they're used as intended and documented, everything should be fine.


cherewaty:
Reopened against release branch to get in today's `v4.0.0` release: https://github.com/folio-org/stripes-components/pull/636

PR #638, opened 2018-10-02
Deprecate craftLayerUrl()

cherewaty:
`craftLayerUrl()` assists with `<SearchAndSort>`'s routing, and will be deprecated for any other use (modules should define their own routes, instead of relying on `<SearchAndSort>` to do it for them).


cherewaty:
Related to https://github.com/folio-org/stripes-smart-components/pull/290

PR #627, opened 2018-09-29
Create wider type stack with consistent variables

cherewaty:
## Purpose
Have enough levels of hierarchy in font sizes (esp. for headlines) that laying out pages is easy.

## Approach
I created these variables, based on the very fresh CSS Fonts Module Level 3 recommendation:  https://www.w3.org/TR/css-fonts-3/#propdef-font-size
```
  --font-size-xx-small: 0.67rem;
  --font-size-x-small: 0.75rem;
  --font-size-small: 0.83rem;
  --font-size-medium: 1rem;
  --font-size-large: 1.17rem;
  --font-size-x-large: 1.5rem;
  --font-size-xx-large: 2rem;
```
The global base is set at 16px, as of https://github.com/folio-org/stripes-components/pull/611

I then updated every use of `font-size` in `stripes-components` to use one of those seven sizes.

The biggest restriction was that the font size of `<input>` elements need to be >=16px.

## Results
I generally set `<label>` and label-like elements to `--font-size-small`:

screen shot 2018-09-28 at 10 03 51 pm

screen shot 2018-09-28 at 10 04 36 pm

Pane headers:
screen shot 2018-09-28 at 10 05 03 pm

Modals:
screen shot 2018-09-28 at 10 05 36 pm

Buttons, at "small" font size by default:
screen shot 2018-09-29 at 8 46 19 am

Filter groups:
screen shot 2018-09-30 at 11 55 25 am

Headlines:
screen shot 2018-09-28 at 10 10 00 pm

## Next steps
We'll probably want to tweak what size certain components use, but this stack should last as a flexible set of sizes that match the proportions browser vendors have converged on. Nearly any `font-size` declaration in any FOLIO module should use one of these variables.

PR #652, opened 2018-10-11
Polish form control sizing

cherewaty:
## Purpose
With the type stack changes in https://github.com/folio-org/stripes-components/pull/611 and https://github.com/folio-org/stripes-components/pull/627, form controls were no longer consistently sized. `<Button>` was significantly shorter than `<TextField>`, and `<Select>` seemed to miss a few memos. That made composing those things together a bit of a mess.

## Approach
- Lean more heavily on the existing `--input-horizontal-padding` and `--input-vertical-padding` CSS variables (which in most cases, were used backwards of their naming). Those values were converted to `rem`, but are otherwise unchanged.
- Make `<TextField>`, `<Select>`, `<Button>`, etc. all the same height through `padding` and `line-height`, instead of setting an explicit `height`.
- The global letter spacing wasn't being reflected with form controls in Storybook. Fixed.
- The `--radius` variable is now being consistently used.
- The `--shadow` variable is now being consistenly used.
- The `<Timepicker>` dropdown uses `<TextField>` now instead of bare `<input>`s.
- `16px` is now the go-to `gutter` size, since that makes the math really easy (`1 rem`).
- `em` or `rem` nearly always preferred over `px`
- `MultiColumnList` layout tightened up; it had been less dense with the move from 14px to 16px base

## Next steps
- Should the icon, icon button, and app icon sizes be revisited? Maybe no changes needed, but worth an evaluation.
- Should we create a touch target that surrounds buttons that's larger than their visual size?

### Before
screen shot 2018-10-10 at 6 56 52 pm

localhost_9001__selectedkind textfield selectedstory basic 20usage full 1 addons 1 stories 1 panelright 0 addonpanel react_storybook 2freadme 2fpanel ipad

### After
screen shot 2018-10-10 at 6 56 28 pm

localhost_9001__selectedkind textfield selectedstory basic 20usage full 1 addons 1 stories 1 panelright 0 addonpanel react_storybook 2freadme 2fpanel ipad 1

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,7 +1,7 @@
 <style type="text/css">
   body {
     margin: 0;
-    padding: 15px;
+    padding: 1rem;
     position: relative;
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add clone option to EntrySelector. Fixes STCOM-364.
 * Shrink `<MetaSection>` font sizes
+* Polish form control sizing
 
 ## [4.1.0](https://github.com/folio-org/stripes-components/tree/v4.1.0) (2018-10-01)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v4.0.0...v4.1.0)
--- a/lib/Accordion/Accordion.css
+++ b/lib/Accordion/Accordion.css
@@ -23,7 +23,6 @@
   width: 100%;
   display: flex;
   align-items: center;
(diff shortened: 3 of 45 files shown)


rasmuswoelk:
Nice work! Looks great 👍

zburke:
99% of this looks great, except that checkboxes now look like radio buttons to me. In my head, since the beginning of time, checkboxes have been square and radio buttons have been round. Is that still a convention or not really?

### Before
before

### After 
after

### Forever
netscape

rasmuswoelk:
@zburke I just noticed this too. They should remain square imo.

PR #661, opened 2018-10-16
Updated border-radius on .labelCheck to 25% 

rasmuswoelk:
Prevents checkboxes from becoming circular.


cherewaty:
Thanks @rasmuswoelk 👏 I definitely didn't intend to make those circular.

PR #669, opened 2018-10-26
Updated margins/paddings for MCL

rasmuswoelk:
Updated `--gutter` variable to correspond to 15px when computed (on a font-size of 16px). 

Updated padding/margin for MCL header, row, and cells using the `-gutter` variable so that there is the correct spacing from edge to content (15px).

Ideally, the gutter should remain a static `15px` which is the system-wide default gutter padding/margin. For now, I have changed it into a corresponding rem-value. We can later on decide to change it into pixels and the spacing will still match.

## Before
before

## After
after

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.css
+++ b/lib/MultiColumnList/MCLRenderer.css
@@ -32,8 +32,8 @@
   border-radius: var(--radius);
   color: var(--color-text);
   text-decoration: none;
-  margin-left: 0.25rem;
-  margin-right: 0.25rem;
+  margin-left: calc(var(--gutter) / 3);
+  margin-right: calc(var(--gutter) / 3);
   outline: none;
 
   &:visited {
@@ -82,7 +82,7 @@
   align-items: center;
   flex-shrink: 0;
   flex-grow: 0;
-  padding: 0.5rem 0.5rem;
+  padding: 0.5rem var(--gutter);
   font-weight: 600;
   color: var(--color-text-p2);
   font-size: var(--font-size-small);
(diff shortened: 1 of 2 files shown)


lib/MultiColumnList/MCLRenderer.css
```diff
@@ -110,7 +110,7 @@
   flex-shrink: 0;
   flex-grow: 0;
   line-height: var(--line-height);
-  padding: 0.125rem 0.5rem;
+  padding: 0.125rem calc(var(--gutter) * 0.666666666666667);
```
cherewaty: I try round to a max of three decimal places in CSS. Why not just take it down a pixel to `0.125rem` for cleaner math?

lib/variables.css
```diff
@@ -25,7 +25,7 @@
   --success: #070;
   --control-margin-bottom: 1rem;
   --radius: 6px;
-  --gutter: 1rem;
+  --gutter: 0.9375rem; /* Computes to 15px when body font-size is 16px */
```
cherewaty: Why `15px` vs. `16px` (`1rem`)? If there's not a specific reason for cutting that single pixel, a gutter of `1rem` makes for much more elegant calculations.

JohnC-80:
Inconsistent paddings between headers and cells make this a tough one to accept. It makes the component even less durable in the end because the places where you see a difference are even more misaligned. See Users>Detail view> Loans for an example of a table that pushes the limits.

lib/variables.css
rasmuswoelk: I'm pretty sure that the default gutter has been 15px from day one which is why I updated the values to match 15px when on a 16px font-size. Ideally, the --gutter is changed back to a static 15px so that it matches the original design system.

E.g. the interaction styles use the gutter and half/third variations of it. That makes for a clean 15px, 10px and 5px. Similar goes for the MCL (or at least in my changes in this PR).

Personally, I don't really care if we switch it to 16px but you would have to discuss it with @filipjakobsen. As always, there's probably a good reason for the designers to choose exactly 15px for gutters.

rasmuswoelk:
Closing because #671 works better.

PR #671, opened 2018-10-26
Fix MCL header padding

cherewaty:
Attempts to solve the same problem as https://github.com/folio-org/stripes-components/pull/669, where the `MultiColumnList` headers are not aligned with their column content.

Maintains `1rem` (`16px`) gutter, not `15px`.

### Before
folio-testing aws indexdata com_users_view_filters active include 20inactive 20users sort name ipad 1

### After
folio-testing aws indexdata com_users_filters active include 20inactive 20users sort name ipad

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.css
+++ b/lib/MultiColumnList/MCLRenderer.css
@@ -32,7 +32,7 @@
   border-radius: var(--radius);
   color: var(--color-text);
   text-decoration: none;
-  margin: 0 0.5rem;
+  margin: 0 calc(var(--gutter) / 2);
   outline: none;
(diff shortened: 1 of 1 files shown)


JohnC-80:
I prefer this over #669 since it keeps consistent padding amounts between headers and cells. 👍

rasmuswoelk:
I closed my issue because this seems to work better.

But can we use the `--gutter` variable for the paddings? (--gutter / 2 instead of 0.5rem). I think it shows other developers that the number isn't totally random.

cherewaty:
Updated to use `--gutter`.

PR #673, opened 2018-10-27
Shrink plus sign icon

cherewaty:
The `plus-sign` icon was significantly bigger than its siblings `closeX`, `right-arrow`, `left-arrow`, etc.

Noticed because we should start using the `plus-sign` icon in "+ Add" and "+ New" buttons so screen readers read them as "add" and "new" instead of "plus add" and "plus new".

The deprecated `ui-scan` repo is the only current use of `plus-sign` in the `folio-org`.

### Before
screen shot 2018-10-27 at 12 08 47 pm

### After
screen shot 2018-10-27 at 12 08 34 pm


cherewaty:
Blocked on https://issues.folio.org/browse/STCOM-386 (but should be taken care of there)

PR #674, opened 2018-10-28
STCOM-386 Resize icons

cherewaty:
## Purpose
To get around icons' lack of flexibility, we were starting to duplicate them with `inline` variants:
https://issues.folio.org/browse/STCOM-386

## Approach
Treat all icons as if they're character glyphs meant to sit inline with text.

- Render icons by default at 1em by 1em, so they just fit their context.
- Changed the `small`/`medium`/`large` props to use `em` that corresponds to the font size ratios defined in `variables.css`.
- Used a consistent `viewBox` that matches the base font size (`0 0 16 16`).
- With most of the icons, I was able to simply crop them from `24x24` to `16x16` without changing the art size. A few needed to be shrunk a little to appropriately fix in the new `viewBox`. At some point, we should do a fresh pass at redrawing all of these for consistent line weights (low priority, I don't see any that are so far off they look bad).
- Added padding to `IconButton` and other uses of Icon, since the SVGs themselves now have far less padding.
- Icons now inherit their color.
- Removed `down-caret-inline` and `up-caret-inline`.

### Before
before

### After
after


cherewaty:
@cchiama I couldn't tag you as a reviewer, but would love your eyes on this too.

filipjakobsen:
Hi Jeffrey,

The reason for having different icons is not that we did not realize they needed to be universally applicable from the beginning, but because the icons are designed to be used in a specific size and context. They are vector graphics, so they can scale if everything else scales, but they are designed to be relative to either A) the pane and accordion headers, or B) the text. The stroke-to-width proportions are going to be off if we go forward with this, making some icons too thin or too thick, depending on which set you delete.

cherewaty:
If all the icons are designed at `16x16` to sit next to `16px` text, they'll appropriately scale at the correct ratios. The line widths will be correct for the text at that size. This is the way icon font libraries like https://fontawesome.com and https://useiconic.com/open/ are structured.

I added some more examples to a new story:
### Before
before

### After
after

filipjakobsen:
@cherewaty, I understand how the math works, and I can appreciate the pragmatism of your suggestion. I am just saying that this is not how the icons were intended to work.

I think this has some downsides from a design perspective, but I can also see why it might be easier to understand and use for app developers, so if we can make sure we update a few other things, then I think we can proceed with this:

1. We need to update all the icon files, so they have a consistent stroke-to-width ratio. I can do that, and provide them all in 16x16 px SVG (as I have mentioned before, this does not mean I agree that we should use a 16px font size as default on desktop)
2. Someone needs to make sure the sizes of icons in the pane header and universal header are updated relatively to the rescaling of the icons. Perhaps @rasmuswoelk can help with that?
3. We need to accept that in time there might be a few places in the system where we need to break away from this and provide purpose-specific vector graphics that do not fit the icon guidelines put forth in this PR. These must then be implemented as something other than an icon component if necessary.

@cherewaty: Jeffrey, how does this sound to you?

lib/Datepicker/Datepicker.js
```diff
@@ -433,29 +432,27 @@ class Datepicker extends React.Component {
...
             id={`datepicker-clear-button-${this.testId}`}
             tabIndex="-1"
-          >
-            <Icon icon="clearX" />
-          </Button>
+            icon="clearX"
+          />
         )}
-        <Button
+        <IconButton
```
JohnC-80: I'm not sure we can simply switch to IconButton in Datepicker and TimePicker as-is - it will need something to kill its margins that inflate the spacing when used in these types of controls... also, the buttonStyle prop doesn't apply to IconButton, since it doesn't simply wrap `<Button>`.

JohnC-80: @rasmuswoelk, @cherewaty  what do you guys think of killing the margin on `<IconButton>` altogether? Let the extra padding there just do the spacing?

rasmuswoelk: I don't have any problems with that. The most important thing is that it has a clickable face of at least 24x24px on desktop and 44x44px on touch devices.

cherewaty: Yeah, I should take the margin off `IconButton`.

cherewaty:
@filipjakobsen I agree we'll need to take another pass at all the icons to ensure the consistent line width polish we want 👌 

I'm wondering if `32x32` SVGs would be easier to work with (more pixels on the grid to work with vs. `16x16`, reflects the pixels on a retina screen). Thoughts?

filipjakobsen:
@cherewaty, I do not have a strong opinion on the artboard size. Would you like me to make the icons 32x32?

lib/Datepicker/Datepicker.js
cherewaty: `IconButton` usage issues fixed. I added margins to `IconButton`s only when they sit next to each other (like we do with `Button`):

cherewaty:
@filipjakobsen yep, let's roll with 32x32. I've resized all of them to a `0 0 32 32` `viewBox` now.

cherewaty:
Follow-up story for redrawing icons: https://issues.folio.org/browse/STCOM-390
@rasmuswoelk @filipjakobsen let me know if I can be of any help with that

PR #677, opened 2018-10-29
Layout: Added more flex props 

doytch:
Thought we were lacking a few props that I frequently use with flexbox, specifically `flex-direction: column` and `justify-content: flex-end`.

And while I was adding `justify-content`, I figured I'd pop in the whole suite.


rasmuswoelk:
@doytch Nice with some additions to the layout component!

I know it's a bit manual extra work but can you add the new styles in the file `classNamePropertiesMapping.js`. 

It's used in the storybook to easily filter and find the class name you need:

image

doytch:
While doing that I realised that I actually duplicated a couple of `justify-content` rules so I removed those classes that I'd be duplicating with. 

The nasty thing is that the existing classes don't really have a uniform naming pattern and to fix it it'd technically be a breaking change to rename `.justified` and `.centerContent`. 🤷‍♂️

rasmuswoelk:
@doytch Yeah.. some of the class names were kept when I updated the component to avoid things from breaking. Some of the older ones should probably be phased out.

PR #679, opened 2018-10-30
MCL Tests

JohnC-80:
# Purpose
A layer of tests for basic rendering of data, transitions between common props, and windowing as we approach a refactor.

## Approach
I employed `faker.js` to generate fake data via a utility function. The tests include a few raw component tests and a few within a harness to test transitions between props (loading data, visible columns etc) in a similar fashion to the way MCL might be used within an app.
I ran into a bit of difficulty trying to use BigTest's `scrollable` interactor, and ended up having to perform an equivalent via `triggerable`.


lib/MultiColumnList/tests/interactor.js
```diff
@@ -1,9 +1,66 @@
...
+  columnCount = count(`.${css.header}`);
+  rows = collection(`.${css.row}`, RowInteractor);
+  headers = collection(`.${css.header}`, HeaderInteractor);
+  scrollBodyTop = getTop(`.${css.rowContainer}`);
+  displaysEmptyMessage = isPresent(`.${css.emptyMessage}`);
+  displaysLoadingIcon = isPresent(`.${css.contentLoading}`);
+
+  scrollBody = function (amount) {
+    this.$(`.${css.scrollable}`).scrollTop += amount;
+    return triggerable(`.${css.scrollable}`, 'scroll');
```
wwilsman: This is exactly what `scrollable` is supposed to do when passed `{ top: amount }`.

```js
scrollBody = scrollable(`.${css.scrollable}`)
// ...
mcl.scrollBody({ top: amount });
```

Any more info on what issue you were having with it? Was it just not scrolling properly?

Also, `triggerable` returns a property descriptor, you want `this.trigger(selector, 'scroll')` here instead.

JohnC-80: Thanks for the tip on 'trigger' vs 'triggerable'. I've still gotta get my head around property descriptors and the primary flow of data through these interactors. 
The list didn't scroll and the handler was never called :/ Looking at the guts of scrollable basically gave me the idea for the way I ended up setting this up. I set some breakpoints within scrollable - the actual code that sets the scrollTop and dispatches the 'scroll' just was not getting executed for some reason... almost as if it just failed to find the selector and crapped out - but I didn't get a console error or anything so I'm really not sure.

wwilsman: If it didn't execute the code inside the `.do` the `find` more than likely failed. Interestingly, `find` converges on `this.$(selector)` which is exactly what you have here too. 

I can investigate a bit more and see if there's an issue with interactor that should be opened over there. This is a good workaround for now though.

lib/MultiColumnList/tests/MultiColumnList-test.js
```diff
@@ -1,23 +1,355 @@
...
+            onNeedMoreData={() => { moreRequested = true; }}
+            autosize
+          />
+        </div>
+      );
+    });
+
+    describe('scrolling the list', () => {
+      beforeEach(async () => {
+        mcl.scrollBody(200);
```
wwilsman: I almost didn't catch this until I went to debug scrollable, we should await on all of these interactions

lib/MultiColumnList/tests/MultiColumnList-test.js
```diff
@@ -1,23 +1,355 @@
...
+      });
+
+      it('calls the supplied onScroll handler', () => {
+        expect(scrolled).to.be.true;
+      });
+    });
+
+    describe('scrolling the list more', () => {
+      beforeEach(async () => {
+        mcl.scrollBody(2500);
```
wwilsman: Looks like it was just the last one and this one that need `await`

lib/MultiColumnList/tests/MultiColumnList-test.js
```diff
@@ -1,23 +1,355 @@
...
+import generate from './generate';
+
+const data = generate(10);
+const data3 = generate(3);
+const data100 = generate(100);
+
+let scrolled = false;
+let moreRequested = false;
+
+describe.only('MultiColumnList', () => {
```
wwilsman: `only` ✂️

JohnC-80: dammit - gets me every time.

lib/MultiColumnList/tests/interactor.js
wwilsman: I pulled this branch to test scrollable and it worked. Maybe it was related to the missing `await`s? That would definitely cause an error to not be raised.

Instead of actually using `scrollable` and rewriting the tests to do `mcl.scrollBody({ top })` I made `scrollBody` utilize `this.scroll`.

```js
// same thing as `scrollable` but allows us to only use scrollTop
scrollBody(amount) {
  return this.scroll(`.${css.scrollable}`, { top: amount });
}
```

JohnC-80: 👍 yeah the missing awaits must have been it.

lib/MultiColumnList/tests/MultiColumnList-test.js
```diff
@@ -1,23 +1,354 @@
...
+const data = generate(10);
+const data3 = generate(3);
+const data100 = generate(100);
+
+let scrolled = false;
+let moreRequested = false;
 
 describe('MultiColumnList', () => {
   const mcl = new MultiColumnListInteractor();
+  const clicker = new Interactor();
```
wwilsman: I don't know how I feel about this. On one hand, this is a perfectly valid use of interactors, but on the other hand, we should keep as much interactor functionality as we can with the component interactor.

My main point with this is it would be really sweet to be able to just import these component interactors into our acceptance testing suites in all the various UI modules. That way we or other devs won't have to redefine interactors for components that already have interactors.

Would it be useful to include these couple non-custom interactors within the component's interactor?

JohnC-80: My thinking was that it's interacting more with the MCLHarness rather than the MCL itself. I suppose I could have imported Button interactors into this to do the same thing. I introduced another interactor later to check the presence of the custom formatter - that one definitely would have a place in the component interactor.

wwilsman: That makes sense. In that case this one is probably fine, but I would move the other one.

JohnC-80: Alright.. in the latest commit I removed the interactor that checked for the formatter and added a 'contains' method to the component interactor to fulfill the same function within the rendered MCL. I used `$$` - please let me know if there's any better way to check for the presence of an element given a selector parameter.

lib/MultiColumnList/tests/interactor.js
```diff
@@ -1,9 +1,69 @@
...
+  scrollBodyTop = getTop(`.${css.rowContainer}`);
+  displaysEmptyMessage = isPresent(`.${css.emptyMessage}`);
+  displaysLoadingIcon = isPresent(`.${css.contentLoading}`);
+
+  scrollBody(amount) {
+    return this.scroll(`.${css.scrollable}`, { top: amount });
+  }
+
+  contains(selector) {
+    return this.$$(selector).length > 0;
```
wwilsman: This is exactly what the `isPresent` property creator does, so I think this is perfect since we want to pass in a selector.

Another option is `!!this.$(selector)`. The difference is `$` will throw an error if the element cannot be found while this length check will return `false`.

Deciding between the two options, would there be a test where we want to expect this to be false?

For example, this will work with `$$` but not `$`:

```js
expect(mcl.contains(selector)).to.be.false
```

wwilsman: I'm leaning towards keeping this length check, but just thought I'd share this other option

JohnC-80: Thanks! I don't have a solid, but the use case is somewhat arbitrary, so I feel better about having a negative boolean returned. Gotta figure out what's up with the scrolling...

wwilsman: One thing it could be is the stuff in the list hasn't rendered yet, but the container has. In this case there is no room to scroll. I believe in eholdings, we wait for records to exist in the DOM before scrolling.

This could probably be an error thrown by interactor. Something like `Cannot scroll container, no overflow`. Of course this would be converged on, so it wouldn't error until after 2s, and if it does pass within that time, it will scroll.

To do that ourselves, we can utilize `.when` in the scroll method.

```js
return this
  .when(() => this.$(`.${css.scrollable}`).innerHeight >= amount)
  .scroll(`.${css.scrollable}`, { top: amount });
```

That would throw a generic `convergence returned false` if it fails though. A more helpful error would look like:

```js
return this
  .when(() => {
    if (this.$(`.${css.scrollable}`).innerHeight < amount) {
      throw new Error('unable to scroll MCL')
    }
  })
  .scroll(`.${css.scrollable}`, { top: amount });
```

wwilsman: https://github.com/bigtestjs/interactor/issues/59 😃

JohnC-80: I haven't seen a test failure since the one that held this up - locally or in Jenkins, so I'm not sure what specifically happened there. The MCL forgives bottoming out in attempting to load more items (which is what's happening in these tests.) Something else is causing branch to fail in Jenkins :/

JohnC-80: If a scrollTop is beyond the scrollHeight of an element, the maximum just gets applied. I'm not sure if throwing is quite the right thing at that point, because (at least in this case) I would still want the scroll to happen. The goal for these tests is to trigger the MCL's callback to use for requesting more data. I haven't seen that particular failure again since I mentioned the issue.

PR #681, opened 2018-10-31
Improve dev tools output for forwardRef

cherewaty:
## Purpose
When prepping a PR to `react-intl` to share our `forwardRef` usage upstream, I noticed some sloppiness around how we label components using `InjectIntl`. 

## Approach
- Now components using `injectIntl()` that don't opt into `withRef` have one less layer of component tree.
- `hoist-non-react-statics` helps with sending around `displayName`.
- The `intlPropName` change just aligns our fork closer to the upstream `InjectIntl`.

### Before
screen shot 2018-10-30 at 6 53 27 pm

### After
screen shot 2018-10-30 at 6 52 23 pm

### DIFF
--- a/lib/InjectIntl/InjectIntl.js
+++ b/lib/InjectIntl/InjectIntl.js
@@ -1,14 +1,17 @@
 import React, { Component, forwardRef } from 'react';
 import PropTypes from 'prop-types';
+import hoistNonReactStatics from 'hoist-non-react-statics';
 import { intlShape } from 'react-intl';
 
+function getDisplayName(WrappedComponent) {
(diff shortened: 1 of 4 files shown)


JohnC-80:
Seems to trade a double appearance of ForwardedRef for a double-appearance of `InjectIntl(<Component>)` - but Better than having 'undefined' show up, due to the cut off display name.... would the change to ReduxFormField alone have been enough to cure the 'undefined'  component name?

cherewaty:
The change to `ReduxFormField` alone would just start printing `ForwardRef(ReduxFormField(Component))`.

Yeah, there's still some weirdness around the `TextField` tree.

PR #683, opened 2018-10-31
Move icon sizing to SVG instead of root element

rasmuswoelk:
The size of the icon should not affect the font-size of the label.

E.g. when using an `<Icon>` inside of a `<Button>` we need the label to enherit the font-size from the button.

I also adjusted the spacing between icon and label.

## Before
screenshot 2018-10-31 11 30 24

## After
screenshot 2018-10-31 11 39 14

### DIFF
--- a/lib/Icon/tests/Icon-test.js
+++ b/lib/Icon/tests/Icon-test.js
@@ -13,7 +13,7 @@ import icons from '../icons';
 import css from '../Icon.css';
 import Icon from '../Icon';
 
-describe('Icon', () => {
+describe.only('Icon', () => {
   const icon = new IconInteractor();
   const iconClassName = 'icon-class';
   const iconRootClass = 'icon-root-class';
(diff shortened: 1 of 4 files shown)


lib/Icon/tests/Icon-test.js
```diff
@@ -121,21 +122,21 @@ describe('Icon', () => {
...
+
+  //     it('It should render a svg', () => {
+  //       expect(icon.svg.isPresent).to.be.true;
+  //     });
+
+  //     it(`SVG should include a className of "${iconClassName}"`, () => {
+  //       expect(icon.svg.className).to.include(iconClassName);
+  //     });
+  //   });
+  // });
```
cherewaty: Did you intend to comment these out?

rasmuswoelk: Nope! That was a mistake. These need to be uncommented. Thanks for noticing 😅

rasmuswoelk: Fixed.

PR #689, opened 2018-10-31
STCOM-391 Use plus-sign icon instead of + character

cherewaty:
Stops screen readers from announcing "plus": https://issues.folio.org/browse/STCOM-391

### Before
screen shot 2018-10-31 at 11 39 48 am

### After
screen shot 2018-10-31 at 11 39 01 am


cherewaty:
Related to https://github.com/folio-org/stripes-smart-components/pull/323

PR #693, opened 2018-11-01
Allow node prop types for labels and validation messages

cherewaty:
Expands prop type validation for `error`, `label`, and `warning` props to make internationalization easier (more `<FormattedMessage />` from `react-intl`, less `formatMessage()`).

Set up for `v4.3.1` release.


cherewaty:
Of note: `PropTypes.node` includes `PropTypes.string`: https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes

PR #696, opened 2018-11-03
Remove injectIntl() fork

cherewaty:
## Purpose
When using `formatMessage()` from `react-intl`, it needs to be accessed from the `intl` object. There a couple different ways we've done that:
- Make the `intl` object available on a global `stripes` god object. We've deprecated that anti-pattern.
- Use `injectIntl()` from `react-intl`.

`injectIntl()` has its own custom API for dealing with refs, since it pre-dates `forwardRef` introduced as the standard in React 16.3. In order to use a more current-React-like API, we created a `injectIntl()` fork in `stripes-components`.

This PR presents a better option.

## Approach
- Exclusively use `<FormattedMessage>`, `<FormattedDate>`, and `<FormattedTime>` from `react-intl`. 
- Discontinue using `intl.formatMessage()`, `intl.formatDate()`, and `intl.formatTime()`.
- Delete `injectIntl()` fork.

## Notes
- `<Datepicker>` and `<Timepicker>` still use legacy React context (as did the `injectIntl()` fork that they previously used).
- All `folio-org` repos should be evaluating their use of `injectIntl()`. There should be very few cases where we need to use `react-intl`'s functions instead of components.


lib/Selection/SelectList.js
```diff
@@ -297,24 +298,32 @@ class SelectList extends React.Component {
...
-                { id: 'stripes-components.selection.filterOptionsPlaceholder' }
+            <FormattedMessage
+              id="stripes-components.selection.filterOptionsPlaceholder"
+            >
+              {placeholder => (
+                <FormattedMessage
+                  id="stripes-components.selection.filterOptionsLabel"
+                  values={{ label: this.props.label }}
+                >
+                  {ariaLabel => (
```
wwilsman: I really dislike this formatted message nesting in order to get two formatted strings. But I see why it's necessary to remove `injectIntl`. I can't think of a better way other than wondering how difficult it would be to have a `FormattedMessages` component that could handle multiple message ids. That would be something to contribute back to `react-intl` though.

PR #698, opened 2018-11-03
Update Modal propTypes

cherewaty:
## Problem
It should be possible to give a `label` for the `<Modal>` header that's a `node`, not just a `string` (makes dealing with `react-intl` `<FormattedMessage>` much easier).
 
But `aria-label` needs to be a string, and `label` was being reused for both purposes.

## Approach
- Split the `ariaLabel` and `label` props.
- Accept a `node` for `label`.

### DIFF
--- a/lib/Modal/tests/Modal-test.js
+++ b/lib/Modal/tests/Modal-test.js
@@ -69,7 +69,7 @@ describe('Modal', () => {
     });
 
     it(`It should have an aria-label of "${label}"`, () => {
-      expect(modal.ariaLabel).to.equal(label);
+      expect(modal.ariaLabel).to.equal(null);
     });
 
     it(`It should render with a label of "${label}"`, () => {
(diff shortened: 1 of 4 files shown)


rasmuswoelk:
Looks good to me. I guess the readme should be updated as well?

lib/Modal/tests/Modal-test.js
```diff
@@ -69,7 +69,7 @@ describe('Modal', () => {
     });
 
     it(`It should have an aria-label of "${label}"`, () => {
```
wwilsman: The title of this test should change, and probably a new one added for when `ariaLabel` is provided.

Or pass `ariaLabel` in the setup and change this test entirely.

wwilsman:
+1 on the modal readme needing an update

JohnC-80:
My concern is that the lack of aria-label (only a label node provided) leaves screen-readers in the dark when the modal is focused. This could be accounted for by applying an aria-labelledby prop to the dialog if no `aria-label` is supplied. OTOH, a screen-reader may simply relay the nearest inner <h*> tag - need to test it out. - potential future iteration.

cherewaty:
@JohnC-80 should I make `ariaLabel` required?

JohnC-80:
@cherewaty I wouldn't, since it is an optional treatment.

cherewaty:
Added commit with updated tests and readme.

PR #699, opened 2018-11-04
Replace <IconButton> with <Button buttonStyle="plain">

cherewaty:
## Problem
*Icons need to be localizable.*

`<IconButton>` makes an assumption that we can't make - that it will only ever contain an icon. Places where we place an icon in a button also need to accomodate the possibility of words next to or in place of the icon. Translators will make the call whether the icon is appropriate for their locale, or if it needs to just be words.

Related to https://github.com/folio-org/stripes-smart-components/pull/323

## Approach
- Deprecate usage of `<IconButton>`.
- Create new `buttonStyle="plain"` variant for `<Button>`.
- Because we will now use `<Button>` across the board, the clickable area of icons used this way increases from 24px tall to 28px. This matches the height of other buttons, `<TextField>`, `<Select>`, etc. for much easier layout composition.
- Introduce `<Button>` sizes that match the `stripes-components` type scale: `xx-small`, `x-small`, `small`, `medium`, `large`, `x-large`, `xx-large`
- `slim` variant of `<Button>` still available and useful for placing more than one `<Button><Icon /></Button>` inside a `<TextField>`.

### Before
```
<IconButton icon="plus-sign" onClick={() => {}} />
```

### After, without internationalization
```
<Button onClick={() => {}} buttonStyle="plain">
  <Icon icon="plus-sign" />
</Button>
```

### After, with internationalization
```
<Button onClick={() => {}} buttonStyle="plain">
  <FormattedXmlMessage
    id="stripes-components.newButton"
    tags={{ 'icon': (<Icon />) }}
  />
</Button>
```

### Possible permutations of the string in a translation file
```
"new": "<icon icon='plus-sign'>New</icon>"
```
```
"new": "<icon icon='plus-sign' />"
```
```
"new": "New"
```

### With a badge
```
<Button buttonStyle="plain">
  <Icon icon="comment">
    <Badge>9+</Badge>
  </Icon>
</Button>
```

## Additional notes
- Includes layout changes to `<Modal>` headers to not have to deal with `z-index`.
- Timepicker test that broke with DST ending skipped for now.

## Next steps
- Wrap every icon in `stripes-components` in `<FormattedXmlMessage>` so it can be translated as necessary.


cherewaty:
@filipjakobsen is going to have some conversations around i18n and icons to validate if we need this approach.

JohnC-80:
I like this a lot.
Decreasing the Component surface by one: 👍 
Despite the pane dropdowns being a thing, I can still imagine real estate being at a premium in those pane headers. Does 'slim' usage throw much off there?

rasmuswoelk:
I guess the only thing that IconButton provides that Button doesn't currently provide is the option to have a bigger clickable face while preserving the same hover/active/focus styles on the inner element.

We could solve this by adding hover, active and focus styles to the inner element instead of the button itself. Then the `<button>`/root element could control the clickable face. 

We would need this in PaneHeader's and ModalHeaders to make sure we have at least 44x44px clickable face.

The only other "objection" I have for removing the IconButton is that developers need to make sure to add the correct props to the Button with the right combination of other components (Badge, Icon etc.) to achieve the same thing. That can lead to inconsistencies and we have less "central control" of the UI. But I guess this could be solved by proper documentation.

PR #704, opened 2018-11-07
AutoSuggest: Added custom search/render functionality

doytch:
AutoSuggest is pretty prescriptive about what it expects from its parents and as a result, wasn't flexible enough to use as-is in the ERM modules. It currently drives everything off of the `value` field of an object which is limiting.

This PR gives us the ability to custom render the value and dropdown items, as well as construct more elaborate searches.

Going forward, this component could use a pass in the beauty parlour but I'm hoping to get at least this _functionality_ in now since I'd like to use it asap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * `color` chokes on rgb/rgba; use hex values instead.
 * Validate presence of `props.onClick()` before calling it.
 * Allow node prop types for labels and validation messages
+* `<AutoSuggest>` updated with custom search and render functionality.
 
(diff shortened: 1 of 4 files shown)


doytch:
I didn't bother to bump the patch version. Does it even matter anymore now that we don't have the ability to require a specific `stripes-components` version from our UI modules since we're just depending on `@folio/stripes`?

JohnC-80:
@doytch yeah, the stripes-components dep is only a stripes dependency now (well, compatible version is still expressed on other stripes-* modules.) Modules shouldn't have to fight over what version gets brought in. Of course, we'll be bumping at least the patch with the next release to npm-folio. Please let us know if you run into any problems with the stripes setup.

PR #705, opened 2018-11-07
STCOM-388: Allow for more advanced action menus

rasmuswoelk:
We need a way to compose and pass a more complex action menu to a PaneHeader.

Currently, you can pass an array of objects to the `actionMenuItems`-prop to render a simple action menu. @JohnC-80 and I decided that we should only have one way of adding action menus so I added a deprecation warning for `actionMenuItems` but kept the functionality for now.

**In this PR I have changed the following:**
- Added `actionMenu`-prop that expects a function. That function should return a node or component
- Added deprecation warning for `actionMenuItems`-prop
- Updated examples and readme

**Next up:**
- Add components that can be useful for composing complex action menus (https://issues.folio.org/browse/STCOM-389)
- Update existing apps to use the new actionMenu prop so that we can remove the `actionMenuItems` prop


cherewaty:
I like this new `actionMenu` prop a lot 👏

PR #708, opened 2018-11-08
include data-total-count attribute in output

zburke:
Previously, the row-count was included in the `title` attribute of the
table header, but `title` isn't a11y-friendly so we've been getting rid
of it. Nonetheless, we still need access to the row-count
programmatically for tests, so here it is.

An alternative would be to include a `data-test-row-count` attribute and
to include `data-test-*` attributes in our builds, but we don't have
those builds yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Remove `injectIntl()` fork
 * Simplify `ReduxFormField` usage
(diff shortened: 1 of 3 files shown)


zburke:
Refs [STCOM-257](https://issues.folio.org/browse/STCOM-257)

PR #714, opened 2018-11-12
Introduce tagName prop on Pane

cherewaty:
## Purpose
Allows `<Pane>` to be a `<form>` and applies any props to the top level element, so we won't have to wrap `<Pane>` in any other kind of tag.

Current spots where this is a problem:
- https://github.com/folio-org/ui-inventory/pull/353
- eHoldings settings: http://folio-snapshot.aws.indexdata.com/settings/eholdings/knowledge-base

## Change
This moves the `...rest` props spread from `<PaneHeader>` to the top level element of `<Pane>`. May have unknown side effects.


wwilsman:
> This moves the `...rest` props spread from `<PaneHeader>` to the top level element of `<Pane>`. May have unknown side effects.

At first glance it doesn't appear `<PaneHeader>` does any `...props` currying. So the only prop we're missing with moving it is `dismissible`. The rest of the `<PaneHeader>` props seem to be accounted for.

cherewaty:
Thanks for checking @wwilsman 👌 I suppose I could've looked. Commit added that passes through `dismissible`.

PR #719, opened 2018-11-14
UIU-689: Move PasswordStrength component to stripes-components.

OleksandrAntonenko1:


cherewaty:
I think I remember there being a test for this component at some point, but we didn't have the infrastructure in place in `stripes-smart-components` yet. Could we get a test and an interactor set up here?

OleksandrAntonenko1:
Tests will be added in scope of other story.

PR #723, opened 2018-11-20
STCOM-389: Add <MenuSection> component

rasmuswoelk:
**Issue:** https://issues.folio.org/browse/STCOM-389

This PR adds a new component called `<MenuSection>`. 

This component is useful for creating multisectioned dropdown menu's (which is the current use-case but it can, of course, be used in other parts of the UI as well).

It basically renders a section with a label. When stacking MenuSections it will automatically apply the appropriate margins.

I also updated the styles for the `dropdownItem` buttonStyle as well as updated the `<Button>` examples/readme.

## Changes:
- Added MenuSection component
- Added examples, readme, and tests for MenuSection
- Updated dropdownItem styles and updated Button example/readme

## Test results
image

## Example of MenuSection's inside a Dropdown
image

**Note:** In an upcoming PR I will adjust the margins/paddings of various components to make sure that they align properly when stacked like the above example.

## Next up:
- Update app's using the `actionMenuItems`-prop to use `actionMenu`-prop instead
- Update documentation regarding Pane Action Menu's to include an example of MenuSection and Button w. buttonStyle "dropdownItem"


lib/Button/Button.css
```diff
@@ -1,5 +1,10 @@
 @import '../variables.css';
 
+:root {
+  --button-default-vertical-padding: 0.125rem;
+  --button-default-horizontal-padding: var(--gutter);
+}
```
cherewaty: We should probably only define global variables in `variables.css`.

rasmuswoelk: I was thinking about that but I chose this path because this variable isn't global. But I can move it to the global variables 👍

lib/MenuSection/MenuSection.js
```diff
@@ -0,0 +1,51 @@
...
+
+const MenuSection = ({ className, children, id, label, labelTag, ...rest }) => {
+  const sectionId = id || uniqueId('menu-section-');
+  return (
+    <section
+      id={sectionId}
+      className={classNames(css.menuSection, className)}
+      aria-labelledby={`${sectionId}-label`}
+      {...rest}
+      data-test-memu-section
```
JohnC-80: `menu` instead of `memu`

PR #724, opened 2018-11-20
Stcom csv export

aditya-matukumalli:


lib/ExportCsv/ExportCsv.js
```diff
@@ -7,13 +7,14 @@ import Button from '../Button';
...
+    intl: PropTypes.object,
+    options: PropTypes.object,
   }
 
   render() {
-    const { data, excludeKeys } = this.props;
+    const { data, options, intl } = this.props;
     return (
-      <Button onClick={() => { exportToCsv(data, excludeKeys); }}>
+      <Button onClick={() => { exportToCsv(data, options, intl); }}>
```
cherewaty: Is there another option besides passing through an `Intl` prop? could the `exportToCsv()` just live in this repo with the component? Is there a reason for keeping it separate with `stripes-util`?

aditya-matukumalli: @cherewaty Yeah, probably through context, but is there a reason it shouldn't be passed via a prop? I see that the exportCSV has been imported from a couple of places (once from STCOM, and once from STUTIL). I am unsure about what the backstory is to incorporate this into the UTIL repo. It probably could live in STCOM itself. But, the reason I am passing intl is because the util function has to translate the passed in translation keys from different modules, for which it would need the intl object

aditya-matukumalli: @cherewaty @JohnC-80 Adding more context here for the actual use case.

JIRAs: https://issues.folio.org/browse/STUTL-1 and https://issues.folio.org/browse/UIREQ-114

So, the CSV export functionality we currently have in place displays the column headers in the form of nested backend keys, i.e. for example: a record with property/value _item : { 'name': '14 cows` }_ has a header name of _item.name_ in the csv. The requirement form the JIRA is to display this  just as we display the UI labels. ( Item name in this case). Also, another requirement is that we need to have all the header names that are mentioned in scenario 2 in UIREQ-114 to exist in the CSV. We do not have a dynamic way yet to pull all these headers from the backend, and hence we statically create the translation keys array https://github.com/folio-org/ui-requests/pull/170/files#diff-a87e73f2920646e92c51c46fcfa8e7a6R173,  and pass it to the exportCSV util function. 

Once this array is passed to the util function, we can make use of https://www.npmjs.com/package/json2csv#example-3 from **json2csv** module that we are using currently to translate them to the https://github.com/folio-org/stripes-util/blob/master/lib/exportCsv.js#L85 UI labels that we desire as the column headers. So, for this purpose we would need basically a way to have a logic to translate the passed in keys to their corresponding UI labels.

Please let me know if you have better ideas/approaches to handle this case here?

PR #726, opened 2018-11-20
Revert "Stcom csv export"

aditya-matukumalli:
Reverts folio-org/stripes-components#724


aditya-matukumalli:
Reverting, as this needs further discussion

PR #727, opened 2018-11-20
add options prop to ExportCSV, refs STUTL-1

aditya-matukumalli:


JohnC-80:
Let's deprecate `excludeKeys`... pass both it and the new prop 'onlyFields' or ?  to the `exportCsv()` call... 
we can remove that prop in the 5.0.0 release.

aditya-matukumalli:
@JohnC-80 Added in the deprecation warning, additional prop, proptypes and the readme file

lib/ExportCsv/ExportCsv.js
```diff
@@ -8,12 +8,24 @@ class ExportCsv extends React.Component {
...
+    onlyFields: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.string),
+      PropTypes.arrayOf(PropTypes.object),
+    ]),
+  }
+
+  componentDidMount() {
+    const { excludeKeys } = this.props;
+    if (excludeKeys && excludeKeys.length) {
+      console.warn('[DEPRECATED] excludeKeys will be removed soon. Use onlyFields to pass in the necessary fields.');
```
cherewaty: You can use `deprecated` from `prop-types-extra` for a shortcut to this. See `Icon` component for an example.

aditya-matukumalli: @cherewaty Thank you. I made the change.

PR #731, opened 2018-11-26
STCOM-407: Add arrow icons

anastasiaHahanova:
## Purpose
[STCOM-407](https://issues.folio.org/browse/STCOM-407): arrow icons should be added to stripes-components due to one of the requirements in [STSMACOM-133](https://issues.folio.org/browse/STSMACOM-133). They will be used for indicating the state of the searching pane.


cherewaty:
Can we name them `chevron-left` and `chevron-right` instead? I know that's a bit inconsistent with the naming of some current icons, but I eventually want to get them all consistently named as  `${shape}-${modifier}`, or in this case, `${shape}-${direction}`. It makes the groups of icons much easier to find when they're alphabetized by name.

PR #736, opened 2018-11-27
allow PropTypes.node for labels

zburke:
Follow on to #693, expanding prop type validation for various button
labels and other labels to make internationalization easier, using
`<FormattedMessage />` in favor of `formatMessage`.

### DIFF
--- a/lib/IconButton/IconButton.js
+++ b/lib/IconButton/IconButton.js
@@ -82,7 +82,7 @@ const IconButton = React.forwardRef(({
 });
 
 IconButton.propTypes = {
-  ariaLabel: PropTypes.node,
+  ariaLabel: PropTypes.string,
   autoFocus: PropTypes.bool,
   badgeColor: PropTypes.string,
   badgeCount: PropTypes.oneOfType([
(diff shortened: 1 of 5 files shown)


lib/Layer/Layer.js
```diff
@@ -22,7 +22,7 @@ class Layer extends React.Component {
       PropTypes.arrayOf(PropTypes.node),
     ]),
     container: componentOrElement,
-    contentLabel: PropTypes.string,
+    contentLabel: PropTypes.node,
```
cherewaty: Looks like `contentLabel` is used for an `aria-label`; we might have to keep this one a string.

lib/IconButton/IconButton.js
```diff
@@ -105,7 +105,7 @@ IconButton.propTypes = {
   ]),
   style: PropTypes.object,
   tabIndex: PropTypes.string,
-  title: PropTypes.string,
+  title: PropTypes.node,
```
cherewaty: Might need a deprecation instead:
```
title: deprecated(PropTypes.string, 'Use ariaLabel instead')
```

Must've missed it when I did https://github.com/folio-org/stripes-components/pull/703

PR #737, opened 2018-11-27
UIDATIMP-5

VictorSoroka1:
1. Extend `InfoPopover` component to make it more flexible. Add new props `contentClass`, `iconSize` and `allowAnchorClick `.

### DIFF
--- a/lib/InfoPopover/InfoPopover.js
+++ b/lib/InfoPopover/InfoPopover.js
@@ -4,50 +4,67 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { FormattedMessage } from 'react-intl';
+
 import Popover from '../Popover';
 import Button from '../Button';
 import IconButton from '../IconButton';
+
 import css from './InfoPopover.css';
 
 const propTypes = {
+  allowAnchorClick: PropTypes.bool,
   buttonHref: PropTypes.string,
   buttonLabel: PropTypes.node,
   buttonTarget: PropTypes.string,
-  content: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+  content: PropTypes.node,
+  contentClass: PropTypes.string,
+  iconSize: PropTypes.oneOf(['small', 'medium']),
 };
 
 const defaultProps = {
-  buttonLabel: 'Read more',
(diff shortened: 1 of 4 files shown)


lib/Button/Button.js
```diff
@@ -58,13 +58,6 @@ function Button(props) {
     );
   }
 
-  function handleAnchorClick(e) {
-    if (e && !props.allowAnchorClick) e.preventDefault();
```
cherewaty: So the `onClick` will still fire on an href, just not include an `event.preventDefault()` out of the box anymore, is that right?

If so, LGTM 👍

lib/InfoPopover/InfoPopover.js
```diff
@@ -4,50 +4,63 @@
...
 import Popover from '../Popover';
 import Button from '../Button';
 import IconButton from '../IconButton';
+
 import css from './InfoPopover.css';
 
 const propTypes = {
   buttonHref: PropTypes.string,
-  buttonLabel: PropTypes.string,
+  buttonLabel: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
cherewaty: `PropTypes.node` includes `PropTypes.string`, so we don't have to declare both.

lib/Button/Button.js
VictorSoroka1: Yep. Actually, I rethink the approach. I can extend `InfoPopover`to pass additional props `allowAnchorClick` which the will be passed to button thus making it clickable. Thus not making changes in `Button` implementation. Are you meaning that?

zburke: That's what I was thinking, @VictorSoroka1: there is no need to change Button at all here. In the current refactoring, `onClick` will _not_ fire because it's removed in the `omitProps` call beginning on line 61.

lib/InfoPopover/InfoPopover.js
VictorSoroka1: Thanks. Will remove excess one.

JohnC-80:
Thanks @VictorSoroka1 ! The code's looking good. We'll need to update the readme with the updated props.

VictorSoroka1:
@JohnC-80, I will handle it in the PR then. Thank you.

PR #744, opened 2018-11-28
Added tests for Avatar

rasmuswoelk:
I created this component but I never added tests for it.

This PR adds tests for the component.


lib/Avatar/tests/interactor.js
```diff
@@ -0,0 +1,16 @@
...
+ * Avatar interactor
+ */
+
+import {
+  interactor,
+  attribute,
+  isPresent,
+} from '@bigtest/interactor';
+
+export default interactor(class ButtonInteractor {
```
cherewaty: Oops, missed this -> should probably rename to `AvatarInteractor`

rasmuswoelk: Whoops 😆I updated it.

PR #745, opened 2018-11-28
Consistently declare redux form field mappings

cherewaty:
## Purpose
Move toward the end goal of building our form field components agnostic of any form state library.

## Approach
- The `touched` styles were no longer in effect, so no `touched` props were doing anything.
- Always declare `dirty`, `error`, `valid`, and `warning` mappings in the same way.
- Removed `loading` mapping. `asyncValidating` was part of Redux Form, but is not part of React Final Form. Consumers of those components can still use the `loading` prop as needed.

## Next steps
In the very near future we'll be able to remove the `reduxFormField` HOC completely. Instead, hooking up the form state to form field components would be opt-in. This will provide more expressiveness for composing forms in more use cases. It will be more verbose, but consuming developers will have more options according to the needs of their feature.

What that would look like with React Final Form:
Need to use `TextField` without any validation styles?
```
<Field component={TextField} />
```

Have a field that should show an error when it's been focused, even if it hasn't changed?
```
<Field 
  render={({ input, meta }) => (
    <TextField {...input} error={meta.touched && meta.error ? meta.error : ''} />
  )}
/>
```

Have a field that should show an error only when it's been changed?
```
<Field 
  render={({ input, meta }) => (
    <TextField {...input} error={meta.dirty && meta.error ? meta.error : ''} />
  )}
/>
```

Have a field that should show an error, even if it hasn't been changed?
```
<Field 
  render={({ input, meta }) => (
    <TextField {...input} error={meta.error} />
  )}
/>
```

Have a field that needs to show a valid state?
```
<Field 
  render={({ input, meta }) => (
    <TextField {...input} valid={meta.valid} />
  )}
/>
```

Have a field that should only show a valid state when it's been changed?
```
<Field 
  render={({ input, meta }) => (
    <TextField {...input} valid={meta.dirty && meta.valid} />
  )}
/>
```

Have a field that should always show a warning, regardless of its value?
```
<Field 
  render={({ input, meta }) => (
    <TextField {...input} warning="I am a warning." />
  )}
/>
```

### DIFF
--- a/lib/AutoSuggest/AutoSuggest.js
+++ b/lib/AutoSuggest/AutoSuggest.js
@@ -177,7 +177,6 @@ export default reduxFormField(
     dirty: meta.dirty,
     error: (meta.touched && meta.error ? meta.error : ''),
     loading: meta.asyncValidating,
-    touched: meta.touched,
     valid: meta.valid,
     warning: (meta.touched && meta.warning ? meta.warning : ''),
   })
--- a/lib/AutoSuggest/AutoSuggest.js
+++ b/lib/AutoSuggest/AutoSuggest.js
@@ -176,7 +176,6 @@ export default reduxFormField(
   ({ meta }) => ({
     dirty: meta.dirty,
     error: (meta.touched && meta.error ? meta.error : ''),
-    loading: meta.asyncValidating,
     valid: meta.valid,
     warning: (meta.touched && meta.warning ? meta.warning : ''),
   })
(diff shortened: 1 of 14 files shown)


JohnC-80:
I know the render props offer more flexibility, but the verbosity kinda kills me a little. I'd prefer to keep the HOC just to have things snap together in a much nicer way... the `render` prop would still be opt-in for particular cases. Anything really forcing us to get rid of the HOC here?

cherewaty:
We could still keep the HOC around as a default - that's definitely an option. I probably should've dropped that content in a JIRA issue instead of this PR, since our long-term strategy is only tangential to this work.

cherewaty:
@rasmuswoelk I'm not sure what happened to those. It was probably me, in some refactor, but I don't remember intentionally removing them. I have a sequel PR to this that'll DRY things up even more, so it should be easy to put that `touched` back in the future.

PR #747, opened 2018-11-29
Minor refactor and added tests for NavList

rasmuswoelk:
The `handleNavigationClick` method and functionalty was not having any effect anymore so I removed 
it from the component and converted it to a functional component instead.

Added tests for the component as well. One step closer to 100% coverage 😀

### DIFF
--- a/lib/NavList/NavList.js
+++ b/lib/NavList/NavList.js
@@ -1,42 +1,22 @@
+/**
+ * Nav List
+ */
+
 import React from 'react';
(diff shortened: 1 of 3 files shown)


rasmuswoelk:
@cherewaty Yes – this change does not change anything regarding implementation. The removed logic was actually just toggling an "undefined" class on the NavList root element.

The NavListSection handles active styles – but it's also possible to skip that component and wire up the active state manually using the NavListItem component as a direct child of NavList.

PR #748, opened 2018-12-01
A11y clean-up - aria button roles, color contrast issues.

JohnC-80:
# Problem
Axe-reported issues in color contrast issues led me to investigate - based on the Axe project's issues/PR's these reports are often flags for deeper issues. It turns out that the CSS rule `pointer-events: none` can cause textual elements to not be accurately compared with their backgrounds. Buttons and Pane headers/dropdowns have `pointer-events: none` applied to keep inner elements from stealing focus as part of the way that focus styling is set up in the system.  This only appears to be an issue in the case that an inner element of a `<button>` is actually focusable (has a tab index, implicit or otherwise.) Testing on Chrome and Firefox for Windows, non-focusable elements didn't affect the appearance of interaction styles and, more importantly, clicks were not blocked from the interactive elements. Buttons have high potential for contrast issues if the primary colors , so it's best to not give any reason for tooling not to see them.

## Approach

Removed `pointer-events: none` from styles within `Button.css` and `PaneHeader.css`. Modified the focus behavior of pane headers so that instead of the `<h2>` being focused directly, it's mapped as a label to the header's outermost element. This results in a better focus order, since all pane menus will be in the focus path in front of the user.

## Additionally
Button was applying a redundant 'button' role to its rendered `<button>`. Fixed it.

### DIFF
--- a/lib/Button/Button.css
+++ b/lib/Button/Button.css
@@ -7,7 +7,6 @@
 .buttonInner {
   display: flex;
   align-items: center;
-  pointer-events: none;
 }
 
 .button {
--- a/lib/Button/Button.css
+++ b/lib/Button/Button.css
@@ -4,11 +4,6 @@
 * Default styling
 */
 
-.buttonInner {
-  display: flex;
-  align-items: center;
-}
-
 .button {
   composes: interactionStyles from "../sharedStyles/interactionStyles.css";
   padding: 0.125rem var(--gutter);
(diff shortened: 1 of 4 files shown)


lib/Button/Button.css
```diff
@@ -7,7 +7,6 @@
 .buttonInner {
   display: flex;
   align-items: center;
-  pointer-events: none;
```
cherewaty: Is it possible to get rid of the `buttonInner` element completely? That would make styling things like the `SegmentedControl` much simpler. Is there anything it's doing that couldn't be achieved with styles directly on `.button`?

rasmuswoelk: Yup – don't see any problem with removing the buttonInner element 👍

PR #753, opened 2018-12-04
STCOM-416: Fix Keyboard navigation issues

anastasiaHahanova:
## Purpose
Regarding [STCOM-416](https://issues.folio.org/browse/STCOM-416), when a user clicks filter options the whole form shouldn't be highlighted. Only the the current option should be outlined.

## Approach
- eHoldings App uses <RadioButton/> for filter options, so this component should be updated with a few additional attributes.

### DIFF
--- a/lib/RadioButton/RadioButton.js
+++ b/lib/RadioButton/RadioButton.js
@@ -156,7 +156,12 @@ class RadioButton extends React.Component {
     const [, inputCustom] = separateComponentProps(this.props, RadioButton.propTypes);
 
     return (
-      <div className={this.getRootStyle()}>
+      <div
+        className={this.getRootStyle()}
+        role="radio"
+        aria-checked={checked}
+        tabIndex={checked ? 0 : -1}
+      >
         <label htmlFor={this.id} className={this.getLabelStyle()}>
           <input
             id={this.id}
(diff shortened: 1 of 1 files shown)


lib/RadioButton/RadioButton.js
```diff
@@ -156,7 +156,12 @@ class RadioButton extends React.Component {
     const [, inputCustom] = separateComponentProps(this.props, RadioButton.propTypes);
 
     return (
-      <div className={this.getRootStyle()}>
+      <div
+        className={this.getRootStyle()}
+        role="radio"
+        aria-checked={checked}
+        tabIndex="-1"
```
Godlevskyi: Don't the tab index depend on whether it checked?

lib/RadioButton/RadioButton.js
```diff
@@ -156,7 +156,12 @@ class RadioButton extends React.Component {
     const [, inputCustom] = separateComponentProps(this.props, RadioButton.propTypes);
 
     return (
-      <div className={this.getRootStyle()}>
+      <div
+        className={this.getRootStyle()}
+        role="radio"
+        aria-checked={checked}
+        tabIndex={checked ? 0 : -1}
+      >
```
JohnC-80: So it appears this is wrapping an <input> (with implicit tabIndex) with a div role="radio" and an explicit tab index... this results in a double-tab on checked inputs, since.
 ```
<div tabIndex="0"> // explicit tabIndex
  <input /> // implicit tabIndex
</div>
```
This appears as 2 radio buttons to screen readers due to the outer role.
The form on eholdings probably doesn't need its own tabIndex (that's what's causing the style issue, since that element is catching the clicks to the label of the radio button. 
I think it would be best to move the tabIndex, labeling, of the search form to a previous element instead of the entire wrapper so this won't continue to happen with the clickable children.

anastasiaHahanova:
@JohnC-80, thank you for the remark!
I will delete this PR and fix this in e-holding repo within `<RadioButton/>`

PR #761, opened 2018-12-07
STCOM-415 Create ButtonGroup component

cherewaty:
## Purpose
Resolves https://issues.folio.org/browse/STCOM-415.

The `<SegmentedControl>` is unusable for eHoldings' use case because it manages which button is selected. Each of the buttons just need to be links hooked up to router state, not buttons with onClicks and highlighting managed by the component itself.

## Approach
`<ButtonGroup>` relies on the consumer to apply the `primary` button style. This also opens up the possibility of having multiple selections, previously impossible with the `<SegmentedControl>`.

- Relied solely on CSS to apply/remove border radii instead of JS logic.
- Used new `<ButtonGroup>` to simplify `<SegmentedControl>`.
- Fixed `<SegmentedControl>`'s RTL support.
- Left in `SegmentedControl.css` simply to make testing the component easier - it doesn't otherwise need to exist anymore.

### Button Group
screen shot 2018-12-07 at 8 37 24 am

### Segmented Control RTL, before
screen shot 2018-12-07 at 8 34 32 am

### Segmented Control RTL, after
screen shot 2018-12-07 at 8 33 33 am

## Next step
Should we deprecate `<SegmentedControl>`? The approach here is similar to what we did for `<NavList>` in https://github.com/folio-org/stripes-components/pull/747.


lib/ButtonGroup/tests/interactor.js
```diff
@@ -0,0 +1,15 @@
...
+import css from '../ButtonGroup.css';
+
+function tagName(selector) {
+  return computed(function () {
+    return this.$(selector).tagName.toLowerCase();
+  });
+}
+
+export default interactor(class ButtonGroupInteractor {
+   getTagName = tagName(`.${css.buttonGroup}`);
```
wwilsman: This won't be very reusable in the future because this would not work if the button group interactor was scoped to a specific button group.

It would be better to set the ``static defaultScope = `.${css.buttonGroup}`;`` and just use `tagName = tagName();`.

That way this interactor will be reusable in the future. Maybe eventually having a `buttons` collection for interacting with buttons in the group.

lib/ButtonGroup/ButtonGroup.css
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  & > :last-child,
+  & > :last-child::before {
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
+  }
+
+  & > :not(:first-child) {
+    margin-left: -1px;
```
wwilsman: If you have say 5 buttons in a single group, wouldn't this make the right side alignment of the group be 5px short? 

I want to say I can kind of see this in the screenshots at https://github.com/folio-org/ui-eholdings/pull/629, but that could just be my eyes playing tricks on me.

cherewaty: Well I had to go back and quadruple-check this, but no, the group is appropriately filling its container.

lib/ButtonGroup/tests/interactor.js
cherewaty: Ah yeah that's much nicer 👌 Changed.

PR #768, opened 2018-12-13
STCOM-420: Update to Storybook 4

rasmuswoelk:
In this PR I have updated Storybook to version 4 along with all installed addons.

I also made some minor tweaks to some stories.

Everything seems to work fine except for the RTL addon: Currently, it displays the RTL toggle in the bottom of all panes. 

rtl-before

I have submitted a PR to the RTL addon repo that hopefully will get merged soon. It's not an critical issue so I think we should be able to live with it until it's fixed.

https://github.com/unindented/storybook-addon-rtl/pull/3


lib/Checkbox/Checkbox.css
```diff
@@ -89,7 +89,6 @@
 }
 
 .checkboxLabel {
-  min-height: var(--controlHeight);
```
cherewaty: 😕 Was this intentional? If so, we should probably move it to a different PR.

lib/InfoPopover/stories/InfoPopover.stories.js
```diff
@@ -4,6 +4,6 @@ import withReadme from 'storybook-readme/with-readme';
 import readme from '../readme.md';
 import BasicUsage from './BasicUsage';
 
-storiesOf('Info Popover', module)
+storiesOf('InfoPopover', module)
```
cherewaty: I had noticed spacing inconsistencies too in the component story titles. I'll stick to this convention from now on. 👍 We should probably comb through the readmes and make sure they're done that way too.

lib/Checkbox/Checkbox.css
rasmuswoelk: Yes – it was intentional but it's of course not related to the PR.

The variable no longer exists and was only referenced in this component. It was throwing a warning inside the terminal. It doesn't look like there's an alternative variable for this.

Should I move it to another PR?

lib/InfoPopover/stories/InfoPopover.stories.js
rasmuswoelk: Cool. I'll update readme's to match this convention 👍

PR #770, opened 2018-12-13
Removed actionMenuItems prop completely

rasmuswoelk:
**Ticket:** https://issues.folio.org/browse/STCOM-405

**Note:** I did not remove the prop from `<EntrySelector>` since it will be removed in the next release anyway.


cherewaty:
Merging this type of stuff into the `release` branch now 👍

PR #780, opened 2018-12-18
Remove stripes-util dependency

cherewaty:
## Purpose
Moves `exportToCsv()` from `stripes-util` into `stripes-components` itself.

## Next steps
Refactor `ui-requests` to use the `<ExportCsv>` component instead of the `stripes-util` function: https://issues.folio.org/browse/STUTL-3


JohnC-80:
Think we should go ahead and export the 'exportToCSV' function at top level? The instant someone has a ui design for this that's not a button., my eyes will roll so hard they will exit my skull.

cherewaty:
@JohnC-80 😂 I'd like to not export it, and instead encourage a component-based API that doesn't involve a button when that inevitably comes up.

PR #785, opened 2018-12-19
STCOM-422: Add Storybook example for <LayoutGrid>

rasmuswoelk:
In this PR I have added a story for the LayoutGrid-component

I added a couple of examples to showcase some of its capabilities but since it's directly exporting the `react-flexbox-grid`-package I have linked to their docs for more examples and information.

**Ticket:** https://issues.folio.org/browse/STCOM-422

image


rasmuswoelk:
@cherewaty 👍 

Aren't we moving towards a grid system that's not based on window width but on pane width instead at some point? For when we get resizable panes etc.

PR #792, opened 2018-12-27
STCOM-439: Add multi selection filter

Yurii-Danylenko:
https://issues.folio.org/browse/STCOM-439

## Purpose
Create a reusable multiselection filter, based on MultiSelection component, to be used in SearchAndSort side panel.

## Approach
- create `onChangeHandler` which will raise filter name and selected values after a filter change.
- add a couple of props
- add storybook
- add tests only specific for this component

### DIFF
--- a/lib/MultiSelectionFilter/MultiSelectionFilter.js
+++ b/lib/MultiSelectionFilter/MultiSelectionFilter.js
@@ -1,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import MultiSelection from '../MultiSelection';
+
+export default class MultiSelectionFilter extends React.Component {
+  static propTypes = {
+    name: PropTypes.string.isRequired,
+    onChange: PropTypes.func.isRequired,
+    options: PropTypes.arrayOf(PropTypes.shape({
+      label: PropTypes.node,
+      value: PropTypes.any,
+    })).isRequired,
+    selectedValues: PropTypes.arrayOf(PropTypes.string),
+  };
+
+  static defaultProps = {
+    selectedValues: [],
+  }
+
+  onChangeHandler = (selectedOptions) => {
+    const {
+      name,
+      onChange,
+    } = this.props;
+
+    onChange({
+      name,
+      values: selectedOptions.map((option) => option.value),
+    });
+  };
+
+  getSelectedOptions() {
+    const {
+      selectedValues,
+      options,
+    } = this.props;
+
+    return selectedValues
+      .map((curValue) => options.find((curAvailableValue) => curAvailableValue.value === curValue));
+  }
+
+  render() {
+    const {
+      options,
+    } = this.props;
+
+    return (
+      <MultiSelection
+        {...this.props}
+        dataOptions={options}
+        value={this.getSelectedOptions()}
+        onChange={this.onChangeHandler}
+      />
+    );
+  }
+}
--- a/lib/MultiSelectionFilter/MultiSelectionFilter.stories.js
+++ b/lib/MultiSelectionFilter/MultiSelectionFilter.stories.js
@@ -1,0 +1,86 @@
+import React from 'react';
+import { action } from '@storybook/addon-actions';
+import withReadme from 'storybook-readme/with-readme';
+import { storiesOf } from '@storybook/react';
+import readme from './readme.md';
+import MultiSelectionFilter from './MultiSelectionFilter';
+
+const options = [
+  { value: 'en', label: 'English' },
+  { value: 'fr', label: 'French' },
+  { value: 'ge', label: 'German' },
+  { value: 'it', label: 'Italian' },
+];
+
+const styles = { maxWidth: '300px' };
+
+
+const renderAddTag = ({ filterValue, exactMatch }) => {
+  if (exactMatch) {
+    return null;
+  } else {
+    return (
+      <div>
+        Add tag for
+        <strong>
+          &quot;
+          {`${filterValue}`}
+          &quot;
+        </strong>
+      </div>
+    );
+  }
+};
+
+storiesOf('MultiSelectionFilter', module)
+  .addDecorator(withReadme(readme))
+  .add('Basic Usage', () => (
+    <div style={styles}>
+      <MultiSelectionFilter
+        name="language"
+        options={options}
+        onChange={action}
+      />
+    </div>
+  ))
+  .add('With label', () => (
+    <div style={styles}>
+      <MultiSelectionFilter
+        name="language"
+        options={options}
+        onChange={action}
+        label="Languages"
+      />
+    </div>
+  ))
+  .add('With placeholder', () => (
+    <div style={styles}>
+      <MultiSelectionFilter
+        name="language"
+        options={options}
+        onChange={action}
+        placeholder="Choose a language!"
+      />
+    </div>
+  ))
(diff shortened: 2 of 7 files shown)


JohnC-80:
Would it be possible to consolidate the needs of these filter components to a single component?
Each needs a `name` and an `onChange` prop that's necessary to update `SearchAndSort`'s filters...

Would an API like this be possible:
```
<FilterField
  name="country"
  component={MultiSelection}
  onFilterChange={onFilterChange}
  {/* other MultiSelection props */}
/>
```

Thank you so much for your work with this, @Yurii-Danylenko !

cherewaty:
+1 on John's `<FilterField>` idea, if that's possible.

Yurii-Danylenko:
> Would it be possible to consolidate the needs of these filter components to a single component?
> Each needs a `name` and an `onChange` prop that's necessary to update `SearchAndSort`'s filters...
> 
> Would an API like this be possible:
> 
> ```
> <FilterField
>   name="country"
>   component={MultiSelection}
>   onFilterChange={onFilterChange}
>   {/* other MultiSelection props */}
> />
> ```
> Thank you so much for your work with this, @Yurii-Danylenko !
Thanks for the feedback! I have checked the interfaces of the form components(Select, Selection, MultiSelection, Checkbox, RadioButton) and I'm afraid it's very hard to go with the suggested approach. The interfaces of the form components are different. They are working with single and multiple values, returns different types of data after the change action(event or selected options), range filters will get another type of props(from, to) and multiple names. But I think it is possible to create a general filter component for selects(Select, Selection, MultiSelection). What do you think @JohnC-80 and @cherewaty?

JohnC-80:
I suppose it's all about conforming to what SearchAndSort will ultimately agree with. It would be nice to avoid limiting developers to what they can use in specific situations. The fewer the components, the better.
Maximum versatility in mind, what would you guys think about a prop that would accept a function to shape the Component's value in a way that might be accepted by SearchAndSort? A "filter adapter" of sorts... This would keep from having to do too much type detection within the components themselves and make it adaptable to any value/type a component can produce. Ex:

```
arrayToBools = (val) => {
//...use val to return {'label': true, ...} or whatever SearchAndSort will accept.
}

<FilterField
  adapter={this.arrayToBools}
  onFilterChange={handleFilterChange}
  component={MultiSelection}
/>
```

JohnC-80:
Ok. With the component approach, I suppose any outlying components not covered by the components could implement their own adaptation within the `<Filters>` component if needed. 

Another conversation point: 
Due to these components being hardwired to the way that SearchAndSort thinks, would `stripes-smart-components` be the better place for them? 

Additionally, I think it would be best for the *Filter versions to share the same prop-names as the presentational components they wrap - their documentation could simply include the extra bits added for the `<*Filter>` component and simply link to the documentation of the presentational component within stripes-components. If we do decide they should stay here, I think giving them a name prefix so that they alphabetize together would be nice.

I would love to get more input on these from other module developers as well @mkuklis, @zburke @doytch

JohnC-80:
Yes. If I'm not mistaken, they will all ultimately have to output filter values in a shape of `{ name: [array]}` or `{name: bool}` for SearchAndSort - is that correct?

Yurii-Danylenko:
> Yes. If I'm not mistaken, they will all ultimately have to output filter values in a shape of `{ name: [array]}` or `{name: bool}` for SearchAndSort - is that correct?

They will raise data in the format: `{ name: 'filter name',  values: ['filterValue1', 'filterValue2']}`. So we are not wired to SearchAndSort. It's just like a common contract for filters to simplify handling of the changes in `Filters` container(not common), which will be created in each UI module. Talking about determining the location to place filter components, I want to say that all these components assuming to be dumb. So as for me, it's better to place them in `stripes-components`. Talking about `Filters`, this container can be a smart component. It should render reusable filters(CheckboxFilter, SelectFilter, ...), handle filter changes according to contracts. Also, it can fetch data, store and provide it for filters.

JohnC-80:
I don't mean to sound like an obstructionist, but - There's another important distinction: the *Filter components might be merely presentational by definition - but - they only save boilerplate that's *specifically* related to usage of the SearchAndSort component. There are not other use-cases within the system for conversion to the shape of object that we've discussed.

If we add components with such specific use-cases to this library, amid all of the others generic components intended for wider use, it can eventually become more difficult to find the generic components necessary to fulfill requirements.

These *Filter components are used with SearchAndSort - it seems fitting that they would also import with it and be documented alongside it as well.

doytch:
First off, nice job! Clean...tested...good stuff. Specifically, I like and appreciate the API for `asyncFiltering`.

**Generic `FilterField` vs specific `FilterX` components**
TL; DR, I see benefits to both approaches and won't be bent out of shape about either way. But I do have some thoughts and I hope they're not too scatterbrained.

The generic approach conforms to the style we already use with `redux-form` (with `Field` and its `component` prop) and the approach using specific components helps with brevity.

The one thing that I'm definitely anticipating is having to use custom filter components in some places. So I really want that ability to be preserved. I think it can be done with both approaches, but I do see how the step from, say, using a `MultiSelectionFilter` to `FilterField component={MyCustomMultiSelect}` isn't as neat as just changing the `component` prop.

Additionally, I do agree with @JohnC-80 that perhaps adhering to the child component prop names might make sense. So in this case, rather than having props of `selectedValues` and `options` we'd have `value` and `dataOptions`. It makes it easier to remember what the prop names are if we stick to conventions across Stripes. 

Consequently, if that's done I really do wonder how much of the work that `MultiSelectionFilter` currently does needs to be done in it rather than the parent component. It seems like its main job is just transferring props and translating between the flat array of selected values and the array of option objects. And it's only really doing that because it looks like we're _assuming_ that the parent would actively _prefer_ to provide `values` and get `onChange` parameters as a flat array of values. I'm not sure whether that's necessarily the case. I know in my uses I wouldn't really care whether I'm getting/providing flat arrays or arrays of objects.

So I'm kind of curious as to what this component would look like if we just used the same prop names and just forced the parent component to provide `selectedValues` (or `value`) in the same form as the `dataOptions`/`options`. Could we even just pass `MultiSelection` directly to SAS?

(Now, I do see where a `FilterCheckboxes` or `FilterRadio` would be pretty helpful rather than having to render everything out manually, but at least for MultiSelects it doesn't seem like we're doing too too much here over just rendering out the MultiSelect directly.)

doytch:
I left out one other benefit to keeping the prop names as-is: it probably matches the prop names used for other `XFilter` components better. I'm looking at the `CheckboxFilter` PR and that component also has `selectedValues` and `options` props and in that scenario, it does make more sense to use those names rather than then forcing it to use `value` and `dataOptions` for consistency with `MultiSelectionFilter`. So....I'm even _more_ on the fence now. 😆

JohnC-80:
@doytch , @zburke , @mkuklis if it's helpful, this is @Yurii-Danylenko 's branch where you can see how MultiSelectionFilter can be used: https://github.com/folio-org/ui-inventory/blob/yd/multiselect-filter/src/filters/filters.js

The `<Filters>` component will be unique from one module to another - it's a container to receive props passed in from SearchAndSort: You can see that connection here:
https://github.com/folio-org/ui-inventory/blob/yd/multiselect-filter/src/Instances.js#L426

This is an awesome step forward in the flexibility of SearchAndSort - Right now I think it's just down to sorting out where the pieces best belong.

PR #804, opened 2019-01-04
STCOM-440: Add checkbox filter

Yurii-Danylenko:
https://issues.folio.org/browse/STCOM-440

## Purpose
Create a reusable checkbox filter, based on Checkbox component, to be used in SearchAndSort side panel. Presents a set of checkboxes working like one filter.

## Approach
- create `onChangeHandler` which will raise filter name and selected values after a filter change.
- add a couple of props
- add storybook
- add tests only specific for this component
- add Sinon to test the parameters passed to the onChange callback from the filter

### DIFF
--- a/lib/CheckboxFilter/CheckboxFilter.js
+++ b/lib/CheckboxFilter/CheckboxFilter.js
@@ -1,0 +1,63 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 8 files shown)


Yurii-Danylenko:
Is moved to smart components

Yurii-Danylenko:
Is moved to smart components

PR #815, opened 2019-01-14
STCOM-441: Update padding/margins

rasmuswoelk:
**Ticket:** https://issues.folio.org/browse/STCOM-441

## Changes
In this PR I updated padding/margins of some components to static pixel values. While doing so I did some minor adjustments for various components to ensure proper alignment when composing UI's containing different types of components. Screenshots below.

The goal was to transform the design on desktop to have pixel density in mind on the browsers default font-size (medium) while still preserving a proper layout when a user has set a browser font-size of large or extra large.

I tested this work with the default font-size of the system and also large and extra large on various screens in the FOLIO app and it seems to work well. The layout does definitely not get prettier when the font-size is set to extra large but it's still decent and useable which is the most important thing.

## New variables
I added some new variables. I chose a naming that I think makes sense but if any of you guys have a suggestion for a better naming then please let me know.

**Min-size variables**
To ensure a min-height of 24px on desktop for components like buttons, inputs, selects etc. And later on we can adjust the same components to use the touch min-size variable on mobile screens (if the user chooses to set their application to touch-mode.)

`--control-min-size-desktop: 24px;`
`--control-min-size-touch: 44px;`

**Static gutter variables**
The default static gutter will be 15px which means that it won't scale when the font-size is adjusted which helps preserve space for small screens with a large font-size.

The static gutter is primarily utilised in thirds so I decided to make two extra variables instead of having to use calc() everywhere.

`--gutter-static: 15px;`
`--gutter-static-one-third: calc(var(--gutter-static) / 3);`
`--gutter-static-two-thirds: calc(var(--gutter-static-one-third) * 2);`

Again, I'm not too sure about the naming here. It can of course easily be changed.

# Comparison
I uploaded my changes to this test site if you want to check it out without pulling these changes locally: http://folio-styling.testing.samhaeng.com

## Padding / alignment test
I made a little test example while working on these changes. Below you can see the before/after alignment. The red line indicates where the default padding (15px) within a Pane, Modal, Dropdown etc. stops and the content begins.

**Note:** There are still some work to be done here but it's atleast more consistent now. This should make it easier to compose nice looking, well-aligned UI's.

### Before
https://www.dropbox.com/s/oz72n0lt5iovhp4/before.mp4?dl=0

### After
https://www.dropbox.com/s/ry7nvwpddb33emw/after.mp4?dl=0

### Before
screenshot 2019-01-14 12 32 01

### After
screenshot 2019-01-14 12 28 33

### Before
screenshot 2019-01-14 12 37 16

### After
screenshot 2019-01-14 12 39 40

### Before
screenshot 2019-01-14 12 32 46

### After
screenshot 2019-01-14 12 30 25

## Large font-size
screenshot 2019-01-14 12 34 59

## Extra large font-size
screenshot 2019-01-14 12 34 44

# Next up
- Create and add PaneHeaderIconButton-component that will be adjusted specifically for PaneHeader's to ensure 44px touch face
- Adjust MainNav padding in stripes-core and app list dropdown padding
- Apply touch min size for some components when "touch-mode" is activated (this will require a user setting since we can't properly determine whether or not a user is on a touch device automatically)


rasmuswoelk:
@cherewaty Thanks for letting me know. I was adjusting the design according to the designs/prototypes but was not aware of this. I'll discuss with Filip possible re-adding the indent (or if there's something else we can do to improve the experience for low-vision users).

PR #820, opened 2019-01-18
UIDATIMP-105: Add EndOfList component below the MLC component.

VictorSoroka1:
## Purpose

To be able to add `EndOfList` component when the user indeed have all the items scrolled to the bottom. The component is shows when the `totalCount` property is provided and greater or equal to the `contentData.length` value.


JohnC-80:
One thing I notice that will come up later is that the icon doesn't remain visible in all cases. I'm all for the rest of the clean-up within the code.

Full view...fine
image
Details open - icon gets hidden.
image

PR #823, opened 2019-01-21
Make <Button> align-prop great again

rasmuswoelk:
While working on something else I needed a way to left align the content of a `<Button>`. I looked in the readme and found that there already was a prop for this – great!

 _The problem:_ the align-prop wasn't really doing anything.

In this PR I added alignment functionality for the `<Button>`-component. I also removed a reference for `css.floatEnd` which aren't defined in `Button.css` anymore and probably has not been for a while.

### DIFF
--- a/lib/Button/Button.css
+++ b/lib/Button/Button.css
@@ -263,3 +263,19 @@
     outline: none;
   }
 }
+
+/**
+ * Alignment
+ */
+
+.align-start {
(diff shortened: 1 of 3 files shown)


zburke:
This is the greatest PR title of all time. It has words in it, some really great words, the best words, if I'm being honest.

PR #826, opened 2019-01-22
STCOM-411: Setup deprecation message for <AppIcon>

MaksymDryha:
### Purpose
`<AppIcon>` is the last place in `stripes-components` which relies on `withStripes`. That's why we want to move it to `stripes-core`.
### Approach 
* `<AppIcon>` was refactored to a class component in order to be able to show a deprecation message in its constructor
* `<PaneHeader>` component was changed to use `stripes-core`'s `<AppIcon>`

This PR can only be merged after [this one](https://github.com/folio-org/stripes-core/pull/567) (since currently there is no `AppIcon` in `stripes-core`, which can also make tests fail)
[JIRA issue link](https://issues.folio.org/browse/STCOM-411)

### DIFF
--- a/lib/PaneHeader/PaneHeader.js
+++ b/lib/PaneHeader/PaneHeader.js
@@ -3,12 +3,14 @@ import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-import css from './PaneHeader.css';
+import AppIcon from '@folio/stripes-core/src/components/AppIcon';
+
 import IconButton from '../IconButton';
 import { Dropdown } from '../Dropdown';
 import DropdownMenu from '../DropdownMenu';
 import Icon from '../Icon';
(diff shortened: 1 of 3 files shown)


lib/PaneHeader/PaneHeader.js
```diff
@@ -3,12 +3,14 @@ import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-import css from './PaneHeader.css';
+import AppIcon from '@folio/stripes-core/src/components/AppIcon';
```
JohnC-80: Let's hold off on importing this from stripes-core just yet. We should not have to depend on `stripes-core` in order to run the `stripes-components` tests. Once we have the static image/icon node logic broken out to its own presentational component, we can use that here.

MaksymDryha: Change added with the last commit

PR #831, opened 2019-01-25
STCOM-463: create DropdownButton component

MaksymDryha:
This is a simple button component which displays a caret icon in order to let users know that the button opens a dropdown. The `open` prop is used to render the appropriate icon and reflect the dropdown's state.
### Usage:
```
<Dropdown
  open={this.state.isDropdownOpen}
  onToggle={this.onDropdownToggle}
>
  <DropdownButton data-role="toggle">
    Toggle dropdown
  </DropdownButton>
  <DropdownMenu data-role="menu">
    <span>This is our dropdown</span>
  </DropdownMenu>
</Dropdown>
```

We want to be able to use it with `<UncontrolledDropdown>`, which manages the dropdown's state by itself, so we just don't have the possibility to pass the `open` prop to the button directly. For this purpose, the `<Dropdown>` component was modified to pass down its `open` prop to the button, so there is no need to do it manually


JohnC-80:
Code looks good. @MaksymDryha!  - could you make a brief readme for this? You don't have to list out all of the props, just the commonly used ones, as you have used in your description - and you can mention that it also accepts the props of Dropdown and DropdownMenu.

lib/DropdownButton/DropdownButton.js
```diff
@@ -0,0 +1,190 @@
...
+      'bsRole',
+      'minWidth',
+      'onClose',
+      'onSelect',
+      'overrideStyle',
+      'rootCloseEvent',
+      'width',
+    ];
+
+    const drowdownProps = omit(this.props, unnededProps);
```
rasmuswoelk: Could we instead just pass a "dropdownProps"–prop to the component? The same with the props for the button (buttonProps).

That way we don't have to update this list of props when the props change in the components used to compose this component.

What do you think, @JohnC-80?

lib/DropdownButton/DropdownButton.js
```diff
@@ -0,0 +1,190 @@
...
+      buttonClass,
+    );
+
+    return (
+      <Button
+        data-role="toggle"
+        buttonClass={buttonClasses}
+        {...otherButtonProps}
+      >
+        <span className={css.dropdownButton__content}>
```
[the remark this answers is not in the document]
MaksymDryha: It's a nice improvement, thanks!

JohnC-80:
After discussing a bit with @MaksymDryha - the simplest solution I believe would be for `<DropdownButton>` to be just that without worrying about forwarding props and supporting all of those things that the internal components support...
It could simply wrap a `<Button>` and accept the `open` prop to render the caret appropriately...
Then the abbreviated API becomes something like
```
<Dropdown open={openVar}>
  <DropdownButton open={openVar}>
  <DropdownMenu>
</Dropdown>
```

rasmuswoelk:
@JohnC-80 

That sounds like the right solution – much cleaner. @MaksymDryha and I talked about this on Slack and noticed that we are going to have a little problem with the UncontrolledDropdown because the open-state aren't exposed.

This could be solved by utilising the render-props pattern to expose the open-state. However, I'm not totally sure if this approach fits with the way the Dropdown/UncontrolledDropdown components are build.

lib/DropdownButton/DropdownButton.js
```diff
@@ -0,0 +1,41 @@
...
+import Icon from '../Icon';
+
+
+const DropdownButton = (props) => {
+  const {
+    children: buttonContent,
+    ...buttonProps
+  } = props;
+
+  const iconType = props.open
```
Godlevskyi: Please add open prop to propTypes

lib/DropdownButton/DropdownButton.stories.js
```diff
@@ -0,0 +1,22 @@
...
+import withReadme from 'storybook-readme/with-readme';
+import DropdownButton from './DropdownButton';
+import stories from './stories';
+import readme from './readme.md';
+
+
+storiesOf('DropdownButton', module)
+  .addDecorator(withReadme(readme))
+  .add('Open', () => (
+    <DropdownButton open>
```
rasmuswoelk: Can we remove these examples and just have one single example called "BasicUsage". I don't think it's necessary to show a specific open/closed example. (We should probably remove these examples from the `<Dropdown>`-examples as well but that's for another ticket)

The single "Basic Usage" example could display both examples (Dropdown/UncontrolledDropdown) as well as maybe an example of a different button style.

lib/Dropdown/Dropdown.js
```diff
@@ -136,11 +136,12 @@ class Dropdown extends React.Component {
         case 'toggle': {
           return cloneElement(child, {
             'ref': isStatefulComponent(child) ? (c) => { this._toggle = c; } : undefined,
-            disabled,
             'onClick': this.handleToggle,
             'onKeyDown': this.handleKeyDown,
             'aria-expanded': open,
             'aria-haspopup': true,
+            open,
```
rasmuswoelk: We already have the `aria-expanded`-prop just above – couldn't we just listen for this prop inside the DropdownButton-component instead of adding the open-prop?

lib/DropdownButton/DropdownButton.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Button from '../Button';
+import Icon from '../Icon';
+
+
+const DropdownButton = (props) => {
```
rasmuswoelk: To make it a little bit cleaner we could do this:

const DropdownButton = ({ children, ...buttonProps }) => { .... }

MaksymDryha:
> I added a few suggestions for changes.

Thanks, @rasmuswoelk! Changes added with the last commits

PR #841, opened 2019-01-30
use static column label for sorting

zburke:
Use columns' static labels, i.e. the keys from `props.columnMapping`,
rather than their translated labels, i.e. the values from
`props.columnMapping`. This means the sort parameter in the querystring
will remain constant regardless of locale.

Refs [STSMACOM-93](https://issues.folio.org/browse/STSMACOM-93), [UIU-479](https://issues.folio.org/browse/UIU-479), [UIU-864](https://issues.folio.org/browse/UIU-864)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-components
 
+## 5.1.0 (IN PROGRESS)
+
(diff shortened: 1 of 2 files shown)


zburke:
Closing in light of #843. This should be easy to re-write in terms of that refactoring, which is much more involved.

PR #843, opened 2019-01-30
STCOM-365 MCL ARIA Roles

JohnC-80:
# Problem(s)
MCL announces that it's a `<table>` but doesn't behave like a table for screen readers. This was partly due to an incorrect `aria-label` on the component's main container.

Additionally, any sorting was left completely unannouced.

# Approach
There are a few "roles" required to make a 'table' or 'grid' work... `grid`,`columnheader`, `row`, `gridcell` - for instances where fewer rows are presented than actually available, `aria-rowcount` exists. Columns are automatically recognized and will be announced when the component is entered. Both JAWS and NVDA announce something like "Grid with 202 rows and 6 columns".

For sorting, the component takes the `sortOrder`, `sortColumn`, and `sortDirection` props into account when building the headers, so those will be announced for screen-reader users as well.

As for the incorrect label, I simply discontinued use and removed it from translation files (so it should be removed from others shortly.)

This will have a corresponding PR in stripes-smart-components to update any `rowFormatter`s applied there (replacing the `listitem` role with `row`, and applying an `aria-rowindex` prop.)
Stripes-smart-components PR: https://github.com/folio-org/stripes-smart-components/pull/419

Test-writers, fear not (much) -  the `aria-label` rendered per row which should probably not be used has had its name changed to `data-label`... but ideally, tests can be refactored not to depend on that at all.

Running axe in the Users view with the MCL displayed gets a clean bill of health: 
image


zburke:
Playing with selectors after the merge, I'm seeing some funny things. e.g. on the Users search page with a single filter checked so we get a long list of results: 

    $('#list-users div[role="row"]:nth-child(1)')

selects the header row but 

    $('#list-users div[role="row"]:nth-child(2)')

selects the _second_ data row, i.e. there is not a way to grab the first row of data with an `nth-child()` selector, which seems wrong. Selecting with 

    $('#list-users div[role="row"][aria-rowindex="2"]')

works as expected, providing access to the first data row, so it _is_ selectable, just not by `:nth-child()`.

PR #849, opened 2019-02-01
ClassNames in MCL use 'mcl' prefix

JohnC-80:
# Problem

Until we're able to import interactors into ui-modules, we have to stick to selecting around in the rendered DOM in order to build tests.

Testing with MCL currently has some challenges due to indistinct classnames making it difficult to select particular elements to assert on. This PR remedies that by prefixing all of the major class names with "mcl".

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.css
+++ b/lib/MultiColumnList/MCLRenderer.css
@@ -6,11 +6,11 @@
   }
 }
 
-.heightSpacer {
+.mclHeightSpacer {
   position: relative;
 }
 
-.rowContainer {
+.mclRowContainer {
   position: absolute;
   min-width: 100%;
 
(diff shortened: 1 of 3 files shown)


JohnC-80:
@doytch - Yes, we really want to make the component interactors available, but we need to figure out the best way to exclude them from the primary js bundle.

doytch:
> @doytch - Yes, we really want to make the component interactors available, but we need to figure out the best way to exclude them from the primary js bundle.

Is that a solvable problem right now or does Webpack/etc lack the necessary tools to make it possible? Just curious and really not trying to rush things along.

JohnC-80:
It's one of those cases where JS-land hasn't landed on a specific 'best way to do this' - there's a couple of potential options that range from giving interactors and other test-related code an exclusive import path (not through index.js) to perhaps something more elegant in our webpack configs that will split the test-related code out to a separate bundle (so we could export through index.js without much worry so long as it's following some convention.)

PR #851, opened 2019-02-05
STCOM-469: Add asterisk for required fields

rasmuswoelk:
**Ticket:** https://issues.folio.org/browse/STCOM-469

In this PR I added an asterisk in the label-area of form components that are set as "required"

## Changes
- Added a re-usable `<Label>` component
- Replaced `<label>`-elements within form components with the new `<Label>`-component
- Added some examples of required form fields in Storybook
- Moved the label-styling out of sharedStyles/form.css into the Label-component folder

## Next up
The asterisk will automatically be applied to any form components that has required = true.

As mentioned in the ticket we now need to:
- Remove any asterisk's in translation files
- Remove any hardcoded asterisk's

## Screenshots
image
image
image
image
image


doytch:
The amount of times I see a PR in my notifications from @rasmuswoelk and am like "OMG yes finally I don't have to do <stupid thing> anymore" is significant.

Thanks man!

PR #858, opened 2019-02-06
STCOM-370: Add popper component

Yurii-Danylenko:
https://issues.folio.org/browse/STCOM-370

## Purpose
Create a component with a basic mechanism for popups.

## Approach
1) Create a component using 3-rd party library Popper.js
2) Create a storybook
3) Write tests

### DIFF
--- a/lib/Popper/Popper.js
+++ b/lib/Popper/Popper.js
@@ -1,0 +1,112 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import PopperJS from 'popper.js';
+
+const AVAILABLE_PLACEMENTS = [
+  'bottom',
+  'top',
+  'left',
+  'right',
+  'top-start',
+  'top-end',
+  'bottom-start',
+  'bottom-end',
+  'left-start',
+  'left-end',
+  'right-start',
+  'right-end',
+];
+
+const [DEFAULT_PLACEMENT] = AVAILABLE_PLACEMENTS;
+
+export default class Popper extends React.Component {
+  static propTypes = {
+    anchorRef: PropTypes.node.isRequired,
+    children: PropTypes.node.isRequired,
+    isOpen: PropTypes.bool,
+    modifiers: PropTypes.object,
+    onUpdate: PropTypes.func,
+    placement: PropTypes.oneOf(AVAILABLE_PLACEMENTS),
+    portal: PropTypes.element,
+  }
+
+  static defaultProps = {
+    placement: DEFAULT_PLACEMENT,
+    onUpdate: () => {},
+  }
+
+  componentDidMount() {
+    if (!this.props.isOpen) return;
+
+    this.createPopperInstance();
+  }
+
+  componentDidUpdate() {
+    this.updatePopperInstance();
+  }
+
+  componentWillUnmount() {
+    this.destroyPopperInstance();
+  }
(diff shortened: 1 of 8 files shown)


lib/Popper/Popper.js
```diff
@@ -0,0 +1,142 @@
...
+
+  componentDidUpdate(prevProps) {
+    const propsChanged = this.props.isOpen !== prevProps.isOpen
+      || this.props.placement !== prevProps.placement
+      || this.props.children !== prevProps.children
+      || this.props.portal !== prevProps.portal
+      || this.props.hideOnOverflow !== prevProps.hideOnOverflow
+      || this.props.onUpdate !== prevProps.onUpdate
+      || this.props.renderAnchor !== prevProps.renderAnchor;
+
```
JohnC-80: Could this be a PureComponent rather than use a conditional and simply call `updatePopperInstance()` when it updates?

lib/Popper/Popper.js
```diff
@@ -0,0 +1,142 @@
...
+  render() {
+    const {
+      isOpen,
+      renderAnchor,
+    } = this.props;
+
+    return (
+      <div style={this.getStyles()}>
+        {renderAnchor(this.anchorRef)}
```
JohnC-80: Instead of requiring a function for the `renderAnchor` prop, could it be made to accept a node, and wrap the node in its own element for it to create a ref to? similar to how it treats the overlay?

lib/Popper/Popper.js
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    this.setState({
+      instance: newInstance,
+    });
+  }
+
+  destroyPopperInstance() {
+    if (this.state.popperInstance) {
+      this.state.instance.destroy();
```
MaksymDryha: Does it work? Here you you refer to `this.state.popperInstance` and then to `this.state.instance`

lib/Popper/Popper.js
Yurii-Danylenko: Wrapping anchor component or node inside Popper will cause style issues, as you mentioned, which will affect on near elements. I think we should avoid side effects. Are there any issues could be with passing `anchor` as render prop?

lib/Popper/Popper.js
```diff
@@ -0,0 +1,137 @@
...
+  'left-start',
+  'left-end',
+  'right-start',
+  'right-end',
+];
+
+export default class Popper extends React.Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    hideOnOverflow: PropTypes.bool,
```
Godlevskyi: please add it to default props

lib/Popper/Popper.js
JohnC-80: I think the only issue that could happen there is if a ref to the anchor is also needed externally...I suppose then you could pass Popper's ref to a function to set both up... but there might be a cleaner way....
What do you think about simply passing in an existing ref or a ref.current or anything that equates a DOM node? This could work since Popper.js essentially operates on DOM nodes... 

```
this.anchorRef = React.createRef();
<Button ref={this.anchorRef} > Menu </Button>
<Popper anchor={this.anchorRef.current}>
  <span>Popper Content</span>
</Popper>
```

Yurii-Danylenko: Done

PR #862, opened 2019-02-07
Fix group filter checkbox id

mkuklis:
The previous id was similar to `clickable-filter-active-Include inactive users` which is invalid. It's impossible to reference it via dom queries. This PR changes:

`clickable-filter-active-Include inactive users`

to:

`clickable-filter-active-include-inactive-users`

### DIFF
--- a/lib/FilterGroups/FilterGroups.js
+++ b/lib/FilterGroups/FilterGroups.js
@@ -1,3 +1,4 @@
+import { kebabCase } from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import FilterControlGroup from '../FilterControlGroup';
@@ -16,7 +17,7 @@ const FilterCheckbox = (props) => {
   const fullName = `${groupName}.${name}`;
 
   return (<Checkbox
-    id={`clickable-filter-${fullName.replace('.', '-')}`}
+    id={`clickable-filter-${kebabCase(fullName)}`}
     label={name}
     name={fullName}
     checked={checked}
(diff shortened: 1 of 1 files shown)


mkuklis:
@zburke let's keep this one open for a bit then so we don't cause more issues. I was hoping since the id was invalid it was actually impossible to use it anywhere.

zburke:
Oh, that's a fair point. If the id is invalid, how could it be used in any tests?

lib/FilterGroups/FilterGroups.js
```diff
@@ -1,3 +1,4 @@
+import { kebabCase } from 'lodash';
```
JohnC-80: nitpicky... this vs `import kebabCase from 'lodash/kebabCase'`?

rasmuswoelk: I prefer `import kebabCase from 'lodash/kebabCase'` to keep bundle sizes small. This, of course, requires that everyone imports this way.. otherwise it doesn't really matter I guess.

And I don't know if there's some webpack magic that handles this type of thing anyway?

Someone did a benchmark on this: https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark

rasmuswoelk: TL;DR: the most effective import is one-by-one.

mkuklis: @rasmuswoelk the chosen style doesn't influence the performance if we have https://github.com/lodash/lodash-webpack-plugin or https://github.com/lodash/babel-plugin-lodash in place (also referenced by the link you shared). Perhaps it would be valuable to add it to to our setup? @ttamj is this something you could help us with?

ttamj: We already have a JIRA for adding the plugins: STCOR-285

mkuklis: @JohnC-80 I just changed it to align it with the rest of the stripes-components lodash imports.

PR #867, opened 2019-02-11
Fix <Pane> border on RTL direction

rasmuswoelk:
I noticed a missing border when on RTL direction and `<Pane>`'s are stacked. This PR's fixes this.

**Note:** I have noticed other areas of the app where missing borders on panes is an issue. After some investigation I found that this is primarily happening when one or more elements within a `<Paneset>` isn't a `<Pane>`. E.g. when on route `/settings/tags/general`:

image

# Screenshots
Notice the border to the right on the first pane.

## Before
screenshot 2019-02-09 at 15 59 32

## After
screenshot 2019-02-09 at 15 56 54


zburke:
Is there a secondary bug/issue here where stripes-smart-components/lib/ConfigManager's form should wrap itself in a `Pane`? Currently it returns
```
    <form id='config-form'>
        <Pane ...>
            {children}
        </Pane>
    </form>
```
I don't think we can just invert the form and `Pane` because the `Pane` contains the form's submit button. Can/should we simple nest multiple `Panes` here?

rasmuswoelk:
@zburke For consistency I would say that direct children of a `<Paneset>` should always be `<Pane>`'s. If that's not the case then the CSS won't work properly – like in the case of ConfigManager form.

I think for that case we can actually convert the `<Pane>` into a `<form>` using the `tagName`-prop. The rest of the props are spread onto the Pane-element so that should work.

Actually, I will give that a shot right now 👍

rasmuswoelk:
Looks like that will work:

image

(The empty div can probably also be removed. Not entirely sure where that comes from.)

zburke:
The non-pane div and the empty div in your example are coming from [ConfigManager](https://github.com/folio-org/stripes-smart-components/blob/406b857445985241281467767fcea618ec6f4ea1/lib/ConfigManager/ConfigManager.js#L116-L119). I think the empty div is caused by the `Callout` when it hasn't rendered. What do you think is the best way to structure this group of components? Move the `Pane` into `ConfigManager`

There is a bunch of minor clean up to be done once we decide on the right way to structure things. 
```
./ui-users/src/settings/ProfilePictureSettings.js
./ui-organization/src/settings/Bindings.js
./ui-organization/src/settings/Locale.js
./ui-tags/settings/TagSettings.js
./ui-circulation/settings/CheckoutSettings.js
```

rasmuswoelk:
@zburke Moving Pane into ConfigManager makes sense to me 👍

PR #908, opened 2019-03-07
UIDATIMP-156: change MetaSection component PropTypes for internationa…

OleksandrYehorovEPAM:
…lization support


lib/MetaSection/MetaSection.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { isValidElement } from 'react';
```
VictorSoroka1: Please add a note to the changelog.

PR #925, opened 2019-03-28
STCOM-492: change filter search to use == not =

hjiebsco:
The "=" syntax is translated to ~ regex search on database side and is very slow. Since filters search the exact key, we should use "==" syntax that translates to LIKE (equivalent to exact in most cases) and that is way faster. For example in location filtered instance search MODINVSTOR-273, we see performance improved about 30x.

### DIFF
--- a/lib/FilterGroups/FilterGroups.js
+++ b/lib/FilterGroups/FilterGroups.js
@@ -185,7 +185,7 @@ export function filters2cql(config, filters = '') {
       let joinedValues = mappedValues.map(v => `"${v}"`).join(' or ');
       if (values.length > 1) joinedValues = `(${joinedValues})`;
 
-      conds.push(`${cqlIndex}=${joinedValues}`);
+      conds.push(`${cqlIndex}==${joinedValues}`);
     }
   }
 


lib/FilterGroups/FilterGroups.js
```diff
@@ -185,7 +185,7 @@ export function filters2cql(config, filters = '') {
       let joinedValues = mappedValues.map(v => `"${v}"`).join(' or ');
       if (values.length > 1) joinedValues = `(${joinedValues})`;
 
-      conds.push(`${cqlIndex}=${joinedValues}`);
+      conds.push(`${cqlIndex}==${joinedValues}`);
```
marcjohnson-kint: Is this only used for exact complete matches, e.g. `ID` properties?

hjiebsco: Yes @marcjohnson-kint that's my understanding (filters2cql). Waiting for UI experts to confirm.

MikeTaylor:
But (and I'm sure @zburke agrees) let's not wait for possible additional optimisation before merging the present PR.

hjiebsco:
Thanks @MikeTaylor and @zburke, I will merge the PR shortly. To answer @zburke's question about `LIKE` and `=`. I had similar question before and @julianladisch pointed me to this link  https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE. It says "If pattern does not contain percent signs or underscores, then the pattern only represents the string itself; in that case LIKE acts like the equals operator." I did compare between `LIKE` and `=` in this case, and did not notice further performance gain by using explicit `=`.

zburke:
Not so fast. This change breaks the [filters test in ui-inventory](https://github.com/folio-org/ui-inventory/blob/ff5fad0a4b8395a395de7e46cef1989d22670c48/test/ui-testing/filters.js) by changing the query from 
```
((title="*" or contributors adj "\"name\": \"*\"" or identifiers adj "\"value\": \"*\"")) 
and languages=("eng" or "spa") sortby title
```
to
```
((title="*" or contributors adj "\"name\": \"*\"" or identifiers adj "\"value\": \"*\"")) 
and languages==("eng" or "spa") sortby title
```
The former returns instance-records that include attributes like 
```
           "languages": [
                "dut",
                "eng"
            ],
```
but the latter query does not. 

Is `languages=("eng" or "spa")` the correct way to query an array-attribute? I remember, very faintly, dealing with this when we first implemented language filters a looooong time ago from [UIIN-3](https://issues.folio.org/browse/UIIN-3) and [FOLIO-921](https://issues.folio.org/browse/FOLIO-921)

hjiebsco:
SQL generated:
```
Before with =
        (
          to_tsvector(
            'simple', instance.jsonb ->> 'languages'
          ) @@ to_tsquery('simple', 'eng')
        ) 
        OR (
          to_tsvector(
            'simple', instance.jsonb ->> 'languages'
          ) @@ to_tsquery('simple', 'spa')
        )
After with ==
        (
          instance.jsonb ->> 'languages' = 'eng'
        ) 
        OR (
          instance.jsonb ->> 'languages' = 'spa'
        )
```

MikeTaylor:
Interesting! It seems to me that the correct interpretation of `==` when used against an array value can only be "exactly matches at least one of the values in the array".

If my interpretation is right, then the present fix to `filters2cql` is correct, but the SQL generation needs to be modified so that it rightly interprets the generated CQL. (And so, as a pragmatic matter, the merging of this PR needs to wait until we have a new cql2pgjson-java release that can do what's needed.)

skoczko:
Guys, for a detailed description about how CQL-to-SQL generation works now vs how it is supposed to to work see CQLPG-81.

@hjiebsco I think we should revert the merge and keep this on the branch until the array searching-related issues are addressed. By switching = with == for all indexes in the generated query you made the language index not use the fulltext (which is NOT using regex anyway) index which can "emulate" array searching.

zburke:
OK, I will revert this PR and restore the branch so we can keep an open PR that is ready to merge as soon as the required changes to cql2pgjson-java are released.

hjiebsco:
Per discussion, I reopened STCOM-492 and made CQLPG-81 as its blocker.

PR #930, opened 2019-04-03
Improve children prop validation in modal footer

mkuklis:
The previous check was failing for me because we can't compare two functions with `===`. This PR uses what react provides for us.

### DIFF
--- a/lib/ModalFooter/ModalFooter.js
+++ b/lib/ModalFooter/ModalFooter.js
@@ -34,15 +34,10 @@ const ModalFooter = ({ children, primaryButton, secondaryButton }) => {
 };
 
 ModalFooter.propTypes = {
-  children: (props) => {
-    let error = null;
-    React.Children.forEach(props.children, (child) => {
-      if (child.type !== Button) {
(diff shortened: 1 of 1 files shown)


doytch:
@mkuklis does this work when we're using HMR via the dev server? Previously we've had to remove checks like this because the HMR wraps components in a `hot()` wrapper that breaks our checks. Eg, the `children` are no longer `Button`s, they're `hot(Button)`s.

mkuklis:
@doytch I'm not sure if that will break with HMR but it was definitely broken before even without HMR. I guess it would be a good thing to test. Is this something you could do @doytch?

doytch:
@mkuklis I can but the issue was not really with HMR _enabled_ but with HMR _as part of the webpack setup_. So assuming this was actually tested with a basic `stripes serve` (as opposed to `stripes build`) and didn't throw errors, then I'm satisfied. :)

doytch:
Tested, works! 👍 

Less console chatter and I learned something new.  Good PR! :)

mkuklis:
@doytch thank you for double testing this!

PR #932, opened 2019-04-10
UIIN-319 Potential fix for gap below results list.

JohnC-80:
This is a difficult issue to reproduce locally.

Short of a heavier refactor for MCL, an initial approach is to stick to a minimum row-height for rendering rows and determining when more data is needed rather than an averaged row height.

This includes a fix to rowCache logic that might have been the culprit as well as the mentioned minimum row height.
With virtualization, the main goal is to render only a portion of the data in situations where displaying all of the data would be quite slow. Despite the loaded amount being more in some cases due to the height number being lower, it's still in the acceptable realm.

This fix is non-breaking and tests still pass for the component.

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -111,6 +111,7 @@ class MCLRenderer extends React.Component {
       averageRowHeight: 0,
       adjustedHeight: null,
       endOfListHeight: null,
+      minimumRowHeight: 24,
     };
 
     this.handlers = Object.assign({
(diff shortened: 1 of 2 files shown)


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -273,16 +274,16 @@ class MCLRenderer extends React.Component {
     // }
   }
 
-  updateDimensions = (height, data, avgHeight) => {
+  updateDimensions = (height, data, minHeight) => {
```
VictorSoroka1: Is `minHeight` param needed? I have not found its usage so probably it is excess.

JohnC-80: Thanks Viktor! I'll take it out.

doytch:
I actually _just_ ran into this bug myself, remembered this PR, checked out the branch, and the bug went away! 😄

lib/MultiColumnList/MCLRenderer.js
VictorSoroka1: Thank you, John, for handling this issue.

PR #935, opened 2019-04-16
STCOM-501: Add basic accordion component

Yurii-Danylenko:
https://issues.folio.org/browse/STCOM-501

## Purpose
Create basic accordion for notes with MCL table,  "Add" and New"  buttons, which will be used to manage work with notes inside the UI module where notes used. It's the beginning part of the whole component. Next step is to add embed modal with possibility to search and attach notes.

## Approach
- Create Accordion
- Create NotesList component
- Add tests
- Add translations
- Add storybook

## TODO
- Add modal component for attaching notes


lib/NotesAccordion/readme.md
```diff
@@ -0,0 +1,36 @@
+# NotesAccordion
+
+Accordion enhanced to work with Notes
+
+## Basic Usage
+
+```js
+  <NoteAccordion
```
zburke:
```
<NotesAccordion...
     ^
     ^
```

Yurii-Danylenko: Ups, thank you @zburke!

JohnC-80:
This seems more specific to Notes functionality than what we usually like to keep in stripes-components. We'd like for new components to be functional at the smallest level rather than larger structures/molecules. Are there still plans for a smart-component? Despite not being connected to data, it seems better to have this live in a notes module/npm package/smart-component than here.

Yurii-Danylenko:
From the conversation with John, it was decided to move notes accordion to smart components

PR #944, opened 2019-04-26
RadioButton and Checkbox style improvements

rasmuswoelk:
In this PR I have made some style improvements for the RadioButton and Checkbox components. The motivation for doing so is related to UIIN-503 because we need to separate the label from a RadioButton while still maintaining a decent layout.

While I was making these updates I decided to do some other improvements as well.

## Updates
- Removed unnecessary margin from the `<label>`-element applied by the `<Label>`-component but not needed in this component and it was causing misalignment.
- Added a min-height of 24px which matches the min-height of form elements. This improves the layout when various form elements are placed inline with checkboxes/radio buttons
- Adjusted the margin of the labelText element of both components so that it only has margin on one side instead of both (adjusted for RTL as well)
- Adjusted the box offset of the interactionStyles for both components since it was a bit too big horizontally
- Added horizontal margin for stacked inline checkboxes/radiobuttons
- Added example for stacked checkboxes
- Added example for a single labelless RadioButton with an external `<Label>` instead

image

image

### DIFF
--- a/lib/Checkbox/Checkbox.css
+++ b/lib/Checkbox/Checkbox.css
@@ -45,6 +45,18 @@
   }
 }
 
+
+/* Apply spacing between inline checkboxes when stacked */
+.inline + .inline {
+  margin-left: var(--gutter-static-two-thirds);
+  margin-right: 0;
+}
+
+[dir="rtl"] .inline + .inline {
+  margin-right: var(--gutter-static-two-thirds);
+  margin-left: 0;
+}
+
 .labelCheck {
   position: relative;
   width: var(--checkbox-size);
@@ -116,7 +128,7 @@
 }
 
 .checkboxInteractionStyles {
-  composes: interactionStyles boxOffsetStartMedium boxOffsetEndMedium from "../sharedStyles/interactionStyles.css";
+  composes: interactionStyles boxOffsetStartSmall boxOffsetEndSmall from "../sharedStyles/interactionStyles.css";
 }
 
 .labelFocused {
@@ -130,10 +142,14 @@
 }
 
 .labelText {
-  margin: 0 var(--gutter-static-one-third);
+  margin: 0 0 0 var(--gutter-static-one-third);
   display: flex;
   align-items: center;
   min-height: var(--control-min-size-desktop);
+}
+
+[dir="rtl"] .labelText {
+  margin: 0 var(--gutter-static-one-third) 0 0;
 }
 
 /* fix for shrinkage issues with checkbox as a flex-child */
--- a/lib/Checkbox/Checkbox.css
+++ b/lib/Checkbox/Checkbox.css
@@ -103,6 +103,7 @@
   position: relative;
   border-radius: var(--radius);
   font-size: var(--font-size-medium);
+  margin-bottom: 0;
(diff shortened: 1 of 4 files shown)


rasmuswoelk:
@JohnC-80 For the RadioButton/Checkbox without an external `<Label>` like e.g.:

image

We would need to control the ID applied to the input explicitly so that we can apply the `htmlFor` for our external label without having the ID applied in the `htmlFor`-attribute on the internal label of the component. I hope that makes sense.

For this, I propose that we add an "inputId"-prop that can control the ID of the input alone. The markup for this example could look something like this:

image

The output would then be:
image

doytch:
@rasmuswoelk looks solid.

One thing that I ran into a while ago was the display of `warning` (and I assume `error`) messages for `Checkbox` elements. I was using a `Checkbox` controlled by a `Field` inside of a `<Col xs={2}>` and the warning/error text was just brutal because it showed up to the right of the Checkbox. As a result, the label, checkbox, and warning had to cram into the horizontal `xs={2}` space.

I ended up reworking it to display the warning on another Field since it was logically possible in that scenario, but I wonder if you could take a quick peek at that use case and see if there's any low-hanging fixes that might be possible? Displaying the warning/error underneath the label and checkbox would help, but I don't know if that would ruin current users of the component.

JohnC-80:
The timeline for these "checkable" inputs has been bumpy. The labels were separate and completely conditional before.
Due to conflicting id's, (`htmlFor`'s screwed up), the label was changed to wrap the input... the double-edge was that the label structure is now a bit more hard-coded in place.
I don't think a conditional rendering of an unwrapped input has ever been introduced, but ideally, the wrapping `<label>` mark-up isn't forced if there's no label prop. Accessibility requires these inputs to always have labels associated, but the means of doing that in the DOM are all optional, thus making it impossible for any to be `required`.
I think I would rather conditionally render the input without the wrapping label if there was no label prop provided.
Being able to set up an external label can give us what we want (vertical layout of label/input) without much fuss within the component - and it's in-line with the openness for accessibility that the component should have anyhow. We ought to be able to accomplish it with just the `id` prop and not add some other weird exception to the list of props.

rasmuswoelk:
@doytch I have noticed the odd placement of warnings/errors. I agree that we should move them to below the label instead. Do you have a screenshot of what you have experienced? And would it be possible for you to create a ticket for this issue?

rasmuswoelk:
@JohnC-80 Yup.. it has been bumby indeed.

The only issue I see here is the fact that our Checkbox/RadioButton components are custom made using a hidden input-element and styling the `<label>`-element to act as the toggle.

I think we can make some workarounds for this but I would prefer to make these changes in another PR. Maybe I can create a ticket to describe this task and we can merge the current changes first?

doytch:
@rasmuswoelk, I went ahead and recreated the issue, took a couple of screenshots, and attached them to [STCOM-514](https://issues.folio.org/browse/STCOM-514). I didn't assign it to any teams/person/sprints so if you want it assigned to you let me know (or toggle it yourself, naturally).

rasmuswoelk:
@doytch Thanks!

PR #949, opened 2019-05-02
Removed gray background color for disabled 'dropdownItem' styled butt…

rasmuswoelk:
**Ticket:** https://issues.folio.org/browse/STCOM-516

- Removed the gray background on disabled Buttons with a buttonStyle of "dropdownItem". 
- Added an example for a disabled "dropdownItem" Button
- Ensured that iterationStyles :active style isn't applied on :disabled buttons

## Before
image

## After
image

### DIFF
--- a/lib/Button/Button.css
+++ b/lib/Button/Button.css
@@ -209,17 +209,17 @@
     padding-right: 0;
   }
 
+  &[disabled] {
+    &:hover {
+      cursor: initial;
+    }
+  }
+
   &[disabled]:not(.dropdownItem) {
     background-color: #ccc;
     border-color: #ccc;
     color: var(--color-text-p2);
     transition: opacity ease-in-out 500ms;
-  }
-
-  &[disabled] {
-    &:hover {
-      cursor: initial;
-    }
   }
 
   &.mega {
--- a/lib/Button/Button.css
+++ b/lib/Button/Button.css
@@ -214,10 +214,7 @@
     border-color: #ccc;
     color: var(--color-text-p2);
     transition: opacity ease-in-out 500ms;
-
-    &:hover {
-      cursor: initial;
-    }
+    pointer-events: none;
   }
 
   &.mega {
@@ -250,7 +247,6 @@
 .inner {
   display: flex;
   align-items: center;
-  pointer-events: none;
 }
 
 /**
--- a/lib/Button/Button.css
+++ b/lib/Button/Button.css
@@ -250,6 +250,7 @@
 .inner {
   display: flex;
   align-items: center;
+  pointer-events: none;
 }
 
 /**
--- a/lib/Button/Button.css
+++ b/lib/Button/Button.css
@@ -210,16 +210,14 @@
   }
 
   &[disabled] {
-    &:hover {
-      cursor: initial;
-    }
-  }
-
-  &[disabled]:not(.dropdownItem) {
     background-color: #ccc;
     border-color: #ccc;
     color: var(--color-text-p2);
     transition: opacity ease-in-out 500ms;
+
+    &:hover {
+      cursor: initial;
+    }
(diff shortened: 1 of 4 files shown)


lib/Button/Button.css
```diff
@@ -210,14 +210,16 @@
   }
 
   &[disabled] {
+    &:hover {
+      cursor: initial;
+    }
+  }
+
+  &[disabled]:not(.dropdownItem) {
```
JohnC-80: The only reservation I have is that `:not()` selectors ramp up specificity (similar to adding a `.notdropdownItem` class to everything that's `[disabled]` so if there are exceptions to these styles, it will be tough to apply them. Could the same be accomplished with`.dropdownItem[disabled]` ?

rasmuswoelk: Good point. Yes, that shouldn't be a problem. I'll make the update 👍🏻

rasmuswoelk: I made the change.

lib/sharedStyles/interactionStyles.css
```diff
@@ -179,9 +179,9 @@
 /**
  * Active
  */
-.isActive,
-.interactionStylesControl:active > .interactionStyles,
-.interactionStyles:active {
+.isActive:not(:disabled),
+.interactionStylesControl:active:not(:disabled) > .interactionStyles,
+.interactionStyles:active:not(:disabled) {
```
JohnC-80: Same issue here for `:not()`s - could these be handled in a less specific way?

rasmuswoelk: Hmm. I don't think so. We don't want to apply any :active-styles when a disabled element becomes active which is why I added these :disabled selectors for all interaction styles.

The alternative would be to overwrite the color for each of the affected components when they are :active and :disabled but that would be just as specific as these styles I think..

I could also remove these selectors but then this will happen:

May-07-2019 18-39-12

Do you have an alternative solution?

JohnC-80: hmm... Yeah, those interactionStyles are immensly complex/specific already. Whichever you decide is best... In this outlying case, I think it's okay to, instead of affect the entirety of :active selectors on non-disabled elements, to single out the single component/class... `.button.dropdownItem` was the one selector to not quite behave the way that we'd like... I would rather add the specificity in one place than in the many as would be impacted by `:not()`.
If you were to remove the `:not()`s and add this to the `.dropdownItem` class of `<Button>`... 
```
&:disabled, &:active:disabled {
    background-color: transparent;
    border: 0;
    & svg {
      fill: var(--color-text-p2);
    }
    & * {
      color: var(--color-text-p2);
    }
  }
```
Thoughts? Unless you imagine this might be an issue in other places besides disabled buttons specifically with the `.dropdownItem` class...

rasmuswoelk: @JohnC-80 The issue also affects the "secondary" buttonStyle as you can see on the GIF above.

But after looking at it again I found out that it actually was caused by the inner element of the button. When adding `pointer-events: none;` to the inner element, there's no longer need for overwriting any styles because the active styles won't be applied in any circumstance.

I pushed a commit for this change. It doesn't change the behavior of the button and it fixes the issue.

JohnC-80: In that case, could we instead add `pointer-events: none` to the [disabled] button as a whole? in Button.css, the selector has a `:hover` block as part of it that we could also eliminate. Doing it this way also shouldn't trigger any color contrast issues with aXe since it's applied directly to the disabled element, rather than one of its children.

rasmuswoelk: @JohnC-80 Even better! I pushed a commit for that change 👍

PR #950, opened 2019-05-02
UIDATIMP-163: Make `EndOfList` component within MCLRenderer to be centered

VictorSoroka1:
Additional changes were made for the EndOfList component within MCL in order to handle the centering issue within data-import app. It turned out that centering feature was worked just fine only for the `SearchAndSort` component, but not just for the MCL outside of `SearchAndSort`.
EndOfList

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -834,11 +834,10 @@ class MCLRenderer extends React.Component {
       <div
         key="end-of-list"
         ref={(ref) => { this.endOfListRef = ref; }}
+        className={css.mclEndOfList}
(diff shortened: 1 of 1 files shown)


JohnC-80:
Hmm... I think you should have a 'reset' button next to the search button... the clear button on TextField is only for the TextField, not for the TextField AND the results... further, does a user need to reset the results in order to search again with a different value?
Requiring a clear button on an empty text field seems like a very odd use case and makes me think there's a better way to handle this.

VictorSoroka1:
@JohnC-80, agree as we discussed in slack I do not need changes in central component and would proceed with an additional Reset button for that purpose alongside with Search button.

PR #956, opened 2019-05-10
accept props.dataTest to provide a test hook

zburke:
BigTest likes to find things in the DOM with selectors like
`[data-test-foo]` but there has been no way to provide a `data-test-`
prop to a `<Pane>` or `<Paneset>` until now. This caused folks to wrap
them like `<div data-test-foo>` but the extra wrapper div can cause
layout problems.

Fixes [STCOM-527](https://issues.folio.org/browse/STCOM-527). Refs [UIORG-166](https://issues.folio.org/browse/UIORG-166).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Create common WYSIWYG editor. STSMACOM-190
 * Additional props validation in `<MultiColumnList>`. Fixes STCOM-515.
 * Panes/Panesets `defaultWidth` supports css-units outside of percent. Fixes STCOM-521.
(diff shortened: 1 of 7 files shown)


JohnC-80:
I was thinking the same thing, @mkuklis .. since `<Pane>` passes `{...props}` to its outer element, it appears to work there... paneset may require something special - either eslint-ignoring a few more destructured props... or doing something like a _.pickBy on props for the data attributes and applying those to the outer `<div>`....

zburke:
Doh! I forgot `data-*` would be passed down automatically. I'll close this and try a simpler approach.

PR #961, opened 2019-05-15
UIDATIMP-198: Add check for the container element existence in `P…

VictorSoroka1:
…aneset` component as sometimes in tests this causes error

### DIFF
--- a/lib/Paneset/Paneset.js
+++ b/lib/Paneset/Paneset.js
@@ -200,7 +200,15 @@ class Paneset extends React.Component {
     let staticSpace = 0;
     const dynamics = [];
     const widths = [];
-    const containerWidth = this.container.current.offsetWidth;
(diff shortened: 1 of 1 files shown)


lib/Paneset/Paneset.js
```diff
@@ -200,7 +200,15 @@ class Paneset extends React.Component {
     let staticSpace = 0;
     const dynamics = [];
     const widths = [];
-    const containerWidth = this.container.current.offsetWidth;
+    const container = this.getContainer();
```
NikitaSedyx: actually I don't see a lot of sense to use `this.getContainer`  as it just returns `this.container.current`

VictorSoroka1: Since it exists I would rather use it instead of writing `this.container.current`.

PR #963, opened 2019-05-16
 UIOR-206 range query support

NikitaSedyx:
original ticket
https://issues.folio.org/browse/UIOR-206

approach:
introduce new attributes to support ranges in cql, it shouldn't break current behavior

Documentation updated

### DIFF
--- a/lib/FilterGroups/FilterGroups.js
+++ b/lib/FilterGroups/FilterGroups.js
@@ -182,10 +182,18 @@ export function filters2cql(config, filters = '') {
         return (obj.length > 0) ? obj[0].cql : v;
       });
 
-      let joinedValues = mappedValues.map(v => `"${v}"`).join(' or ');
-      if (values.length > 1) joinedValues = `(${joinedValues})`;
+      if (group.isRange) {
+        const { isIncludingStart = true, isIncludingEnd = true, rangeSeparator = ':' } = group;
+        const [start, end] = mappedValues[0].split(rangeSeparator);
+        const startCondition = `${cqlIndex}>${isIncludingStart ? '=' : ''}"${start}"`;
+        const endCondition = `${cqlIndex}<${isIncludingEnd ? '=' : ''}"${end}"`;
 
-      conds.push(`${cqlIndex}=${joinedValues}`);
+        conds.push(`(${startCondition} and ${endCondition})`);
+      } else {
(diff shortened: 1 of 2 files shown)


lib/FilterGroups/FilterGroups.js
```diff
@@ -182,10 +182,15 @@ export function filters2cql(config, filters = '') {
...
       });
 
-      let joinedValues = mappedValues.map(v => `"${v}"`).join(' or ');
-      if (values.length > 1) joinedValues = `(${joinedValues})`;
-
-      conds.push(`${cqlIndex}=${joinedValues}`);
+      if (group.isRange) {
+        const [start, end] = mappedValues[0].split(group.rangeSeparator || ':');
+        const equal = group.isEqual ? '=' : '';
+        conds.push(`(${cqlIndex}>${equal}"${start}" and ${cqlIndex}<${equal}"${end}")`);
```
craigmcnally: These ranges are exclusive (e.g. after start and before end, but not exactly start or end)... should they be inclusive?  Maybe a question for Dennis

craigmcnally: nevermind I misread the code

lib/FilterGroups/readme.md
```diff
@@ -83,6 +83,9 @@ filter groups. Each group is represented by an object with four keys:
   query clause is always included. (See
   [STCOM-204](https://issues.folio.org/browse/STCOM-204) for
   rationale.)
+* `isRange` -- The flag that is used to build range cql query. By default `false`.
+* `isEqual` -- The flag is used only with `isRange: true` to define equality operator for range values. By default `false`.
```
AndreiKandratsenka: Would you think about name ```isIncludeEdge```

julianladisch: There should be two flags: `isIncludingStart` and `isIncludingEnd`. The default should be `true` for both of them because that is the most natural query.
References: https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals

NikitaSedyx: actually @craigmcnally 's comment gave me idea us just `isIncluding`

NikitaSedyx: @julianladisch good point, agreed!

NikitaSedyx: @julianladisch updated

PR #964, opened 2019-05-20
UIDATIMP-200: Handle horizontal scroll issue for mac os when scrolls …

VictorSoroka1:
…are always displayed.

Screen Shot 2019-05-20 at 12 46 19

It turned out that when the scroll is always shown the `scollContainer` is continued to accumulate its scroll value even when it reached the most right end, so it is not the right element for endOfList element for calculation its position whereas table header row is a proper one as it does not do accumulation.

chrome-capture

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.css
+++ b/lib/MultiColumnList/MCLRenderer.css
@@ -153,7 +153,7 @@
 
 .mclEndOfList {
   color: var(--color-text-p2);
-  padding: 7px 0;
-  position: absolute;
+  padding: var(--gutter-static-two-thirds) 0;
+  position: relative;
   left: 0;
 }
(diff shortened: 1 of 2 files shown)


rasmuswoelk:
I was able to reproduce a bug that looks very similar to the one on your screenshot:

https://www.dropbox.com/s/1ctkjdf0qwn1hbb/mcl-scrollbug.mp4?dl=0

1. Click on a result
2. Scroll the results list
3. Close the preview pane

VictorSoroka1:
@rasmuswoelk, sorry but this is not related to what I am fixing.

rasmuswoelk:
I can't really tell if the solution is the right one but I can confirm that it works as intended on Chrome / Mac. I think @JohnC-80 is more suited to review the implementation itself.

PR #965, opened 2019-05-20
STCOM-460: Fixed "closeOnBackgroundClick" behaviour for <Modal> 

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/STCOM-460


rasmuswoelk:
@JohnC-80 We already have tests in place that checks if the `onClose`-callback runs when the overlay is clicked and the `closeOnBackgroundClick`-prop is `true`. Should I add a test for the opposite? 🤔

JohnC-80:
Clicking when `closeOnBackgroundClick` is falsey/undefined? The more coverage the better. The un-covered branches of statements can add up.

rasmuswoelk:
@JohnC-80 Yes. I'll add a test for it 👍

rasmuswoelk:
@JohnC-80 I added the test

PR #966, opened 2019-05-20
provide a list of international currencies

zburke:
Provide a list of international currencies derived from the ISO-4217
currency standard at
https://www.currency-iso.org/dam/downloads/lists/list_one.xml. Include a
static list of constants, as well as a script for generating said list
and translations.

Refs [STSMACOM-205](https://issues.folio.org/browse/STSMACOM-205)


zburke:
eh; too much effort.

PR #967, opened 2019-05-20
implement currency select menu

zburke:
Refs [STSMACOM-205](https://issues.folio.org/browse/STSMACOM-205)


zburke:
Replaces #966.

PR #971, opened 2019-05-24
added name and onFocus properties

matteopascucci:
added name and onFocus properties


lib/AutoSuggest/AutoSuggest.js
```diff
@@ -133,7 +140,7 @@ class AutoSuggest extends React.Component {
                 {...getMenuProps({
                   id: autoSuggestId,
                   style: {
-                    width: `${listWidth}px`,
+                    width: `${this.getInputWidth()}px`,
```
JohnC-80: Thanks @matteopascucci - it looks like we need to just remove the `listWidth` variable that was replaced here so that SonarCube will give this a passing grade. Otherwise LGTM.

PR #973, opened 2019-05-27
Update AutoSuggest.js

matteopascucci:


doytch:
Code looks fine and all, but just a note in case it saves you the time it cost me:

I think `AutoSuggest` hasn't really kept up with the times and I ran into a lot of issues with styling and getting it to fit nicely into a Stripes app. I eventually switched to using the `Selection` component because with its `formatter`, `onChange` and `onFilter` props, it seems to offer everything `AutoSuggest` does but in a more developer-friendly (and user-friendly, actually) package.

(In fact, I'd be tempted to start a conversation about what it'd take to deprecate `AutoSuggest` 🤔 )

PR #974, opened 2019-05-28
<Checkbox> refactoring and STCOM-529 (vertical label)

rasmuswoelk:
In this PR I have made some refactoring of the Checkbox-component as well as added the option for a vertical label as requested in https://issues.folio.org/browse/STCOM-529.

## Changes
- Prop cleanup and refactoring of Checkbox-component
- Updated and simplified storybook example
- Added readme for Checkbox-component
- Updated positioning and font-size of feedback element (warning/error)
- Updated warning color variable to be WCAG compliant
- Added some additional tests for Checkbox – coverage is now 100% 🎉

### Changed props
I have added some new props for the component and also removed some obsolete/legacy props

**Removed props:** 
- hover
- labelStyle
- marginBottom0

**Added props:**
- vertical (changes the location of the label to render vertically instead of horizontally)
- inputRef (adds a ref for the input element)
- labelClass (adds a custom class name for the label element)
- innerClass (adds a custom class for the inner element – currently only used to make the checkboxes within the FilterGroups-component work as intended)

### Warning color
When testing with aXe I found that the warning color contrast ratio, when used on a white background, was insufficient so I updated the `--warning` color variable to be slightly darker since this color is most often used on a white background color.

After the update:
Screenshot 2019-05-27 15 21 00

### Feedback element
As described in https://issues.folio.org/browse/STCOM-514, the location of the feedback element can cause some issues – especially if the character length of the warning/error is long.

To improve this, I have updated the placement of the feedback element to be below the checkbox/label instead of next to it. 

This matches the design of other form components in stripes-components, it makes the warning more visible and it adds improvement for the issue in STCOM-514. 

If the feedback text is lengthy and the surrounding space is small, we will obviously still have an issue but I think this is better than before and we can always improve further later on (e.g. using a tooltip or similar).

**Before:**
image

**After:**
image

### Vertical label
By adding the `vertical`-prop on a Checkbox it will now render the label above the checkbox itself.

image

## Testing
To ensure that the component still works as intended I have:
- Run the components tests
- Tested with aXe accessibility plugin for Chrome
- Tested in UI modules in my local version of the FOLIO platform

**Note:** In order to achieve the custom design with CSS alone, the component is wrapped in a `<label>`-element. aXe will complain when the component is used with an external label-element because multiple label-elements isn't allowed. It's an edge-case and it will still work – afaik most screen readers won't have an issue with this. Hopefully, @JohnC-80 and I can find a proper solution for this.


rasmuswoelk:
@JohnC-80 Thanks! Yup – I always squash and merge.

PR #976, opened 2019-05-28
STCOM-534: Card component

doytch:
- Added new `Card` component to facilitate reuse of styles when creating card-based layouts including those [with remote-resource-lookup patterns](https://issues.folio.org/browse/UXPROD-1681).
- Has `default`, `positive` and `negative` styles to apply background styles.
- Has `roundedBorder` and `hasMargin` props to compose styles.
- Supports application of classes directly onto the card, header, and body elements.
- Complete rendering control of header and body.

cards


lib/Card/readme.md
```diff
@@ -0,0 +1,36 @@
...
+  headerEnd={<Button>Click Me!</Button>}
+  headerStart={<span className="my-card-header">My Header</span>}
+>
+  <div>This is my card body content!</div>
+</Card>ma
+```
+
+## `cardStyle`
+
+The `default` card style is selected automatically. Optionally, you can set the `cardStyle` prop to `lookupPreview` and `lookupEmpty` for easy application of the styles required for a remote resource lookup pattern.
```
JohnC-80: Is there a more defined system-standard for these styles where we could say "use 'lookupPreview' style for this state, Use 'lookupEmpty' ... etc"
Maybe even an example of code that would apply the prop based on the presence of a resource. @filipjakobsen ?

lib/Card/Card.js
```diff
@@ -0,0 +1,105 @@
...
+            >
+              {this.props.headerEnd}
+            </span>
+          )}
+        </div>
+        <div
+          data-test-card-body
+          className={this.getBodyStyle()}
+        >
+          {this.props.children}
```
JohnC-80: Minor, but maybe destructure `headerEnd`,`headerStart`, and `children`

lib/Card/readme.md
doytch: Yeah I'd love to know the specific nomenclature we'll be using for this going forward as well since I don't know it :) 

I don't know exactly how we'd automatically apply it though. The place where we'd use it would be [something like here](https://github.com/folio-org/ui-agreements/blob/master/src/components/Agreements/ViewAgreement/Sections/Organizations.js#L40).

JohnC-80: hmm... without particular meaning for these specific names, I'd probably go with something more generic `bordered` and `solid` perhaps since the prop itself just assigns a preset style... With that, I'd say we could merge this in.

filipjakobsen:
I actually don't think the empty state should be a part of the card itself, but rather be a part of whichever area the card exists in. In some places, users will have to add multiple cards in one section. This parent section may then be empty or not.

filipjakobsen:
If the two different styles are intended to support nesting, then "figure" and "ground" work, respectively; or "positive" and "negative". The "bordered" version is actually not just outlined; it also has a white background, which will be visible when place on top of a grey area.

lib/Card/Card.css
```diff
@@ -0,0 +1,71 @@
+@import '../variables.css';
+
+.card {
+  width: 100%;
+  border-color: var(--color-border-p2);
+  border-style: solid;
+  border-width: 1px;
+  margin-bottom: var(--gutter);
+
+  &.default {
```
rasmuswoelk: I think we could avoid nesting here to make the CSS less specific (making it easier for devs to overwrite styles if needed).

This class will end up being `.card.default` – instead, we could just target `.default`.

The goes for other nested style class names.

lib/Card/Card.css
```diff
@@ -0,0 +1,71 @@
...
+    border-width: 1px 1px 0;
+    margin-bottom: 0;
+
+    &:last-of-type {
+      border-bottom: 1px solid var(--color-border-p2);
+      margin-bottom: 4px;
+    }
+  }
+
+  &.lookupEmpty,
```
rasmuswoelk: I would definitely prefer some more generic style names – e.g. "bordered" and "radius" or similar.

lib/Card/Card.css
```diff
@@ -0,0 +1,71 @@
...
+    margin-bottom: 0;
+  }
+}
+
+.header {
+  background-color: var(--bg-hover);
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+  padding: 8px var(--gutter);
```
rasmuswoelk: Here I would recommend using the `--gutter-static` variable instead of `--static`. The static variable doesn't expand if the root font-size changes.

Instead of the 8px vertical padding, you can use `--gutter-static-one-third`. This would add 5px instead of 8px. If you want the header height to be a little bit bigger then you could add a min-height. Smaller vertical padding combined with a min-height would also ensure that the header doesn't get too tall if the content inside is big (making the distance between the body and the header too big).

lib/Card/Card.js
```diff
@@ -0,0 +1,107 @@
...
+      PropTypes.arrayOf(PropTypes.node),
+    ]),
+    headerStart: PropTypes.oneOfType([
+      PropTypes.node,
+      PropTypes.arrayOf(PropTypes.node),
+    ]).isRequired,
+  }
+
+  static defaultProps = {
+    cardStyle: 'default',
```
rasmuswoelk: If the default style is "default" and this style adds margin-bottom: 0; then the `marginBottom0`-class name becomes obsolete unless you actually add a`cardStyle`, right? 

In that case, maybe the card shouldn't have a "cardStyle" default prop? Coming back to the more generic style names then we could maybe have a "nested" style that removes the bottom-margin for all nested cards except the last one. And maybe also removes the top-border for all .nested cards except the first one.

lib/Card/Card.css
```diff
@@ -0,0 +1,71 @@
...
+  border-color: var(--color-border-p2);
+  border-style: solid;
+  border-width: 1px;
+  margin-bottom: var(--gutter);
+
+  &.default {
+    border-width: 1px 1px 0;
+    margin-bottom: 0;
+
+    &:last-of-type {
```
rasmuswoelk: The "default" style card has no border-bottom which looks odd if there's only one rendered card.

Could we instead do something like:

```css
  .default + .default {
     border-top: 0;
   }
```

This way a single "default"-styled card will have a border-bottom and when multiple cards are nested it removes the border-top instead to avoid double border.

doytch:
Refactored based on the suggestions. Main things are:

- Removed the concept of `lookupPreview` and `lookupEmpty` for the cards. Those styles can now be achieved via setting a `cardStyle` of `positive` or `negative` and applying `hasMargin` and `roundedBorder` props. While I originally wanted to be able to turn on all the styles with one prop, I get why more discrete props can be cleaner.
- I ensured that the default style that's most likely to be used (the repeating card pattern) is the one that requires the fewest props.
- Split out a `DefaultCardHeader` component.

lib/Card/Card.js
```diff
@@ -0,0 +1,89 @@
...
+
+    return (
+      <div
+        className={this.getCardStyle()}
+        {...customProps}
+      >
+        <HeaderComponent {...this.props} />
+        <div
+          data-test-card-body
+          className={`${css.body} ${bodyClass || ''}`}
```
rasmuswoelk: Maybe we can use the "classnames" package here instead?

lib/Card/DefaultCardHeader.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import className from 'classnames';
+
+import css from './DefaultCardHeader.css';
+
+export default class DefaultCardHeader extends React.Component {
```
rasmuswoelk: To keep the component folder clean; could we move this component into its own folder and then place it inside a "headers"-folder?

lib/Card/readme.md
```diff
@@ -0,0 +1,35 @@
...
+</Card>
+```
+
+## Props
+
+| Name | Type | Required | Note |
+--- | --- | --- | --- |
+| `bodyClass` | string | |
+| `cardClass` | string | |
+| `cardStyle` | string | | Must be undefined or one of `default`, `negative`, or `positive`. These generally affect the background of the card.
```
rasmuswoelk: Should we maybe elaborate these styles a bit in the readme?

lib/Card/Card.js
```diff
@@ -0,0 +1,89 @@
...
+      'headerStart',
+      'roundedBorder',
+    ]);
+
+    return (
+      <div
+        className={this.getCardStyle()}
+        {...customProps}
+      >
+        <HeaderComponent {...this.props} />
```
rasmuswoelk: I'm wondering if we should only pass the props related to the header into the header component..  🤔What do you think?

lib/Card/Card.js
doytch: I thought about it but it seemed like overkill given that I'm literally just appending the one class.

lib/Card/Card.js
doytch: Yeah, but if we want to allow users to add arbitrary props to the header (`data-test-etc`) what are the props that are intended for the header? 

Sometimes people add a `headerProps` object to destructure but I'm not a huge fan of it because it grows the props list and it can get you into situations where you have to define a prop in `props` and `headerProps` if you want it going to the card and the header.

I don't feel too strongly about it here, but that's why I wrote it this way instead of the `headerProps` way after thinking about it for a bit. I'm willing to change it to something else though.

JohnC-80:
Whatever we decide on for style names, the usage must have an explanation in the readme if we want the standard to be adopted in the system - they don't even need to be 1-to-1 use-cases, just need to be enough to help a developer (or PO) decide which they need - possibly even based on how @doytch 's module will use them - just with generic names that don't couple it up with the particular use-case. 
A `renderHeader` function that's passed the props might be a better prop option for max composability/separation of props for the header rather than leaving it only with the props passed to the Card.

doytch:
Added a **Common Patterns** section per @JohnC-80's suggestion

lib/Card/Card.js
rasmuswoelk: Fair :) And you are probably right. I just like `classnames` because it automatically prevents undefined or empty classes to be appended.

PR #978, opened 2019-05-30
Add marginBottom0 to Autosuggestion

cchiama:
Add marginBottom0 to Autosuggestion to pass it to parent component

### DIFF
--- a/lib/AutoSuggest/AutoSuggest.js
+++ b/lib/AutoSuggest/AutoSuggest.js
@@ -129,8 +129,8 @@ class AutoSuggest extends React.Component {
                   {...getInputProps(
                   { ...inputProps,
                     onBlur }
-                  marginBottom0
                 )}
+                  marginBottom0
                 />
               </div>
               <ul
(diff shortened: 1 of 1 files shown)


lib/AutoSuggest/AutoSuggest.js
```diff
@@ -122,10 +125,12 @@ class AutoSuggest extends React.Component {
                 aria-live="assertive"
                 aria-relevant="additions"
               >
-                <TextField {...getInputProps(
+                <TextField
+                  {...getInputProps(
                   { ...inputProps,
                     onBlur }
                 )}
+                  marginBottom0
```
rasmuswoelk: It looks like this should be `marginBottom0={marginBottom0}` instead.

cchiama: Ok. I fix the issue now. Thanks @rasmuswoelk

cchiama:
I fix the syntax for marginBottom0 inAutosuggestion. @rasmuswoelk please as soon as you can merge the fix with matteopascucci modify also, because i I need it as soon as possible. Thanks in advance

rasmuswoelk:
@cchiama You closed the issue – was that on purpose?

I can merge it as soon as you re-open it and add the prop in the readme for the component 👍

PR #979, opened 2019-06-05
Export Card component from top-level

doytch:
Realised I forgot to do this literally as I merged the PR... 🤦‍♂


JohnC-80:
Urps... Thanks for handling it!

PR #981, opened 2019-06-07
UIOR-294 Support disable of Add/Remove buttons for RepeatableField

NikitaSedyx:
https://issues.folio.org/browse/UIOR-294

purpose: we need a way to disable creation and remove fields in RepeatableField component

approach: new properties to disable buttons

documentation is added
tests are added


lib/RepeatableField/RepeatableField.js
```diff
@@ -10,6 +10,8 @@ import css from './RepeatableField.css';
 class RepeatableField extends Component {
   static propTypes = {
     addLabel: PropTypes.node,
+    disabledAdd: PropTypes.bool,
+    disabledRemove: PropTypes.bool,
```
JohnC-80: @NikitaSedyx Thank you for this! Any time I see props named in a similar way as these, `disableAction = true` It gives me the thought of "flipping on the off switch" - I suggest something like `canAdd={false} canRemove={false}` where it would default true ( I understand it's easier to add boolean props with just the prop name, but think about use-cases here - disabling based on whether or not permissions are present or anything else - it's best if the prop is mapped to the exact value, not an opposite of) food for thought?

NikitaSedyx: @JohnC-80 agreed, sounds good to me, I'll do it my next working day
thanks for the review

NikitaSedyx:
@JohnC-80 @rasmuswoelk updated, please take a look

PR #983, opened 2019-06-10
Minor PaneHeader appIcon tweak to prevent icon from shrinking

rasmuswoelk:
I noticed that the AppIcon in the PaneHeader was shrinking and being cut in half when the title became too big.

Changing the AppIcon tag to a `div` fixes this.

## Before
image

## After
image


rasmuswoelk:
@JohnC-80 Yes.

PR #988, opened 2019-06-18
I don't think "children" is an actual prop

MikeTaylor:


doytch:
It's the one accessed as `this.props.children` from within the `Card` component. 

Agree you don't _set_ it as a normal prop, but I figured it'd be understood what was meant, esp since the pattern is used in other prop lists in readmes (though without the backticks).

MikeTaylor:
Right, but you don't typically pass it as `<Card children={<p>Hello</p>} />, do you?

So the backticks should go.

MikeTaylor:
I'm not committed to the parentheses — I just think formatting it in `typewriter` is misleading.

But, heck, I don't feel strongly about any of this. Leave it as it is if you prefer.

JohnC-80:
Let's just leave it.

Freaky as heck, but some people do actually pass `children` as a prop... haven't seen it happening on FOLIO, but I have in other places in the codebases at uh... some undisclosed big company somewhere.... I just imagine there's a dogmatic DRY concept to blame for it. I can just hear someone saying "I don't even want to type the component's name again for a closing tag!"

PR #990, opened 2019-06-19
UIU-894: Extend `FilterGroups` component to support optional `displayName` prop for the `values` object for the intl purposes

VictorSoroka1:
Refers to the hardcode [issue](https://issues.folio.org/browse/UIU-894) for the active/inactive entries for the search filter in the *ui-users* application.


lib/FilterGroups/FilterGroups.js
```diff
@@ -244,13 +246,14 @@ const FilterGroups = (props) => {
...
+
     for (const value of group.values) {
       if (typeof value === 'object') {
         if (!value.hidden) {
-          names.push(value.name);
+          names.push({ value: value.name, displayName: value.displayName || value.name });
         }
       } else {
-        names.push(value);
+        names.push({ value: value.name, displayName: value.name });
```
zburke: If `value` isn't an object, shouldn't this be
```
names.push({ value, displayName: value })
```
I don't think `value.name` is defined here.

VictorSoroka1: Indeed. Thanks for spotting this.

PR #995, opened 2019-06-24
UIEH-728: use AppIcon from stripes-core in PaneHeader

MaksymDryha:
To get rid of the deprecation messages in the console, I've made PaneHeader use an AppIcon from stripes-core, not from stripes-components


JohnC-80:
@MaksymDryha Despite resolving console messages, this will create a circular dependency with stripes-core. Does passing an `<AppIcon />` to the `Pane` at the application level work for you? This is the documented basic usage is documented in `PaneHeader`'s `readme.md` - We do need to put this example in `Pane`'s `readme.md` as well...  ```<Pane ... appIcon={<AppIcon ... />}  />```

PR #996, opened 2019-06-24
Update docs for AppIcon API for Pane and PaneHeader

JohnC-80:
We're on the verge of going through with the breaking change. Would like to remove the `stripes-components` version of AppIcon soon. 
Shoring up documentation for both Pane and PaneHeader now.


MaksymDryha:
Thank you!👍

PR #997, opened 2019-06-26
STCOM-497 follow up: Updates for component, readme, tests and examples

rasmuswoelk:
**In this follow-up PR I updated/fixed a few things:**
- Fixed mistake in the readme and updated multiple columns usage example
- Updated "multiple fields" storybook example after testing with aXe
- Added aria-label for remove button
- Moved redux-form related tests into the correct file and removed duplicated tests


lib/RepeatableField/RepeatableField.js
```diff
@@ -88,6 +89,7 @@ class RepeatableField extends Component {
                         onClick={() => onRemove(index)}
                         size="medium"
                         disabled={!canRemove}
+                        aria-label={<FormattedMessage id="stripes-components.deleteThisItem" />}
```
JohnC-80: since `<FormattedMessage>` renders a `<span>`, would it be best to use the render-props style here? Does this work as expected?

rasmuswoelk: Whoops. Nope. I'll fix this. Thanks!

PR #1000, opened 2019-07-02
STCOM-542: Allow for passing custom icons for <Icon>

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/STCOM-542

## Changes
- Updated `<Icon>` to allow for passing a custom icon function that renders a svg
- Updated readme with documentation about adding custom icons
- Updated Icon tests to include test for custom icons and updates for other tests to reach 100% test coverage 🎉
- Moved control of icon props (focusable, viewBox, className) from `icons.js` into the Icon-component itself for central control. This also ensures that our icons match the custom icon example (spreading props onto the SVG)
- Added simple custom icon storybook example
- Changed SVG icons list to be required from icons folder instead of being inlined


rasmuswoelk:
The tests currently fail because of a missing `exclude` in the webpack base config in stripes-core.

I will submit a PR for this change to stripes-core if/when my suggested solution is approved.

JohnC-80:
I like where this is going. Not sure if we could localize the change to the component via a webpack loader override syntax... `!!react-svg-loader!./filename.svg` rather than affect the centralized webpack setup...

rasmuswoelk:
@JohnC-80 I didn't know that you could overwrite! Neat. And TIL.

I updated my PR.

JohnC-80:
This is so awesome I can't even! 
FINALLY the icons are in some easily malleable state!

image

Thank you for the work on this, @rasmuswoelk!

PR #1003, opened 2019-07-09
Make svg loader a direct dep

JohnC-80:
Since `react-svg-loader` was added as a devDep in #1000 ci environments have suffered since it's only a devDep (only installed in local yarn/npm-installs of stripes-components)


rasmuswoelk:
Thanks for fixing this, @JohnC-80!

PR #1006, opened 2019-07-12
STCOM-538 better missing metadata

zburke:
When metadata elements are missing, display placeholder data rather than
omitting the element.
    
Fixes [STCOM-538](https://issues.folio.org/browse/STCOM-538)


zburke:
@funkymalc, [SonarCloud appears to be misconfigured here](https://sonarcloud.io/component_measures?id=org.folio%3Astripes-components&metric=new_coverage&pullRequest=1006&view=list): the coverage statistics include files inside `.../tests/...`, which are the tests themselves.

PR #1011, opened 2019-07-17
STCOM-363 Refactor MCL (columns)

JohnC-80:
MCL currently measures columns before rendering, which can lead to inaccuracies (to say the least). By pushing this measurement to after a DOM render, we can measure based on existing DOM nodes which can give us much better spacing (no more defaults and large spaces between columns)

## Approach
This is a completely revised rendering process. Rather than applying multiple render passes, like before, cells and rows are rendered out-of-site, measured, and positioned JIT. The code works through a subscription model with a cache class that can be made to store a list of single values... so we measure as little as possible.
At the basic level, the process works like this: 

1. Rows with contained cells are rendered off-canvas and heights/final positions are subscribed to since we don't have them up front.
2. Cells are rendered and widths measured (unless provided a width via props.)
3. Once enough cells of each column have been collected, the 3rd quartile value of that collection is committed to state.
4. Once the state has committed column widths, row heights can be measured and cached.
5. Once row-heights are measured, they can be 'accumulated' to obtain a position.

```
<Scrollable>
  <RowContainer>
    <Row>
    <Row>
     ...
  </RowContainer>
</Scrollable>
```
The typical approach to windowing in React is applied here, with rows being absolutely positioned on a larger `RowContainer`. The positions of rows never change, but the component knows which part of the larger container is visible via the `scrollTop` of the `<Scrollable>` container.
It only renders the visible rows, updating the first rendered index via a throttled `onScroll` handler.

- The return JSX of the `render()` function is simplified, removing conditional rendering.
- `componentWillReceiveProps` has been removed.
- Some of the more complex aspects of the component have been moved out to exterior functions/Components.
- Loading is componentized through new `<LoaderRow>` components. These are visible and let users know more data is on the way...
- `<EndOfList>` is componentized.
- Components for the above process (`<RowMeasurer>`, `<RowPositioner>`, `<CellMeasurer>`) also added.
- `deferredCache` - the workhorse of the measuring/positioning approach is added.

It will be possible to use similar techniques/components to virtualize other components such as lists on `<Selection>` and `<MultiSelection>`.


lib/MultiColumnList/tests/MultiColumnList-test.js
```diff
@@ -150,7 +150,11 @@ describe('MultiColumnList', () => {
         await mcl.scrollBody(2500);
       });
 
-      it('calls the supplied onNeedMoreData handler', () => {
+      it('expects loaderRow to be present', () => {
+        expect(mcl.loaderPresent).to.be.true;
+      });
+
+      it('expects loaderRow to be present', () => {
```
doytch: This case should probably be called the `calls the supplied onNeedMoreData handler` that got deleted, no?

JohnC-80: U right.

JohnC-80:
@doytch Thanks for the feedback! I'll pull Licenses down into a workspace and see what I can resolve.

doytch:
Updates based on your updates:

- No MUDE crash anymore. 🎉 
- Not as much blank space at the top of lists and not as often but still some. 
- My Agreements results MCL shift appears to scroll down upon selecting an agreement. The length of the scroll appears to be correlated with the position of the agreement in the visible results list. Eg, when I select an agreement near the top of the list, it'll only scroll down three or four rows, but if I click and view an agreement near the bottom, it'll scroll down like twenty.
- There appear to be a couple of bugs with how the row height is being calculated. See the following two examples:

Screen Shot 2019-07-18 at 9 57 04 AM
Screen Shot 2019-07-18 at 9 56 51 AM

VictorSoroka1:
Looks good. However, I also found a case when it does not work right (I tested on mac). I tested on *ui-users*. When I scrolled to the end of the list and then open details pane and after a while closed it and scrolled a bit up the page may end up blank and in order to see something user would need to do more scrolling to the top.

JohnC-80:
Thanks @VictorSoroka1 and @doytch for the continued feedback. I've seen the 'whitespace at top' behavior that you've guys have described as well as the 'blank-out' that can happen when you transition to a detail and back to a larger width. Hopefully I'll home in on those today - I've got some ideas for the cause there.
With the way that the cache now works (rebase-able from an index within the list) shifting scroll on click may be unnecessary now. The scroll update was my first step in keeping the selected row visible, since the current behavior of MCL is that all the rows shift because the row-height changes - but if the row-position can stay the same/scroll stays the same, then the row should still be visible when the width changes.... stay tuned!

JohnC-80:
I've worked to address the symptoms mentioned earlier... now just a few more tests are needed. @doytch and @VictorSoroka1 you guys want to give it another shot? Thanks!

zburke:
The height does not correctly adjust when adding a second row to a table, as when checking in/out multiple items. The problem appears unique to the second row; the height is correct when adding rows >= 3. 

image

zburke:
I see extra scrollbars in some circumstances:
 
Screen Shot 2019-08-02 at 6 18 14 AM

VictorSoroka1:
https://folio-styling.testing.samhaeng.com/settings/data-import/action-profiles/view?sort=-action%2C-mapping

Found that for `ui-data-import` app, after introducing that behavior, _select all_ and _select_ individual row buttons no longer works. What would be the recommended way to make them work for a new approach?

doytch:
Still not there for me with the `Covered E-resources` list in Agreement view panes.

I went ahead and created an agreement with three agreement lines, each responsible for a single covered e-resource. When I view the agreement and the MCL for the covered e-resources, I initially get this:
init

Within a half-second or so, it coalesces to the following view. (If I close/reopen the accordion, I get this view immediately, rather than moving through the previous view where the rows overlap.)
blank

I can scroll and scroll and scroll through that MCL to my heart's content...even though the `contentData` I'm passing into it is only three items long.

As a refresher, here's my usage:
```
        <MultiColumnList
          columnMapping={this.columnMapping}
          columnWidths={this.columnWidths}
          contentData={visible ? eresources : []}
          formatter={this.formatter}
          height={800}
          id="eresources-covered"
          interactive={false}
          onNeedMoreData={onNeedMoreEResources}
          virtualize
          visibleColumns={this.visibleColumns}
        />
```

JohnC-80:
@doytch I'll look at a different behavior if there's a `virtualize` prop, but no `totalCount` prop. Without the `totalCount`, MCL doesn't know when to quit so it'll keep on with the loader rows. Something that should be possible now is a maxHeight, rather than the height, so the list would shrink down if it's shorter than 800px.

VictorSoroka1:
Accidentally found an issue while being on the branch. When I typed something in the `code` field the cursor for some reason jumped to the `name` field.

chrome-capture (1)

JohnC-80:
@VictorSoroka1 That was an interesting one. It was due to the `autofocus` prop being used within the `EditableList` component and the `MCL` seeing the change in the data and thinking that the data had changed so it needed to refresh.

lib/MultiColumnList/calculateWidth.js
```diff
@@ -0,0 +1,42 @@
+// averages a 3rd quartile width among cell samples and compares with header to derive
+// column widths.
+
+export function calculateColumnWidth3q( // eslint-disable-line import/prefer-default-export
+  widthCache,
+  headerCache,
+  columnName,
+) {
+  const extracted = [];
+  for (const k in widthCache.cache) {
```
VictorSoroka1: I would recommend using [forOwn](https://lodash.com/docs/4.17.15#forOwn) method from `lodash` for it.

It will reduce the boilerplate for iterating and checking for `hasOwnProperty`.
```
for (const k in widthCache.cache) {
 if (Object.prototype.hasOwnProperty.call(widthCache.cache, k)) {
```

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -128,310 +210,327 @@ class MCLRenderer extends React.Component {
...
-      // } else {
-      if (!this.props.virtualize) {
-        newState.amountToRender = this.props.contentData.length;
-        // if(onNeedMoreData){
-        //   onNeedMoreData();
-        //   newState.loading = true;
-        // }
+    function getColumnWidths() {
+      const columnsFromProps = {};
+      for (const c in columnWidths) {
```
VictorSoroka1: I would recommend using [forOwn](https://lodash.com/docs/4.17.15#forOwn) method from `lodash` for it.

lib/MultiColumnList/RowMeasurer.js
```diff
@@ -0,0 +1,81 @@
...
+  componentWillUnmount() {
+    const {
+      heightCache,
+      rowIndex,
+    } = this.props;
+
+    heightCache.clearRequest(rowIndex);
+  }
+
+  handleFocus() {
```
VictorSoroka1:
```suggestion
  handleFocus = () => {
```
so it would become possible to get rid of binding in constructor and thus of constructor.

lib/MultiColumnList/CellMeasurer.js
```diff
@@ -0,0 +1,43 @@
...
+  element = React.createRef();
+
+  measureNode = () => {
+    const { widthCache, rowIndex, measure, onMeasure, columnName } = this.props;
+    if (measure) {
+      // console.log(`measuring ${this.props.rowIndex}`);
+      if (this.element.current && this.element.current.offsetParent !== null) {
+        const width = this.element.current.offsetWidth;
+        widthCache.set(rowIndex, width);
+        if (onMeasure) onMeasure(rowIndex, columnName, width);
```
VictorSoroka1: I suggest setting `onMeasure` in `defaultProps` to be `noop` from lodash which is equivalent to `() => {}`, so in that case we can remove the check for it.

lib/MultiColumnList/calculateWidth.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+  extracted.sort((a, b) => a - b);
+  // if delta between shortest and longest non-zero is small, use the longest.
+  let shortestLength = extracted.find((n) => n > 0);
+  if (!shortestLength) {
+    shortestLength = 0;
+  }
+  let possibleWidth;
+  if (extracted[extracted.length - 1] - shortestLength < 80) {
```
VictorSoroka1: Please add comment why 80 is chosen or more it to a descriptive variable.

lib/MultiColumnList/calculateWidth.js
```diff
@@ -0,0 +1,42 @@
...
+    possibleWidth = extracted[q3Index];
+    if (possibleWidth === 0) {
+      possibleWidth = shortestLength;
+    }
+  }
+
+  let resWidth;
+  const headerWidth = headerCache.request(columnName);
+  if (possibleWidth < headerWidth) {
+    resWidth = headerWidth + 22;
```
VictorSoroka1: Please add a comment why 22 is chosen or more it to a descriptive variable.

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -73,52 +86,121 @@ const defaultProps = {
...
+        props.containerRef(ref);
+        this.container.current = ref;
+      };
+    } else {
+      this.container = props.containerRef || React.createRef();
     }
 
+    this.renderHeaders = this.renderHeaders.bind(this);
+    this.debouncedLoadMore = debounce(this._loadMore.bind(this), 200, { trailing: true, leading: true });
+    this.handleLoadMore = this.handleLoadMore.bind(this);
```
VictorSoroka1: We can avoid the usage of `bind` by using arrow functions for methods. Up to you @JohnC-80 really)

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -128,310 +210,327 @@ class MCLRenderer extends React.Component {
...
-      const dimensions = this.updateDimensions(this.props.height, this.props.contentData);
-      Object.assign(newState, dimensions);
-      // } else {
-      if (!this.props.virtualize) {
-        newState.amountToRender = this.props.contentData.length;
-        // if(onNeedMoreData){
-        //   onNeedMoreData();
-        //   newState.loading = true;
-        // }
+    function getColumnWidths() {
```
VictorSoroka1: I think the function can be taken out of `constructor` by making `columnWidths` as a parameter.

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -128,310 +210,327 @@ class MCLRenderer extends React.Component {
...
+          newState.columnWidths = getColumnWidths();
+        }
+      }
 
-      this.headerHeight = this.headerRow.offsetHeight;
+      const columnVar = columns || newState.columns;
+      if (newState.columnWidths &&
+        columnVar &&
+        (Object.keys(newState.columnWidths).length === columnVar.length)) {
+        newState.rowWidth = Object.keys(newState.columnWidths)
```
VictorSoroka1: I see the similar logic for calculating row width [here](https://github.com/folio-org/stripes-components/pull/1011/files#diff-8f72149c49820410eb93cd710df215aaR349) and [here](https://github.com/folio-org/stripes-components/pull/1011/files#diff-8f72149c49820410eb93cd710df215aaR378) and [here](https://github.com/folio-org/stripes-components/pull/1011/files#diff-8f72149c49820410eb93cd710df215aaR908). Could you please extract it to a method?

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -128,310 +210,327 @@ class MCLRenderer extends React.Component {
...
 
-      // if we're 2 container-heights away (or less ) from the end of the data, fetch more...
-      if (!this.state.loading && (height - currentScroll < height * 2)) {
-        onNeedMoreData();
-        nextState.loading = true;
+        const eOL = this.endOfList.current;
+        if (eOL) {
+          eOL.style.left = `${this.scrollContainer.current.scrollLeft}px`;
+          // 5 pixels are needed to allow recalculation of the list row items when details pane is closed
+          // eOL.style.left = `${Math.max(this.headerRow.current.scrollLeft - 5, 0)}px`;
```
VictorSoroka1: Can it be removed?

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -778,86 +1067,95 @@ class MCLRenderer extends React.Component {
...
+
+    let newHeight;
+
+    /* if the height prop is not defined, the size of the physical container is ever-growing
+    *  with the size of the contentData.
+    */
+    if (height !== undefined) {
+      // if we have a totalCount, we can set size based on that... if not, the receivedRows count.
+      // this.containerHeight comes from the determined position of the EndOfList.
+      if (totalCount) {
```
VictorSoroka1: I would simplify 
```
      if (totalCount) {
        newHeight = Math.max((totalCount * minimumRowHeight), this.containerHeight || height || maxHeight);
        return { height: `${newHeight}px`, width, position };
      } else {
        newHeight = Math.max((receivedRows * minimumRowHeight), this.containerHeight || height || maxHeight);
        return { height: `${newHeight}px`, width, position };
      }
```
to 
```
        newHeight = Math.max(((totalCount || receivedRows) * minimumRowHeight), this.containerHeight || height || maxHeight);
        return { height: `${newHeight}px`, width, position };
```

Does it make sense?

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -473,6 +581,241 @@ class MCLRenderer extends React.Component {
...
+            position: 'absolute',
+            overflow: 'hidden'
+          };
+          if (position !== null) {
+            endOfListContainerStyle.top = `${position}px`;
+          }
+          return (
+            <div data-end-of-list={endIndex} style={endOfListContainerStyle}>
+              <EndOfList
+                width={width - 20}
```
VictorSoroka1: Please add a comment about why 20 is chosen or move it to descriptive variable.

lib/MultiColumnList/MCLRenderer.js
JohnC-80: I'll do a clean-up pass on these and unbind the ones that I can. Some of these started as arrow functions, but ended up throwing errors where `this` would be null for some reason.

doytch:
Still some weird behaviour.

This is what I get when expanding the **Agreement Lines** accordion.
Screen Shot 2019-08-09 at 1 46 21 PM

If I expand the **Internal Contacts** accordion above it, the first MCL correctly renders with an appropriate height.

The second Covered E-resources MCL is still being picky. Here's its usage, now setting the `totalCount` prop.
```
        <MultiColumnList
          columnMapping={this.columnMapping}
          columnWidths={this.columnWidths}
          contentData={visible ? eresources : []}
          formatter={this.formatter}
          height={800}
          id="eresources-covered"
          interactive={false}
          onNeedMoreData={onNeedMoreEResources}
          totalCount={eresourcesCount}
          virtualize
          visibleColumns={this.visibleColumns}
        />
```

Additionally, if I switch from `height` to `maxHeight`, my `onNeedMoreData` function is no longer being called despite the MCL only having a `contentData` array of length 100 and `totalCount` being ~290.

VictorSoroka1:
@JohnC-80, I found the performance drop while typing something when creating new `Contributor type` for /settings/inventory/contributortypes settings. This is similar to issues with stripes-connect which we had previously. Without the changes, while I type something there I found that everything goes smoothly.

Note that, for  the other hand for `Alternative title types` the same works fine.

JohnC-80:
Thanks for the report @VictorSoroka1 I'll try and see where it can be optimized... I bet we would get better performance if EditableLists were all virtualized, rather than re-rendering the entire set of items each time. Also, for cases like this, the production build of React is waaay faster.

VictorSoroka1:
Indeed @JohnC-80, I have not tried with prod build.

PR #1015, opened 2019-07-22
STCOM-524 - Increase stripes-components|key value testing to 80%

maximdidenkoepam:
# Purpose
Increase KeyValue testing coverage to 80%

# Link
https://issues.folio.org/browse/STCOM-524

# Screenshot
Screen Shot 2019-07-23 at 11 41 08

### DIFF
--- a/lib/KeyValue/tests/interactor.js
+++ b/lib/KeyValue/tests/interactor.js
@@ -17,5 +17,4 @@ export default interactor(class KeyValueInteractor {
   label = text('[class^="kvLabel"]');
   hasValue = isPresent('div:nth-child(2)');
   value = text('div:nth-child(2)');
-  children = text();
 });
--- a/lib/KeyValue/tests/interactor.js
+++ b/lib/KeyValue/tests/interactor.js
@@ -1,7 +1,6 @@
 import {
   interactor,
-  isPresent,
-  text,
+  scoped,
 } from '@bigtest/interactor';
 
 import { selectorFromClassnameString } from '../../../tests/helpers';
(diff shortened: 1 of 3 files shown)


lib/KeyValue/tests/interactor.js
```diff
@@ -0,0 +1,20 @@
...
+
+import css from '../KeyValue.css';
+
+const iconClassSelector = selectorFromClassnameString(`.${css.kvRoot}`);
+
+export default interactor(class KeyValueInteractor {
+  static defaultScope = iconClassSelector;
+
+  hasLabel = isPresent('[class^="kvLabel"]');
+  label = text('[class^="kvLabel"]');
```
OleksandrAntonenko1: I think there might be problems with a couple of `KeyValues` on the same page so this can bring complexity to reusing interactor. Can you check it out by creating a couple of `KeyValues` on the same page and test them out? I think if I right you can use `scoped` instead of `label` and `text`.

lib/KeyValue/tests/interactor.js
```diff
@@ -0,0 +1,20 @@
...
+  hasValue = isPresent('div:nth-child(2)');
+  value = text('div:nth-child(2)');
```
JohnC-80: This is a perfect opportunity to perform this selection with a data-test attribute than the rigid 2nd div of the selector. It's fine to add `data-test` attributes to the component - just have to be sure you're able to test individual instances of the component, as @OleksandrAntonenko1 mentioned earlier.
For class names, I tend to use template string syntax `.${css.kvLabel}` just for brevity over `[class^="kvLabel"]

maximdidenkoepam: Done

lib/KeyValue/tests/interactor.js
maximdidenkoepam: Done

PR #1017, opened 2019-07-23
STCOM-526 - increased coverage for Accordion smart component

maximdidenkoepam:
# Purpose
To increase coverage for Accordion component.

# Approach
Seems we have duplicated code which can be refactored in order to increase coverage up to 87%

# Link
https://issues.folio.org/browse/STCOM-526

# Screenshot
Screen Shot 2019-07-23 at 13 13 01

### DIFF
--- a/lib/Accordion/Accordion.js
+++ b/lib/Accordion/Accordion.js
@@ -97,20 +97,8 @@ class Accordion extends React.Component {
     // prop/state changes to open status set off the animation callbacks.
     let shouldAdjust = false;
     // if using state...
-    if (typeof this.state.isOpen === 'boolean') {
-      if (this.state.isOpen !== prevState.isOpen) {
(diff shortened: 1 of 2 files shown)


lib/Accordion/Accordion.js
```diff
@@ -97,20 +97,8 @@ class Accordion extends React.Component {
...
-          this.transitioningOpen = true;
-          this.syncedHeight = false;
-          this.expandRaf();
-        } else {
-          shouldAdjust = true;
-          this.transitioningClosed = true;
-          this.collapseRaf();
-        }
-      }
-    } else if (this.props.open !== prevProps.open) {
```
maximdidenkoepam: Seems the else if condition is obsolete and can be simplified to next construction that will increase code coverage.

PR #1024, opened 2019-08-02
STCOM-429: Fixed footer for Save and Cancel buttons

SergiyVSergiyenko:
### Purpose
Create element with Save and Cancel buttons, fixed to the bottom of a pane containing a form, with styling as shown on [link](https://xd.adobe.com/view/395087a8-d2cf-42cf-643b-e2a5ba6b4728-4f61/?fullscreen) with changes according to [story](https://issues.folio.org/browse/STCOM-429).

Image_1

Image_2

Storybook

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Correctly set `<Accordion>`'s `id`. Refs STCOM-551. 
 * `<MetaSection>` elegantly handles missing metadata. Refs STCOM-538.
 * `<Datepicker>` cleanup: better i18n, issues with untouched fields.
+* Add an optional `footer` prop for `Pane` component for fixed footer feature. Refs STCOM-429.
 
 ## [5.4.2](https://github.com/folio-org/stripes-components/tree/v5.4.2) (2019-06-12)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v5.4.1...v5.4.2)
--- a/lib/Pane/Pane.js
+++ b/lib/Pane/Pane.js
@@ -20,6 +20,7 @@ const propTypes = {
   ),
   firstMenu: PropTypes.element,
   fluidContentWidth: PropTypes.bool,
+  footer: PropTypes.element,
   height: PropTypes.string,
   lastMenu: PropTypes.element,
   noOverflow: PropTypes.bool,
@@ -165,6 +166,7 @@ class Pane extends React.Component {
       dismissible,
       firstMenu,
       fluidContentWidth, // eslint-disable-line no-unused-vars
+      footer,
       height, // eslint-disable-line no-unused-vars
       lastMenu,
       noOverflow, // eslint-disable-line no-unused-vars
@@ -204,6 +206,7 @@ class Pane extends React.Component {
         <div key={`${this.id}-content`} className={this.getContentClass()} onFocus={this.setLatestFocused}>
           {children}
         </div>
+        {footer}
       </Element>
     );
   }
--- a/lib/PaneFooter/PaneFooter.css
+++ b/lib/PaneFooter/PaneFooter.css
@@ -1,0 +1,20 @@
+@import '../variables.css';
+
+.paneFooter {
+  overflow: auto;
+  display: flex;
+  justify-content: center;
+  border-top: 1px solid var(--color-border);
+  box-shadow: var(--shadow);
+  min-height: var(--control-min-size-touch);
+  padding: var(--gutter-static);
+}
+
+.paneFooterButtons {
+  width: 50em;
+  display: flex;
+  justify-content: space-between;
+  flex-wrap: wrap;
+  padding-top: var(--gutter-static-one-third);
+  margin-right: var(--gutter-static);
+}
--- a/lib/PaneFooter/PaneFooter.js
+++ b/lib/PaneFooter/PaneFooter.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Button from '../Button/Button';
+
+import css from './PaneFooter.css';
+
+const PaneFooter = ({ children }) => (
+  <div className={css.paneFooter}>
+    <div className={css.paneFooterButtons}>
+      {children}
+    </div>
+  </div>
+);
+
+PaneFooter.propTypes = {
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(Button),
+    PropTypes.objectOf(Button),
+  ]),
+};
+
+export default PaneFooter;
--- a/lib/PaneFooter/readme.md
+++ b/lib/PaneFooter/readme.md
@@ -1,0 +1,30 @@
+# PaneFooter
+Render pane footer at the bottom of the `<Pane>`-component containing a form.
+
+### Usage
+```js
+import { Pane, PaneFooter, Button } from '@folio/stripes/components';
+
+const footer = (
+  <PaneFooter>
+    <Button onClick={() => {...}}>
+      Cancel
+    </Button>
+    <Button
+      buttonStyle="primary"
+      type="submit"
+    >
+      Save
+    </Button>
+  </PaneFooter>
+);
+
+<Pane footer={footer} ... >
+  Pane Content
+</Pane>
+```
+
+### Props
+Name | type | description | default | required
+--- | --- | --- | --- | ---
+children | | Set of `<Button>`s. |  |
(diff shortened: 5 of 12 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Correctly set `<Accordion>`'s `id`. Refs STCOM-551. 
 * `<MetaSection>` elegantly handles missing metadata. Refs STCOM-538.
 * `<Datepicker>` cleanup: better i18n, issues with untouched fields.
+* Fixed footer for Save and Cancel buttons. Refs STCOM-429.
```
YevheniiMaltsev: Please, use Present Simple time for the comments in chengelog

SergiyVSergiyenko: Thanks. It's my fail.

lib/PaneFooter/PaneFooter.js
```diff
@@ -0,0 +1,47 @@
...
+
+import Button from '../Button';
+
+import css from './PaneFooter.css';
+
+class PaneFooter extends Component {
+  static propTypes = {
+    invalid: PropTypes.bool.isRequired,
+    onCancel: PropTypes.func.isRequired,
+    pristine: PropTypes.bool.isRequired,
```
YevheniiMaltsev: As far as understand this  property is used to check if the data was changed. Maybe, it will be better to use names "changed", "hasChanges", "isChanged"?
This is only my thought (not a change request).

lib/PaneFooter/PaneFooter.css
```diff
@@ -0,0 +1,20 @@
+@import '../variables.css';
+
+.paneFooter {
+  width: 100%;
```
VictorSoroka1: When we have `display: flex`, the width should be 100% by default.

lib/PaneFooter/PaneFooter.js
```diff
@@ -0,0 +1,47 @@
...
+    } = this.props;
+
+    return (
+      <div className={css.paneFooter}>
+        <div className={css.paneFooterButtons}>
+          <Button
+            buttonStyle="default mega"
+            onClick={onCancel}
+          >
+            <FormattedMessage id="stripes-components.cancel" />
```
VictorSoroka1: I would suggest moving out the button message out of the component as a prop.

lib/PaneFooter/PaneFooter.js
```diff
@@ -0,0 +1,47 @@
...
+          </Button>
+          <Button
+            buttonStyle="primary mega"
+            type="submit"
+            disabled={pristine || submitting || invalid}
+          >
+            <FormattedMessage id="stripes-components.saveAndClose" />
```
VictorSoroka1: I would suggest moving out the button message out of the component as a prop.

lib/PaneFooter/PaneFooter.js
```diff
@@ -0,0 +1,47 @@
...
+          <Button
+            buttonStyle="default mega"
+            onClick={onCancel}
+          >
+            <FormattedMessage id="stripes-components.cancel" />
+          </Button>
+          <Button
+            buttonStyle="primary mega"
+            type="submit"
+            disabled={pristine || submitting || invalid}
```
VictorSoroka1: I would suggest providing just `disabled` prop and not implementation detail which connected to forms. So instead of providing `pristine || submitting || invalid` we need something more abstract. Please follow the example of Modal footer component.

https://github.com/folio-org/stripes-components/blob/master/lib/ModalFooter/ModalFooter.js

CHANGELOG.md
VictorSoroka1: @SergiyVSergiyenko, as I understand this is not fail as you need `fixed` for the `fixed footer` feature. In order to prevent confusion, please rename things to look like

```
Add an optional `footer` prop for `Pane` component for fixed footer feature. Refs STCOM-429.
```

lib/Pane/Pane.js
```diff
@@ -193,8 +195,8 @@ class Pane extends React.Component {
           paneSub={paneSub}
           appIcon={appIcon}
           firstMenu={firstMenu}
-          lastMenu={lastMenu}
-          actionMenu={actionMenu}
+          lastMenu={footer ? null : lastMenu}
+          actionMenu={footer ? null : actionMenu}
```
JohnC-80: We don't want to disregard the last menu and action menu props if a footer is present - there still may need to be elements there. In most cases migration will be simple... something like
```
const lastMenu = (
  <PaneMenu>
    <Button ...>Cancel</Button>
    <Button>Save</Button>
  </PaneMenu>
);
<Pane lastMenu={lastMenu} ... />
```
to
```
const footer = (
  <PaneFooter>
    <Button ...>Cancel</Button>
    <Button>Save</Button>
  </PaneFooter>
);
<Pane footer={footer} .../>
```
So `<PaneFooter>` can just simply render its children and apply appropriate css. (The last part may be difficult, would love for @rasmuswoelk to chime in here) This way, it can leave all the disabled/submitting/etc props up to the JSX of those children (as I think @VictorSoroka1 has suggested) They're may be instances where we want to have more than just save/cancel buttons here.

lib/PaneFooter/PaneFooter.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Button, { propTypes as buttonPropTypes } from '../Button/Button';
+
+import css from './PaneFooter.css';
+
+const PaneFooter = ({ children, primaryButton, secondaryButton }) => {
```
YevheniiMaltsev: Could you please clarify the next points:
- Why do we support only two additional buttons?
- Do we need to support some extra buttons at all (As fas as I understand there are not any requirements for it now)?

CC: @VictorSoroka1

YevheniiMaltsev: It looks like the buttons props are mentioned here (not buttons components) so please rename it

lib/PaneFooter/tests/interactor.js
```diff
@@ -0,0 +1,12 @@
...
+  interactor,
+  property,
+} from '@bigtest/interactor';
+
+import ButtonInteractor from '../../Button/tests/interactor';
+
+export default interactor(class PaneFooterInteractor {
+  primaryButton = new ButtonInteractor('[data-test-pane-footer-primary-button]');
+  secondaryButton = new ButtonInteractor('[data-test-pane-footer-secondary-button]');
+  disabledPrimaryButton = property('[data-test-pane-footer-primary-button]', 'disabled');
```
YevheniiMaltsev: There are 'primaryButtonDisabled' and "disabledPrimaryButton" names that describe the same thing. It looks like this field should be renamed

lib/Pane/Pane.js
```diff
@@ -20,6 +20,7 @@ const propTypes = {
   ),
   firstMenu: PropTypes.element,
   fluidContentWidth: PropTypes.bool,
+  footer: PropTypes.element,
```
VictorSoroka1: Please update [readme](https://github.com/folio-org/stripes-components/blob/master/lib/Pane/readme.md) file with information about `footer`.

VictorSoroka1: Please add example usage of the `footer` prop in [stories](https://github.com/folio-org/stripes-components/tree/master/lib/Pane/stories).

lib/PaneFooter/stories/BasicUsage.js
```diff
@@ -0,0 +1,33 @@
...
+        buttonStyle="primary"
+        onClick={action('You clicked save')}
+      >
+        Save
+      </Button>
+    </PaneFooter>
+  );
+
+  return (
+    <Pane
```
VictorSoroka1: Wrap with `<Paneset>`.

SergiyVSergiyenko: Thank you. Done. Works.

lib/PaneFooter/readme.md
```diff
@@ -0,0 +1,30 @@
+# PaneFooter
+The default pane footer for the Pane-component containing a form.
```
VictorSoroka1: It is a default pane footer. As if the footer prop is not passed, then there won't be any footer at all. Please come up with another description. Use `PaneMenu` story for the reference.

lib/PaneFooter/stories/BasicUsage.js
```diff
@@ -0,0 +1,36 @@
...
+        onClick={action('You clicked save')}
+      >
+        Save
+      </Button>
+    </PaneFooter>
+  );
+
+  return (
+    <Paneset>
+      <Pane
```
VictorSoroka1: Please add the title for the pane. `paneTitle="Pane with a footer"`.

PR #1026, opened 2019-08-04
UX-325: Updated <Button> border-radius to be circular

rasmuswoelk:
**Ticket:** https://issues.folio.org/browse/UX-325

image


doytch:
@JohnC-80 are you finding this new design a hard pill to swallow?

PR #1029, opened 2019-08-05
STCOM-546: increase ModalFooter code coverage

MaksymDryha:
`ModalFooter` can receive buttons either via `children` or with `primaryButton` and `secondaryButton` props.  `primaryButton` and `secondaryButton` props are deprecated and not covered by tests. If someone is sure that they are not used anywhere and can be removed, please let me know. For now, I'm adding some more tests covering the case when those deprecated props are used.

### DIFF
--- a/lib/ModalFooter/tests/ModalFooter-test.js
+++ b/lib/ModalFooter/tests/ModalFooter-test.js
@@ -15,32 +15,67 @@ describe('ModalFooter', () => {
   const secondaryButtonLabel = 'Secondary Button';
   const secondaryButtonId = 'modalFooterSecondaryButton';
 
-  beforeEach(async () => {
-    await mount(
-      <ModalFooter>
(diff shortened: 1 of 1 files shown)


zburke:
{{primaryButton}} and {{secondaryButton}} are [in use](https://github.com/folio-org/ui-users/blob/master/src/components/AddServicePointModal/AddServicePointModal.js#L72-L82), though it would be easy to fix this. 

Even if they weren't, however, removing them would be a major-version change.

Godlevskyi:
> `primaryButton` and `secondaryButton` are [in use](https://github.com/folio-org/ui-users/blob/master/src/components/AddServicePointModal/AddServicePointModal.js#L72-L82), though it would be easy to fix this.
> 
> Even if they weren't, however, removing them would be a major-version change.

@zburke, this PR is not about removing deprecated functionality but to cover with tests to increase the coverage to 80%.

PR #1038, opened 2019-08-16
STCOM-554 apply ref correctly to avoid React stacktrace

zburke:
When the pane was narrow, the ref was incorrectly created on the component
rather than the DOM element. `<MultiSelectFilterField>` expects to receive
`props.inputRef`, but `<MultiSelectOptionsList>` was passing `props.ref` 
instead.

Refs [STCOM-554](https://issues.folio.org/browse/STCOM-554)

### DIFF
--- a/lib/MultiSelection/MultiSelection.js
+++ b/lib/MultiSelection/MultiSelection.js
@@ -393,14 +393,14 @@ class MultiSelection extends React.Component {
               ...filterResults,
             };
 
-            const handleControlClick = (e) => {
+            const handleControlClick = () => {
               if (disabled) {
                 return;
               }
 
               toggleMenu();
               if (!isOpen && this.input.current) {
-                e.target.focus();
+                this.input.current.focus();
               }
             };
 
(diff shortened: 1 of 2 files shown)


doytch:
I don't fully understand how this solves the problem since I didn't really understand the problem from the start, but I'm here for it if it fixes it.

If possible, can you explain _why_ using a `ref` was failing?

lib/MultiSelection/MultiSelection.js
```diff
@@ -393,14 +393,14 @@ class MultiSelection extends React.Component {
...
-            const handleControlClick = () => {
+            const handleControlClick = (e) => {
               if (disabled) {
                 return;
               }
 
               toggleMenu();
               if (!isOpen && this.input.current) {
-                this.input.current.focus();
+                e.target.focus();
```
mkuklis: @zburke should `toggleMenu` pass `e` here?

JohnC-80:
The ref is actually working as refs are supposed to, it's just applied to a component rather than the DOM element itself.

@zburke I think you mentioned it in the JIRA comment - line 137 of MultiSelectOptionList.js
`ref: inputRef` just needs to be changed to `inputRef` and the component will properly catch it/assign to the `<input>` where it needs to be. 

Assigning `ref` as a prop to the component is returning the component instance, which has no `focus()` method, so it bombs out.

PR #1046, opened 2019-09-02
STCOM-571: Make PaneFooter flexible enough

SergiyVSergiyenko:
### Purpose
Support the ability to add CSS properties to the PaneFooter via the parent component according the [story](https://issues.folio.org/browse/STCOM-571).

**Screenshot**
View_6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Display `<Button>`s as pills (circular/rounded sides). Refs STCOM-564.
 * Correctly pass `ref`s to `<MultiSelection>` internal components. Refs STCOM-554.
 * Better `<KeyValue>` test infrastructure.
+* Update `<PaneFooter>`: consolidate into a single div with adding CSS classes. Refs STCOM-521.
 
 ## [5.5.0](https://github.com/folio-org/stripes-components/tree/v5.5.0) (2019-07-22)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v5.4.2...v5.5.0)
--- a/lib/PaneFooter/DefaultPaneFooter.js
+++ b/lib/PaneFooter/DefaultPaneFooter.js
@@ -1,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import css from './DefaultPaneFooter.css';
+
+const DefaultPaneFooter = ({ renderStart, renderEnd, className }) => (
+  <div className={classNames(css.paneFooterContent, className)}>
+    <span
+      data-test-pane-footer-start
+      className={css.renderStart}
+    >
+      {renderStart}
+    </span>
+    {renderEnd && <span data-test-pane-footer-end>{renderEnd}</span>}
+  </div>
+);
+
+DefaultPaneFooter.propTypes = {
+  className: PropTypes.string,
+  renderEnd: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.arrayOf(PropTypes.node),
+  ]),
+  renderStart: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.arrayOf(PropTypes.node),
+  ]),
+};
+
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Display `<Button>`s as pills (circular/rounded sides). Refs STCOM-564.
 * Correctly pass `ref`s to `<MultiSelection>` internal components. Refs STCOM-554.
 * Better `<KeyValue>` test infrastructure.
+* Make `<PaneFooter>` flexible enough. Refs STCOM-521.
```
YevheniiMaltsev: I'm not sure that 'flexible enough' describes the changes meaning. Please add a little bit more informative message.

lib/PaneFooter/PaneFooter.css
```diff
@@ -1,20 +1,10 @@
...
-}
-
-.paneFooterButtons {
-  width: 50em;
-  display: flex;
-  justify-content: space-between;
-  flex-wrap: wrap;
-  padding-top: var(--gutter-static-one-third);
-  margin-right: var(--gutter-static);
+  min-height: calc(var(--control-min-size-touch) * 1.3);
```
YevheniiMaltsev: Could you please clarify what does it mean "1.3" and why we use this value here?

lib/PaneFooter/PaneFooter.js
```diff
@@ -1,23 +1,22 @@
...
+    {children}
   </div>
 );
 
 PaneFooter.propTypes = {
   children: PropTypes.oneOfType([
-    PropTypes.arrayOf(Button),
-    PropTypes.objectOf(Button),
-  ]),
+    PropTypes.node,
```
doytch: `PropTypes.string` is part of `PropTypes.node`. `node` is anything renderable.

Also, does passing a function actually work? You're not evaluating the function in the `PaneFooter`.

lib/PaneFooter/PaneFooter.js
```diff
@@ -1,23 +1,20 @@
...
 
 import css from './PaneFooter.css';
 
-const PaneFooter = ({ children }) => (
-  <div className={css.paneFooter}>
-    <div className={css.paneFooterButtons}>
-      {children}
-    </div>
+const PaneFooter = ({ children, footerClass, ...rest }) => (
+  <div className={classNames(css.paneFooter, footerClass)}>
```
rasmuswoelk: Could we use `className` instead of `footerClass` here?

lib/PaneFooter/PaneFooter.js
```diff
@@ -1,23 +1,20 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 
-import Button from '../Button/Button';
+import DefaultPaneFooter from './DefaultPaneFooter';
 
 import css from './PaneFooter.css';
 
-const PaneFooter = ({ children }) => (
```
rasmuswoelk: Can you add React's `forwardRef` here so that developers can pass refs for the footer?

And it would be nice if developers could decide the root element of the component. E.g. if they need it to be a `<form>` instead of a div.

You can do something like:
```js
const PaneFooter = forwardRef(({ ..., element: Element }, ref) => (<Element ref={ref}....</Element>));
```

And then set a default prop for `element` that is a string of `div`.

Note: When using forwardRef, you need to set the display name before exporting:

PaneFooter.displayName = 'PaneFooter';

(Otherwise, it will show up in react inspect as "forwardRef()...")

lib/PaneFooter/DefaultPaneFooter.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import css from './DefaultPaneFooter.css';
+
+const DefaultPaneFooter = ({ renderStart, renderEnd, innerClassName }) => (
```
rasmuswoelk: `innerClassName` could just be `className` here

PR #1048, opened 2019-09-03
UIIN-199 allow lodash templates in FilterGroup

zburke:
The `cql` attribute of a filter simply corresponds to the field the
value must match, e.g. given a `cql` of `foo`, the generated query will
be
```
foo=("value1", "value2", ... "valueN")
```
In order to support more sophisticated substitutions, providing a
`template` attribute, as an alternative to `cql`, will use lodash's
`template` function to handle value substitution into the generated
query.

Refs [UIIN-199](https://issues.folio.org/browse/UIIN-199)


zburke:
This is an abandoned POC.

PR #1049, opened 2019-09-04
STCOM-567: Add <Tooltip> component

rasmuswoelk:
In this PR I have added a new component: `<Tooltip>`. This work is still in progress. I will add documentation and tests when I'm closer to the final result.

The tooltip is meant to provide additional information for the user when hovering or focusing some parts of a UI. This can be useful for providing context to the user for e.g. icon buttons that don't have a visible label attached.

**I have focused on:**
- Developer experience – how easy it is for a developer to implement
- Accessibility – making sure that the information inside the tooltip is available for all users
- User experience – making sure that the tooltip isn't an annoyance for users as they operate a FOLIO app

## Component API
The recommended way of using the component is to pass the trigger as a child using the render-prop pattern. The props needed for enabling the tooltip will be passed as the first parameter and can be spread directly onto the trigger element.

The props that will be passed:
- `ref` (required for positioning and more)
- `ariaIds` (required to support announcement for screen reader users. Should be applied either as `aria-describedby` and/or `aria-labelledby`

```js
        <Tooltip
          text="Delete record"
          sub="CMD+D"
          id="tooltip-example-2"
        >
          {({ ref, ariaIds }) => (
            <IconButton
              icon="trash"
              ref={ref}
              aria-labelledby={ariaIds.text}
              aria-describedby={ariaIds.sub}
            />
          )}
        </Tooltip>
```

## Screenshots
image

image

## See it in action
The tooltip can be seen on https://folio-styling.testing.samhaeng.com in modules that use the `<SearchAndSort>` (Users, Inventory, etc.) when toggling the filter pane.


doytch:
I really like this! Couple of corner-cases:

-  Would it be possible to use [Option 1](https://github.com/folio-org/stripes-components/pull/1049/files#diff-42559da61ddeea9e2ebcc24e63eaf67dR22) to show multiple tooltips off of a single ref hover?
- Would we be able to build out those "tutorial tooltips" workflows with this? Sort of more programmatic, where once a user has done what they were instructed to do (eg, triggered an `onClick`), one tooltip is hidden and another is shown?

I really don't care too much about those, they just came to mind right now as things we may want eventually.

rasmuswoelk:
Hey @doytch! Thanks for your questions :)

1. I just tested it and it looks like the latter will take precedence if there are multiple tooltips attached to a single trigger ref.

2. We could potentially extend the tooltip component for this purpose. It would require passing a controlled open-state to the component and we would probably also have to do some work in regard to screen reader users – something like using aria-live to announce the tutorial tooltip's content.

PR #1054, opened 2019-09-19
Bug fix for better row widths.

JohnC-80:
Sometimes the striping of MCL rows can render too short or too long... sometimes there are unneccessary horizontal scrollbars.

This PR cleans up the rowWidth logic by memoizing a function being passed the columnWidths. It ensures that the rowWidth is equal to the greater value of the total of ColumnWidths and the width of the containing element.


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -672,11 +680,15 @@ class MCLRenderer extends React.Component {
...
       columns,
+      prevWidth,
       columnWidths,
       rowWidth
     } = this.state;
 
-    const defaultStyle = (rowWidth) ? { minWidth: `${rowWidth}px` } : null;
+    const widthVar = prevWidth || width || 0;
+    const rowWidthVar = rowWidth || 0;
+    const rowMargin = 5;
```
doytch: This feels a tad magic-numbery. If it's coming from the CSS, can you note it in a comment?

PR #1057, opened 2019-09-20
UIOR-387 wrap long text value

aliaksei-chumakou:
*Purpose*:
Fix for https://issues.folio.org/browse/UIOR-387
I think it's a good idea to wrap any long text value
fixed

### DIFF
--- a/lib/KeyValue/KeyValue.css
+++ b/lib/KeyValue/KeyValue.css
@@ -13,3 +13,7 @@
   margin-bottom: var(--control-margin-bottom);
   font-size: var(--font-size-medium);
 }
+
+.kvValue {
+  overflow-wrap: break-word;
+}
--- a/lib/KeyValue/KeyValue.js
+++ b/lib/KeyValue/KeyValue.js
@@ -22,7 +22,10 @@ function KeyValue(props) {
       <div className={css.kvLabel}>
         {props.label}
       </div>
-      <div data-test-kv-value>
+      <div
+        className={css.kvValue}
+        data-test-kv-value
+      >
         {props.children || props.value}
       </div>
     </div>


aliaksei-chumakou:
@zburke we do this already in ui-organizations, ui-orders, but I think we need to fix this across platform. To me it looks like a common issue within KeyValue and a very long value without whitespaces. Please look at the ticket

aliaksei-chumakou:
@zburke in any case, we can end up with acq component, like AcqKeyValue, that includes the fix, somewhere in stripes-acq-components. What do you think?

zburke:
I did look at the ticket. It describes a problem specific to the ui-orders module. By changing this at the component level, however, you make this change apply to all modules, hence my raising the issue with the component developers. 

I understand you are keen to get that merged quickly and why you want to implement the fix at the stripes-components level in one place instead of having to change the layout in several apps. It's good to think about it that way, where can we make the smallest change with the greatest benefit. But I really think we need feedback from the component developer, @JohnC-80, in that case. He’s stuck in a meeting right now. Hopefully we get some input from him when he is out. If you can’t wait, I would understand if you want to move ahead with something like `AcqKeyValue`. Your call.

aliaksei-chumakou:
No hurry, I can wait forever, this bug was always there) Also, I don't think it's a local ui-orders problem. at least https://issues.folio.org/browse/UIORGS-2. Do you have any idea what could be wrong in layout, so this fix isn't required?

JohnC-80:
Thank you, @zburke for covering this... 
I think `overflow-wrap` is suitable (prefer `word-break` for legacy sake as both are similar/supported in similar ways) to keep the layout intact and resolve the issue displayed in the screenshot on https://issues.folio.org/browse/UIOR-387.
Seems like it would also nicely conform when there are spaces/applicable breaking characters present and just handle the lengthy words.

Despite this being a widely used component, this is also has a low impact in the codebase - it's easy to remove/revise if it causes more issues than it fixes. If it causes more issues, this can also be adjusted via props as an optional behavior (or default, depending on how many issues/how isolated the issues are that it causes.)

Really want further chip-in from @rasmus, @filipjakobsen to collect all the opinions before merging this.

JohnC-80:
It's difficult to assume what sort of content/length of content a field will contain. If the value is arbitrary and without any true system-wide convention, you can expect lengthy strings, possibly non-breaking characters... but it's tough to judge layout with this since it can result in inefficient use of horizontal space when values are very short.

rasmuswoelk:
Hey guys!

So.. I looked at the ticket, the code, the result, and all of you guys' comments.

First of all; it's clear that something needs to be done regarding those long strings. Either by updating the `<KeyValue>` component or by changing the implementation entirely. Or both.

I agree with @JohnC-80 and @zburke that we need to be careful with making this change system-wide because of the potential implications that it might have since the word breaking can get a bit arbitrary. But I also agree that it's pretty low-impact code-wise so it would be very easy to change it back if it caused any major issues.

@zburke's suggestion on using col/row to make the columns wider/narrower depending on the data that will be rendered is also a good way to prevent this issue. But the word wrapping would probably still be needed at some point – especially on small screens.

For the `<KeyValue>` specifically, my suggestion would be to add a prop that activates word-break so that it's opt-in. That way, developers could add it when they know it is/could be a necessity.

Another option for these long values could be to render the value inside a read-only and full width `<TextField>` – making it easy to copy/paste the string if needed. The text field would then take up the available space.

aliaksei-chumakou:
Guys, thanks for your input!
@rasmuswoelk  @zburke regarding row-col usage, even if we make it one column per row, it's still could be an issue if value more than provided screen space. So I don't think it's an solution.
More proper solution might be a common solution for fading out/ellipsis of lengthy values, but with some popover (on hover or click) with full value. Of course, this requires to set up requirements and ux designer's input. What do you think?
Current fix is a try to spread our experience, and maybe kind of ad-hoc solution, because we have to display all the value to user and make it available to select and copy. Hiding this behavior behind the additional prop doesn't look like a good idea to me, since there a lot of undiscovered places where could be a lengthy value and, IMHO, it's the same as to put just css class there, but without touching common component.
It looks like at least Rasmus is agree that it's a common issue to handle. Does it make sense to discuss fading out/ellipsis common solution then?

JohnC-80:
As I mentioned, it's difficult to really know which fields will truly need this in order to build the appropriate layout or apply a prop.
Ellipsis or other forms of truncation have accessibility concerns as these techniques may hide potentially important data.

I think we can merge this and just be vigilant of potential layout impact... if we find serious problem, revisit with a prop later.

After some review of the `platform-core` apps with this change, it seems to work better more often than not:
Before:
image

After
image

PR #1058, opened 2019-09-23
STCOM-581 react-router-dom

zburke:
Update `react-router-dom` to v5 and move it from dependencies to 
peerDependencies in order to guarantee that all modules in a 
platform are on the same version.

Refs [STCOM-581](https://issues.folio.org/browse/STCOM-581)


zburke:
Replaced by #1321

PR #1059, opened 2019-09-23
Tooltip: Loosen text/sub proptypes

doytch:
While reviewing https://github.com/folio-org/ui-agreements/pull/289, it looks like the `Tooltip` component handles React nodes for text/sub perfectly, but the proptypes checker complains on the console about failed proptypes.

This loosens the proptypes allowed so that we can easily pass in `FormattedMessage` without having to resort to its render-props form.


JohnC-80:
For a11y's sake we do want these things to remain simple and not include complex layouts of elements. `node` might be a bit much.
 
We may need a custom type to fit this to either string or an instance of `<FormattedMessage>` that suites `react-hot-loader`

doytch:
Yeah...makes sense. Or is not really desired and we shouldn't loosen it at all, forcing the render-props version of FormattedMessage?

JohnC-80:
I could live with a specific proptype there because I think it would be a little unweildy to wrap multiple `<FormattedMessages>` for both `text` and `sub` props... even `IntlProvider` feels weird for just that (but can also be used.)

JohnC-80:
Going to take a *quick moment to try and appease the cloud on this one, then I'll merge this.

package.json
```diff
@@ -111,6 +111,7 @@
     "tai-password-strength": "^1.1.1"
   },
   "peerDependencies": {
-    "@folio/stripes-core": "^3.0.0"
+    "@folio/stripes-core": "^3.0.0",
+    "react-hot-loader": "^4.0.0"
```
zburke: @JohnC-80, @doytch, I just noticed this peerDep in here. `platform-core#snapshot` doesn't include RHL at all; the `#master` branch does include it but only in `resolutions`. Are we confident it needs to be a peer? If yes, we should open PRs to add it in each of the platform repos.

doytch: It should be provided by `stripes-core`. The only reason we need to do this rigamarole is bc `stripes-core` uses the hot-loader in its webpack dev server builds.

PR #1073, opened 2019-10-10
Add <FormattedUTCDate>

MikeTaylor:
Copies of the same `<FormattedUTCDate>` exist in several apps, including ui-licenses and ui-shipping. Rather than add yet another copy in ui-courses, this PR adds it to stripes-components so we only have to maintain it once.


lib/FormattedUTCDate/readme.md
```diff
@@ -0,0 +1,12 @@
+# FormattedUTCDate
+
+Displays an data only -- with no time component -- in UTC.
```
doytch:
```suggestion
Displays a date only -- with no time component -- in UTC.
```

lib/FormattedUTCDate/readme.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Usage
+
+When you would normally use `react-intl`'s `<FormattedDate>` to display a date field interpreted according to the local timezone, `<FormattedUTCDate>` can instead be used to render it interpreted as UTC -- i.e., showing precisely what is actually stored in the back-end. This is appropriate when recording only a date without an associated time, as for example the expiry date of a loan. For this, UTC is used both for storage and display, so that the displayed date does not vary when crossing timezones.
```
doytch:
```suggestion
When you would normally use `react-intl`'s `<FormattedDate>` to display a date field interpreted according to the local timezone, `<FormattedUTCDate>` can instead be used to render it interpreted as UTC -- i.e., showing precisely what is actually stored in the back-end. This is appropriate when recording only a date without an associated time, eg a birthdate. For this, UTC is used both for storage and display, so that the displayed date does not vary when crossing timezones.
```

PR #1079, opened 2019-10-22
STCOM-611 receive correct prop-type for scope

zburke:
`PropTypes.object` is correct for receiving `document.body`.
Interestingly, `PropTypes.node` does not include `object`.

Refs [STCOM-611](https://issues.folio.org/browse/STCOM-611)


zburke:
Duplicates #1077

PR #1081, opened 2019-10-23
STCOM-613 add placeholder and translate it

aliaksei-chumakou:
Here are two reasons for this PR:
* to remove default non-translated placeholders for all apps if any search index selected, it's a old known issue;
* feature request in the ticket to make these placeholders custom.

1. if placeholder prop is provided to SearchField - it will be displayed always, no matter of selected index;
2. if placeholder prop isn't provided to SearchField - we look for this prop in selected index config object. If it exists - it will be displayed as is, so any app that needs to use it should apply the same pattern as for translation of labels for SearchAndSort.

Any comments/suggestions are welcome, thank you in advance

### DIFF
--- a/lib/SearchField/SearchField.js
+++ b/lib/SearchField/SearchField.js
@@ -91,7 +91,7 @@ const SearchField = (props) => {
   let inputPlaceholder = placeholder;
   if (!placeholder && hasSearchableIndexes && selectedIndex) {
     const selectedIndexConfig = searchableIndexes.find(index => index.value === selectedIndex) || {};
-    inputPlaceholder = selectedIndexConfig.placeholder;
+    inputPlaceholder = selectedIndexConfig.placeholder || '';
   }
 
   return (
--- a/lib/SearchField/SearchField.js
+++ b/lib/SearchField/SearchField.js
@@ -7,6 +7,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { injectIntl, intlShape } from 'react-intl';
+
 import TextField from '../TextField';
 import Select from '../Select';
 import TextFieldIcon from '../TextField/TextFieldIcon';
@@ -20,6 +22,7 @@ const propTypes = {
   id: PropTypes.string,
   inputClass: PropTypes.string,
   inputRef: PropTypes.object,
+  intl: intlShape,
   loading: PropTypes.bool,
   onChange: PropTypes.func,
   onChangeIndex: PropTypes.func,
@@ -49,6 +52,7 @@ const SearchField = (props) => {
     selectedIndex,
     searchableIndexesPlaceholder,
     inputClass,
+    intl: { formatMessage },
     ...rest
   } = props;
 
@@ -83,7 +87,15 @@ const SearchField = (props) => {
   const searchIcon = (<TextFieldIcon iconClassName={css.searchIcon} icon="search" />);
 
   // Placeholder
-  const inputPlaceholder = (hasSearchableIndexes && selectedIndex) ? `Search for ${selectedIndex}` : placeholder;
+  let inputPlaceholder = placeholder;
+  if (hasSearchableIndexes && selectedIndex) {
+    const selectedIndexConfig = searchableIndexes.find(index => index.value === selectedIndex) || {};
+    const placeholderTranslationId = selectedIndexConfig.placeholderId || 'stripes-components.searchField.placeholder';
+    inputPlaceholder = formatMessage(
+      { id: placeholderTranslationId },
+      { label: selectedIndexConfig.label },
+    );
+  }
 
   return (
     <div className={rootStyles}>
@@ -110,4 +122,4 @@ const SearchField = (props) => {
 
 SearchField.propTypes = propTypes;
 
-export default SearchField;
+export default injectIntl(SearchField);
--- a/lib/SearchField/SearchField.js
+++ b/lib/SearchField/SearchField.js
@@ -7,7 +7,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { injectIntl, intlShape } from 'react-intl';
 
 import TextField from '../TextField';
 import Select from '../Select';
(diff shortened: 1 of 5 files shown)


aliaksei-chumakou:
@mkuklis do you mean `placeholder` prop of `SearchField` component? It's already exists, and it steps in if no search index is selected. I don't want to change existing logic, to reduce any side affects, so I don't see any easy solution to use `placeholder` here... If I understand correctly, of course.

taras-tkachenko:
@mkuklis we need to define its own placeholder for each search index jus the way @aliaksei-chumakou has done this.

JohnC-80:
I'm okay with this, but just to explore possibilities - Rather than engage in `react-intl` context within this component, could we optionally accept a function for placeholder here?
```
placeholder=(searchIndex)=> ...whatever your intl logic can produce? 
```
Thoughts?

aliaksei-chumakou:
@JohnC-80  is it so bad to couple on react-intl in stripes-components? The way with function requires a lot more code. Also it doesn't fix the existing issue - default placeholder will not be translated.

JohnC-80:
Not necessarily bad, just seems like everything you'd need to get a translated string would already exist outside of this component with all the props passed to the searchableIndex functionality.
```
<IntlConsumer>
  { intl =>
    <SearchField
      SelectedIndex={selectedIndex}
      onChangeIndex={ ... }
      placeholder={(index) => intl.formatMessage({id:`searchIndex-${selectedIndex}`})}
    />
  }
</IntlConsumer>
```

aliaksei-chumakou:
@JohnC-80 but for now `placeholder` is a string, and we have to support this in case it's used already. So we have to check if type of `placeholder` is a function - then call fn. If not - show the buggy not translated label, as before, right?

JohnC-80:
Good point - for strings, are we not able to structure the intl logic in a similar way?
```
<IntlConsumer>
  { intl =>
    <SearchField
      SelectedIndex={selectedIndex}
      onChangeIndex={ ... }
      placeholder={intl.formatMessage({id:`searchIndex-${selectedIndex}`})}
    />
  }
</IntlConsumer>
```

aliaksei-chumakou:
@JohnC-80 right, it works, but requires to change all the places where `SearchField` is used with `placholder` as a value (not the translation key)

mkuklis:
What I had in my original comment is exactly what @JohnC-80 proposed. It would be a breaking change as @aliaksei-chumakou mentioned tough.

zburke:
Is there some circumstance you face that prevents this data from being translated ahead of passing it to `<SearchAndSort>`? For example, in ui-inventory, we [stash translation-keys in the `label` key](https://github.com/folio-org/ui-inventory/blob/3cccee2bae135064448f451c645f04c1082fd22e/src/constants.js#L37-L47):
```
export const searchableIndexes = [
  { label: 'ui-inventory.search.all', value: 'all', ... },
  { label: 'ui-inventory.barcode', value: 'item.barcode', ... },
  { label: 'ui-inventory.instanceId', value: 'id', ... },
  ...
```
and then [translate them before passing to SAS](https://github.com/folio-org/ui-inventory/blob/master/src/Instances.js#L381-L386):
```
    const formattedSearchableIndexes = searchableIndexes.map(index => {
      const { prefix = '' } = index;
      const label = prefix + intl.formatMessage({ id: index.label });

      return { ...index, label };
    });

  <SearchAndSort
    searchableIndexes={formattedSearchableIndexes}
    ...
```
Would implementing a similar pattern eliminate the need for this work?

JohnC-80:
If nothing else `Search for - ` has to be translated somehow anyways. Hate that it's gone this long in that condition.

taras-tkachenko:
We originally wanted this to be configured like the following:
`
const indexes = [{
    label: 'fileName',
    value: `${LOGS_FILTER} AND fileName`,
    placeholder: 'Search by file name'
  }, {
    label: 'hrId',
    value: `${LOGS_FILTER} AND hrId`,
    placeholder: (<FormattedMessage id="ui-data-import.searchByHrid" />)
  }, {
    label: 'uuId',
    value: `${LOGS_FILTER} AND id`,
  }];
`

doytch:
So, couldn't we just avoid the whole `placeholderId` thing with something like:
```
<FormattedMessage id={selectedIndex ? "stripes-components.searchForIndex" : "stripes-components.search"} values={{ selectedIndex }}>
  {placeholder => (
    <TextField
      ...
      placeholder={placeholder}
    />
  }
</FormattedMessage
```
```
"stripes-components.search": "Search",
"stripes-components.searchForIndex": "Search for {selectedIndex}",
```

The examples given on the JIRA issue seem to basically just sub in the selected index to the existing message, so wouldn't the above handle what we're looking to do here? Or are those examples contrived and do you want to do something where the placeholder contains text that is _not_ the value of `selectedIndex`.

taras-tkachenko:
@doytch Exactly so!
We need to be able to add completely custom placeholders to each of search indexes defined in the config.

doytch:
Separate from that customisation though...yes, the default message definitely should be translated. 😆

doytch:
Like, what if we just fixed `SearchField` so it's like this:
```
  return (
    <div className={rootStyles}>
      {searchableIndexesDropdown}
      <FormattedMessage id="stripes-components.searchForIndex" values={{ selectedIndex }}>
        {defaultPlaceholder => (
          <TextField
            {...rest}
            ariaLabel={ariaLabel}
            clearFieldId={clearSearchId}
            focusedClass={css.isFocused}
            id={id}
            hasClearIcon={typeof onClear === 'function' && loading !== true}
            inputClass={classNames(css.input, inputClass)}
            loading={loading}
            onChange={onChange}
            onClearField={onClear}
            placeholder={placeholder || (selectedIndex ? defaultPlaceholder : '')}
            startControl={!hasSearchableIndexes ? searchIcon : null}
            type="search"
            value={value || ''}
          />
        )}
      </FormattedMessage>
    </div>
  );
```
```
"searchForIndex": "Search for {selectedIndex}"
```

1. If you don't want a placeholder and don't pass in a selectedIndex, it defaults to an empty string. This maintains non-breaking functionality with existing modules.
1. If you pass a selectedIndex but no placeholder, it defaults to a translated, default placeholder.
1. If you pass in a placeholder, it'll _always_ use that, as it always should've. This behaviour would be new but IMO is a bugfix and not a breaking change.

That last case is the behaviour it seems you want in your apps.

JohnC-80:
`<SearchField>` was too smart in the first place. I think we should just pass placeholder through so that it works in the way suggested and apps can have the level of control they want rather than manipulate the API in any way. It's 'breaking' in the sense of how the component currently works, but the real case, this is a bug-fix... both simplicity under the hood and complete customization without the assumed `Search for ` which could be very weird with pluralization etc anyhow.

aliaksei-chumakou:
@taras-tkachenko could you please elaborate more on business requirements? Are these custom placeholders crucial to have? Maybe it's a good idea to leave one translated placeholder or nothing.

aliaksei-chumakou:
@JohnC-80 I agree to make `SearchField` less "smart", so I compiled Mark's solution with your idea and Zak's patter:

1. if `placeholder` prop is provided to `SearchField` - it will be displayed always, no matter of selected index;
2. if `placeholder` prop isn't provided to `SearchField` - we look for this prop in selected index config object. If it exists - it will be displayed as is, so any app that needs to use it should apply pattern that Zak described above, the same as for translation of labels.

taras-tkachenko:
Great job!
Suits me well.

aliaksei-chumakou:
thanks for review guys

PR #1084, opened 2019-10-29
STCOM-320 Convert Dropdown to use Popper

JohnC-80:
# Purpose
Transitioning the library to use the more efficient `Popper.js` library over React-tether. This inefficiency was illustrated by [CHAL-56](https://issues.folio.org/browse/CHAL-56?jql=project%20%3D%20CHAL%20AND%20text%20~%20%22Loans%20list%22) Where displaying a high number of dropdowns bogged down the UI.

## Changes
This will eventually amount to a breaking change... currently, the Tether version should only be rendered if the `tether` prop of Dropdown is used.
## Approach
I took the clean hooks-based 'Popdown' component and defaulted to rendering it within `<Dropdown>` if no `tether` prop is used. Tests ensure that both API's currently function, but it would be great to get everyone moved over to using the `renderTrigger` prop rather than the `data-role` attributes on the children.
Most implementations that I've seen will place the children in different areas of the code anyhow.
This transition will ultimately result in a far simpler component.
This PR is gross but temporary, and gets us working with the more efficient code sooner.
The bulk of code within `Dropdown` is there just to make the old patterns behave nicely with new patterns. There are a few console warnings too...

This deprecates the use of the `data-role="toggle"` and `data-role="menu"` attributes to designate the trigger and menu elements. Instead, they're provided with separate render functions.

This also antiquates use of local state for managing the open/closed status of the component... so get rid of those toggle handlers and lose that `ddOpen` key if there's no reason for them outside "have to". `<UncontrolledDropdown>` has a deprecation warning as well. It will be removed entirely.

## To-do
- [x] update PaneHeader to use the new API.
- [x] update `stripes-core` to use the new API.
- [x] add RTL support to `<Popper>`
- [x] docs
- [x] tests


lib/Dropdown/Dropdown.js
```diff
@@ -54,12 +60,18 @@ class Dropdown extends React.Component {
...
+    this._menu = React.createRef();
+    this._renderTrigger = this._renderTrigger.bind(this);
+    this.renderTetherChildren = this.renderTetherChildren.bind(this);
   }
 
   componentDidUpdate(prevProps) {
     // If the dropdown was just closed and the dropdown menu had focus, focus on the dropdown toggle button
     if (!this.props.open && prevProps.open) {
-      if (this._menu && this._menu.node && this._menu.node.contains(document.activeElement)) {
+      if (this._menu.current && this._menu.current.contains(document.activeElement)) {
```
rasmuswoelk: Got this error when trying out the "Toggleable" example in Storybook:

image

lib/Button/Button.js
```diff
@@ -36,7 +36,7 @@ const defaultProps = {
   type: 'button',
 };
 
-function Button(props) {
+const Button = React.forwardRef((props, ref) => {
```
rasmuswoelk: I have found that when using `forwardRef` you should define a `displayName` for the component. Otherwise, it will look like this in when debugging:

image

JohnC-80: Good grief. >.<

And good catch! 👍

lib/Dropdown/Dropdown.js
```diff
@@ -47,64 +61,134 @@ const defaultProps = {
...
   },
+  usePortal: true,
 };
 
-class Dropdown extends React.Component {
-  constructor(props) {
-    super(props);
-    this.handleToggle = this.handleToggle.bind(this);
-    this.handleKeyDown = this.handleKeyDown.bind(this);
+function menuHandleKeyDown(e, open, disabled, onToggle, menuRef, triggerRef, onClose, focusLast, focusFirst) {
```
doytch: This amount of parameters sets off warning lights for me about whether we can take fewer params and has them structured as an object. Mostly for ease of making changes in the future.

lib/Dropdown/Dropdown.js
```diff
@@ -47,64 +61,134 @@ const defaultProps = {
...
+      onClose(e);
+      break;
+    default:
   }
+}
 
-  handleKeyDown(e) {
-    if (this.props.disabled) {
-      return;
+function triggerHandleKeyDown(e, open, disabled, onToggle, menuRef, onClose, focusLast) {
```
doytch: Same comment as `menuHandleKeyDown` about param count.

lib/Dropdown/Dropdown.js
```diff
@@ -47,64 +61,134 @@ const defaultProps = {
...
+    if (this._isLegacy && this.props.open && prevProps.open !== this.props.open) {
+      if (this._menu.current && this._toggle.current === document.activeElement) {
+        const menuContainer = this._menu.current.node || this._menu.current;
+        const elem = getNextFocusable(menuContainer, true, true);
+        if (elem) elem.focus();
+      }
     }
+  }
+
+  _handleTriggerKeys = (e) => {
```
doytch: What's with all the `_foo` method names? I've only really seen this used in the past when supporting refs to a React component, allowing a parent component to call the methods of the React component.

lib/Dropdown/Dropdown.js
```diff
@@ -134,35 +230,48 @@ class Dropdown extends React.Component {
...
           }
-          return cloneElement(child, {
-            open,
-            pullRight,
-            labelledBy: id,
-            className,
-            onToggle: this.handleToggle,
-            ref: isStatefulComponent(child) ? (c) => { this._menu = c; } : undefined,
-          });
+          if (!this._warned) {
```
doytch: I can't tell 100%, but would it be possible to use `deprecated` from `prop-types-extra` instead for this stuff? Eg, [`<Headline>`](https://github.com/folio-org/stripes-components/blob/4a9326534b69ec90db6c6751a41ffa27d32731a2/lib/Headline/Headline.js#L59) but with a custom prop-type validator function that only kicks off `deprecated` if `data-role` is provided? Just spitballing, really not sure.

lib/Dropdown/Dropdown.js
JohnC-80: They can do that anyways through refs for class components. No real convention I'm settling on yet, but it is convenient for differentiation. Here it just helped me differentiate functions/variables created in the Dropdown vs those in `<Popdown>` In Dropdown, those won't live through the breaking change since the guts of it will be prettymuch replaced by `<Popdown>` internals. Lots of temporary code here.

lib/Dropdown/Dropdown.js
JohnC-80: 👍

lib/Dropdown/Dropdown.js
JohnC-80: the data-role props can be added to arbitrary things/children, so it's kinda tough to wrap those up in a different way.
I could use `prop-types-extra` on `onSelect` and `onSelectItem` though, as those have no defaults applied.

JohnC-80: Eh... I don't know.. no incredible benefits over what I've got. would rather fight a single eslint max char warning than multiple.

VictorSoroka1:
I found that their behavior when upon clicking on the area which is not clickable the dropdown is getting closed. Is it correct behavior? I think that this behavior is not introduced by the refactoring though and exists now.
chrome-capture (1)

JohnC-80:
Thanks @VictorSoroka1 I believe I have resolved this now.
The new code will toggle the dropdown if focus goes outside of the menu/trigger. The legacy implementation which I ported over at first was to use `<RootCloseWrapper>` from `react-overlays`. Removing that was the main part of the solution.
In the current state, you'll see the browser's focus styling (an outline of some sort) move to the menu's container, but the menu shouldn't close if you click within white space within the menu.

VictorSoroka1:
@JohnC-80, thank you very much.

lib/Dropdown/Dropdown.css
```diff
@@ -27,6 +36,7 @@
   border: 1px solid #ccc;
   border-radius: var(--radius);
   box-shadow: var(--shadow);
+  pointer-events: all;
```
VictorSoroka1: Could you please clarify the need for this? I am just really curious about the practical side of it.

JohnC-80: In the case of rendering the Dropdown to a portal via the `usePortal` prop (true by default for now) it renders the dropdown to the `div #OverlayContainer` rendered by `stripes-core`. The `#OverlayContainer` can overlap parts of the UI, so its CSS has `pointer-events: none;` as you can see [here](https://github.com/folio-org/stripes-core/blob/master/src/components/OverlayContainer/OverlayContainer.css#L3). so it's up to potential DOM children to counteract this by setting their own `pointer-events: all` or else your clicks will go right through them.

util/getFocusableElements.js
```diff
@@ -0,0 +1,102 @@
...
+
+export function getLastFocusable(container) {
+  const focusable = getVisibleFocusableElements(container);
+  return focusable.length > 0 ? focusable[focusable.length - 1] : null;
+}
+
+export function getFirstFocusable(container) {
+  const focusable = getVisibleFocusableElements(container);
+  return focusable.length > 0 ? focusable[0] : null;
+}
```
VictorSoroka1:
```
export const getFirstFocusable = (container) => first(getVisibleFocusableElements(container));
export const getLastFocusable = (container) => last(getVisibleFocusableElements(container));
```

lib/PaneHeader/PaneHeader.js
```diff
@@ -164,26 +137,38 @@ class PaneHeader extends Component {
...
-    const { paneTitle, paneSub } = this.props;
-    const { actionMenuOpen } = this.state;
-    const paneActionMenu = getActionMenu();
+  renderActionDDTrigger = ({ getTriggerProps, open }) => (
+    <button
+      className={css.paneHeaderCenterButton}
+      type="button"
+      {...getTriggerProps()}
+    >
+      { this.getContent(open) }
```
VictorSoroka1:
```suggestion
      {this.getContent(open)}
```

util/getFocusableElements.js
```diff
@@ -0,0 +1,102 @@
...
+        return element.offsetWidth > 0 ||
+        element.offsetHeight > 0 ||
+        element === document.activeElement;
+      });
+    return focusableElements;
+  }
+  return [];
+}
+
+function getItem(collection, current, loop, next) {
```
VictorSoroka1: Could you please come up with a better name for the method? As I see some decent logic comes in the method and `getItem` does not help to understand enough.

util/getFocusableElements.js
```diff
@@ -0,0 +1,102 @@
...
+      if (element === currentElement) {
+        return element;
+      }
+      return !contains(currentElement, element);
+    });
+    focusable = outside;
+  }
+
+  if (focusable.length > 0) {
+    return getItem(focusable, currentElement, loop, false);
```
VictorSoroka1: As I see `getNextFocusable` and `getPreviousFocusable` only differ by the last parameter to the `getItem` function. Could you please cut the duplication logic by either combining them or something else.

lib/Dropdown/Popdown.js
```diff
@@ -0,0 +1,299 @@
...
+          'tabIndex': '-1',
+          'onBlur': menuBlur,
+          'onFocus': menuFocus,
+          'data-test-dropdown-menu-overlay': true
+        }}
+        portal={usePortal ? portalEl : null}
+      >
+        <React.Fragment>
+          {usePortal && focusProxy(focusTrigger, { 'data-reverse-proxy': true })}
+          { menu }
```
VictorSoroka1:
```suggestion
          {menu}
```

VictorSoroka1:
@JohnC-80, as I understand you have improved the accessibility side of the dropdowns, right? I am saying by testing the tab button behavior both for the dropdowns like in checkouted loans and in the navbar dropdown which opens the list of the apps. It looks great!

JohnC-80:
Yes! Tab keys should work for the items/menus. Arrow keys will navigate among menu items, looping to the first/last items if the current item is the first/last.

util/getFocusableElements.js
JohnC-80: I'll see how much I can combine the contents. I don't know about combining the two functions to one since they are the external API of this... just preferring the clarity of `getNextFocusable(...params)` and `getPreviousFocusable(...params)` to something like `getFocusable(...params, plusSomethingForPrevNext)`.

util/getFocusableElements.js
JohnC-80: your lodash-fu is relentless! 💪 💪

PR #1090, opened 2019-11-06
mountWithContext accepts translations

zburke:
# Summary
Provide `mountWithContext` with the ability to include additional
translations to those found locally.

# Approach
`mountWithContext` is exported as a test helper for use by other
repositories. As such, it should be able to include the translation
strings used by those repositories when running tests in order to avoid
errors like this during testing:
```
ERROR: '[React Intl] Missing message: "ui-circulation.settings.common.delete" for locale: "en-US"'
ERROR: '[React Intl] Cannot format message: "ui-circulation.settings.common.delete", using message id as fallback.'
```
By accepting an additional optional prop, `translations`, containing an
array of `{ translations, prefix }` objects, it can now load arbitrary
translations in addition to those it finds locally.

### DIFF
--- a/tests/Harness.js
+++ b/tests/Harness.js
@@ -16,19 +16,26 @@ const reducer = combineReducers(reducers);
 const store = createStore(reducer);
 
 // mimics the StripesTranslationPlugin in @folio/stripes-core
-function prefixKeys(obj) {
+function prefixKeys(obj, prefix) {
   const res = {};
   for (const key of Object.keys(obj)) {
-    res[`stripes-components.${key}`] = obj[key];
+    res[`${prefix}.${key}`] = obj[key];
   }
   return res;
 }
 
 class Harness extends React.Component {
   render() {
+    const t = prefixKeys(translations, 'stripes-components');
+    if (this.props.translations) {
+      this.props.translations.forEach(tx => {
+        Object.assign(t, prefixKeys(tx.translations, tx.prefix));
+      });
+    }
+
     return (
       <Provider store={store}>
-        <IntlProvider locale="en" key="en" timeZone="UTC" messages={prefixKeys(translations)}>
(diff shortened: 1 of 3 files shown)


tests/Harness.js
```diff
@@ -16,19 +16,26 @@ const reducer = combineReducers(reducers);
...
   for (const key of Object.keys(obj)) {
-    res[`stripes-components.${key}`] = obj[key];
+    res[`${prefix}.${key}`] = obj[key];
   }
   return res;
 }
 
 class Harness extends React.Component {
   render() {
+    const t = prefixKeys(translations, 'stripes-components');
```
VictorSoroka1: Please give a more descriptive variable name here.

tests/Harness.js
```diff
@@ -16,19 +16,26 @@ const reducer = combineReducers(reducers);
...
+    if (this.props.translations) {
```
VictorSoroka1: Does it make sense to set a default prop `translations` to an empty array and thus remove this check?

VictorSoroka1:
@zburke, could you please give an example of usage inside tests in other modules. I am more interested in what would be the value for translations. Is it imported translations files from other modules or some hardcoded strings.

zburke:
All this change does is remove some console noise. Previously, using `mountWithContext` outside `stripes-components` would cause warnings wherever it found `<FormattedMessage...>` et al because the required translations were not loaded into the test's context. 

Now it can provide translations along with the component to mount: 
```
import translations from '../../../../../translations/stripes-smart-components/en';
  
  mountWithContext(
    <DateRangeFilter ... />, 
    [{ translations, prefix: 'stripes-smart-components' }], 
  );  
```

zburke:
Thanks, @VictorSoroka1, for the thoughtful suggestions about how this could be further improved.

VictorSoroka1:
@zburke, thank you for the example. Now it is clear to me.

PR #1104, opened 2019-11-20
Icon: Support Tooltips

doytch:
@EthanFreestone found an issue where you couldn't wrap an `<Icon>` in a `<Tooltip>`. 

At first, it crashes due to the lack of a forwarded ref. Once I converted Icon to use `forwardRef`, we still needed some way of setting `aria-labelledby`. I just added support for this via another prop.

**Questions**
- Is this worth it? There is a workaround which is to wrap the `Icon` in a `<span>` and set the ref/aria-labelledby on the span instead.
- Does anyone know of a way we could be breaking a parent's expectations for what `ref` should here? I couldn't find any in the code.
- Should we _not_ add a specific `ariaLabelledBy` prop and instead do some `pickBy(rest, (_, key) => key.startsWith('aria-'))` grouping of props and apply them wholesale instead?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Export the `<Popper>` component at top level.
 * Added `<MessageBanner>` component (STCOM-592)
 * Optionally include translations in `mountWithContext` test helper.
+* Support `ref` in `<Tooltip>` via `forwardRef`.
 
 ## [5.8.0](https://github.com/folio-org/stripes-components/tree/v5.8.0) (2019-09-25)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v5.7.1...v5.8.0)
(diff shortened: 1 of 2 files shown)


rasmuswoelk:
I think this is a nice change. In general, I think that all components should support refs.

Regarding the use-case: The tooltips should be accessible by keyboard focus, so you shouldn't put Tooltips directly on non-focusable elements like the Icon. However, I can definitely see that this is something you would want to do in some cases. 

@JohnC-80 Do we have a recommended approach for these use-cases?

rasmuswoelk:
The Icon component will now show up as `forwardRef` in react dev tools.

See here how to make sure it says `Icon`: https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools

doytch:
> Regarding the use-case: The tooltips should be accessible by keyboard focus, so you shouldn't put Tooltips directly on non-focusable elements like the Icon. However, I can definitely see that this is something you would want to do in some cases.

Yeah I hear that. The particular scenario we want it for here is [the mockup in this issue](https://issues.folio.org/browse/ERM-463).

rasmuswoelk:
@doytch Yeah.. maybe add a `focusable`-prop so that it can be toggled on/off when needed.

rasmuswoelk:
@doytch I talked to @JohnC-80 and he suggested that we make the `tabIndex`-prop on the Icon changeable and then just default it to `-1`.

Then whenever you want to add a tooltip to an icon, you would just change the tabIndex. And we should probably document this.

doytch:
Updated with:
- `displayName` set
- All aria attributes passed down since we should _at least_ support not just labelledby but also describedby as well for this functionality.
- `tabIndex` prop

PR #1107, opened 2019-11-26
STCOM-623 validate args before accessing them

zburke:
Validate the shape of `groups.cql` before attempting to access it in
order to avoid an NPE when it is not defined.

Fixes [STCOM-623](https://issues.folio.org/browse/STCOM-623)


zburke:
Thinking about this more, maybe `filters2cql` should warn about the mismatched config/filters values it received in an `else` clause? I'm thinking it should, rather than just tiptoeing around the error without saying anything.

PR #1110, opened 2019-12-02
STCOM-628: Fix aria-labelledby parameter in Select component

Yurii-Danylenko:
https://issues.folio.org/browse/STCOM-628

## Purpose
Currently, `aria-labelledby` always points to label HTML element. So when there is no label, it points to nonexistent node and it causes an issue for screen readers.

## Approach
Define `aria-labelledby` parameter only when label is provided.

### DIFF
--- a/lib/Select/Select.js
+++ b/lib/Select/Select.js
@@ -68,6 +68,19 @@ class Select extends Component {
     );
   }
 
+  getAriaLabelledBy() {
+    let ariaLabelledBy;
+    const label = this.props.label && `${this.id}-label`;
+    const readOnly = this.props.readOnly && `${this.id}-read-only-message`;
+    const ariaLabelledByString = [label, readOnly].filter(item => !!item).join(' ');
+
+    if (ariaLabelledByString.length) {
+      ariaLabelledBy = ariaLabelledByString;
+    }
+
+    return ariaLabelledBy;
+  }
+
   render() {
     /* eslint-disable no-unused-vars */
     const {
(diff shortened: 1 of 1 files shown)


Yurii-Danylenko:
> Approved. I believe this component should also support a direct `ariaLabelledBy` prop (it doesn't look like it currently does) but that can be added in a future PR. Thanks @Yurii-Danylenko !

Thanks for approving @JohnC-80! I read a little bit more about `aria-labelledby` attribute here https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships#aria-labelledby. One of the use cases is composing a label from multiple elements. From this perspective, I don't think that composing the `aria-labelledby` inside form control component is a good idea. What about giving a user the freedom to pass whatever `aria-labelledby` he wants? From outside of the component it's possible to refer to a label id and I think it's enough to build well accessible labels.

JohnC-80:
@Yurii-Danylenko I do agree that a developer should be able to pass their own `aria-labelledby` prop I think this could be added to the implementation that you have here - in our case, we use `aria-labelledby` on `selects` for when the `readonly` prop is in use. Here's the interesting part/reason why it is this way: `select` elements don't support the `readonly` attribute, and thus, don't have the attribute read out through screen readers like `input[type='text']` or other controls... so the only way to have it read out is by adding it on to the end of the `aria-labelledby` IDREFs... and for that, the string would have to be in the format of `"<labelIDREF> <readonlyIDREF>"` for both to effectively be announced (Otherwise aria-labelledby is annouced **instead of** the `<label> 's text.)

lib/Select/Select.js
```diff
@@ -68,6 +68,19 @@ class Select extends Component {
     );
   }
 
+  getAriaLabelledBy() {
+    let ariaLabelledBy;
+    const label = this.props.label && `${this.id}-label`;
+    const readOnly = this.props.readOnly && `${this.id}-read-only-message`;
+    const ariaLabelledByString = [label, readOnly].filter(item => !!item).join(' ');
```
JohnC-80: Seems like we could just insert an external ariaLabelledBy prop here like so: 
```suggestion
    const ariaLabelledByString = [this.props.ariaLabelledBy, label, readOnly].filter(item => !!item).join(' ');
```

Yurii-Danylenko: Yeah, we could, but now we are hardcoding an order of references in the `aria-labelledby` string. I will merge it as is to include in stripes-components release and create another story to think over how to manage `aria-labelledby`.

PR #1112, opened 2019-12-03
STCOM-553: Added "Actions"-button in PaneHeader

rasmuswoelk:
**Ticket:**  https://issues.folio.org/browse/STCOM-553

## Changes
- Removed action menu toggle from the pane title
- Added an "Actions"-button in the right side of the pane menu which will now render the action menu

## Screenshot
image


zburke:
@JohnC-80, @rasmuswoelk, there is a long [thread in Slack#development about this topic](https://folio-project.slack.com/archives/C210RP0T1/p1574722646290800). It would be great to get this PR squashed and merged before it diverges even further from `#master`.

lib/PaneHeader/PaneHeader.js
```diff
@@ -10,107 +10,93 @@ import PaneMenu from '../PaneMenu';
...
+  header,
+  firstMenu,
+  dismissible,
+  paneTitleAutoFocus,
+  paneTitleRef,
+  lastMenu,
+  id
+}) => {
+  const _id = `paneHeader${id}`;
+  const hasActionMenu = typeof actionMenu === 'function' && actionMenu();
```
zburke: [Many](https://github.com/folio-org/ui-requests/blob/4cd33a453eff8aaf348a06a5c2d1fdc21db60c55/src/routes/RequestsRoute.js#L659) [existing](https://github.com/folio-org/ui-users/blob/01caad7a7493e42b9f5b59c23e3e0246a4ee1f35/src/views/UserSearch/UserSearch.js#L132) [implementations](https://github.com/folio-org/ui-inventory/blob/7e06ba00ddf69bcbff1c15662fde91bcda5bfede/src/components/InstancesList/InstancesList.js#L140)  of `actionMenu` rely on it being called with `onToggle`. Calling it here with no arguments causes an NPE.

rasmuswoelk: @zburke I thought I pushed a commit for that already but apparently not. I will fix it ASAP.

rasmuswoelk:
@zburke I have just returned from traveling and I will get right on this one!

lib/PaneHeader/PaneHeader.js
rasmuswoelk: Yup.. had a few commits that I hadn't pushed yet. It should fix the issue.

PR #1115, opened 2019-12-05
Support ref in Modal via forwardRef

aditya-matukumalli:
In relation to https://github.com/folio-org/ui-plugin-find-user/pull/112, we need to have a reference to the Modal so that the condition [here](https://github.com/folio-org/ui-plugin-find-user/pull/112/files#diff-96e3556dd15eba52e6c744f5d35fd2eeR56) would resolve to true if the **X** dismissible button when in focus (when clicked), is still inside the Modal.

Previously, we were just targeting the [modal content](https://github.com/folio-org/ui-plugin-find-user/pull/112/files#diff-96e3556dd15eba52e6c744f5d35fd2eeL56) and due to this the **X** iconButton in the modals header wasn't contained in the element targeted by the modalContent `ref`, and hence the focus on the modal trigger [here](https://github.com/folio-org/ui-plugin-find-user/pull/112/files#diff-96e3556dd15eba52e6c744f5d35fd2eeR57) when the modal was dismissed wasnt being called. This PR fixes this.

### DIFF
--- a/lib/Modal/Modal.js
+++ b/lib/Modal/Modal.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 import { FormattedMessage } from 'react-intl';
 import { Modal as OverlayModal } from 'react-overlays';
 import classNames from 'classnames';
@@ -75,7 +75,7 @@ const defaultProps = {
   wrappingElement: 'div',
 };
 
-const Modal = (props) => {
+const Modal = forwardRef((props, ref) => {
   const {
     children,
     closeOnBackgroundClick,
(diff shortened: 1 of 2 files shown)


lib/Modal/Modal.js
```diff
@@ -117,7 +117,7 @@ const Modal = (props) => {
     contentClass,
   );
 
-  const renderBackdrop = ({ onClick, ref, style }) => { /* eslint-disable-line react/prop-types */
+  const renderBackdrop = ({ onClick, ref, style }) => { /* eslint-disable-line react/prop-types, no-shadow */
```
doytch: We shouldn't just disable eslint here. If we can't write this without shadowing, it's better to rename the `ref`, eg,

```suggestion
  const renderBackdrop = ({ onClick, ref: backdropRef, style }) => { /* eslint-disable-line react/prop-types, no-shadow */
```

I'm not saying use that name in particular, because honestly, I can't really follow who is providing that inner ref.

doytch:
Oops, I **really** wasn't suggesting to use that code suggestion because you need to actually use that renamed ref as well, no?

aditya-matukumalli:
> Oops, I **really** wasn't suggesting to use that code suggestion because you need to actually use that renamed ref as well, no?

oops

PR #1119, opened 2019-12-09
UIU-1374: Add proper description about attribute

annamelnyk:
### Description
Fix `autocomplete` attribute description in README.md


lib/TextField/Readme.md
```diff
@@ -27,7 +27,7 @@ Name | type | description | default | required
 --- | --- | --- | --- | ---
 `ariaLabel` | string | Applies an `aria-label` attribute - prefer visible `label` prop. Use only if the labeling case warrants. | |
 `ariaLabelledBy` | string | Applies an `aria-labelledBy` attribute to the `<input>` - prefer visible `label` prop. Use only as the labeling case may warrant. | |
-`autoComplete` | bool | If this prop is `true`, control will automatically focus on mount | |
+`autoComplete` | string | Applies `on`|`off` value.
```
VictorSoroka1:
```suggestion
`autoComplete` | string | Applies `on` | `off` value. 
```

doytch: Should actually be something like this, no? To prevent Markdown from splitting the description into multiple columns.
```suggestion
`autoComplete` | string | Applies `"on"` or `"off"` value. 
```

VictorSoroka1: Fair enough.

doytch:
OK now that I actually look at the implementation of `TextField`, this description doesn't really explain what it does. The prop's value is just set directly to the `autoComplete` attribute of the `input`, so you could set it to `"chicken"` if you really wanted to, not just on/off as the description states. 

So I'd say the description should instead match the description pattern used for `placeholder` or `type`.

annamelnyk:
@doytch, the `autocomplete` is default input's attribute, which can accept only 'on'/'off' values by specification.

doytch:
> @doytch, the `autocomplete` is default input's attribute, which can accept only 'on'/'off' values by specification.

Right, but this component doesn't actually coalesce the values to "on" and "off". It blindly passes on the value to the `input` without any type-checking, putting the onus on the developer using this component to pass a sane value. Just like how `type` should only be one of a specific set of strings.

VictorSoroka1:
As I see from the Zak's [comment](https://github.com/folio-org/stripes-core/pull/768#discussion_r355588836) and your @doytch reducing the set of values to just `on` or `off`  might not always be the desired result, so I also agree with more abstract wording.

annamelnyk:
> [`autocomplete` can actually accept various values.](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) Furthermore, this component doesn't actually coalesce the values to "on" and "off". It blindly passes on the value to the `input` without any type-checking, putting the onus on the developer using this component to pass a sane value. Just like how `type` should only be one of a specific set of strings.
Thanks for explanation link, I agree, Already changed description.

PR #1123, opened 2019-12-13
Add renderHeader prop for <Pane>

rasmuswoelk:
In this PR I have added a `renderHeader`-prop for the `<Pane>` that can be used to render a `<PaneHeader>`.

Currently, we pass all of the header related props for the down from the `<Pane>` to the `<PaneHeader>`. These props aren't needed in the `<Pane>`-component itself.

With the `renderHeader`-prop we can simply render the `<PaneHeader>` directly using the render-prop pattern without doing any prop-drilling. With this change, it's also possible to pass `null` to the `renderHeader`-prop to remove the pane header entirely or even rendering a custom header if needed.

This change is backward compatible and should not break anything. I haven't added any deprecation warnings yet because we might still want to experiment with this new API. However, I have updated the documentation to reflect this change.


rasmuswoelk:
Thanks for the approve, @JohnC-80. As discussed on Slack, I will make some documentation updates soon and merge this in after.

PR #1126, opened 2019-12-17
STCOM-610: Increase FilterGroups BigTest coverage to 80% 

annamelnyk:
### Description
Covering FilterGroups stripes-component with tests 

---

### Story
https://issues.folio.org/browse/STCOM-610

--- 

### Screenshot
image


lib/FilterGroups/tests/FilterGroups-test.js
```diff
@@ -0,0 +1,283 @@
...
+      expect(filterGroupInteractor.hasFilterCheckbox).to.be.true;
+    });
+
+    it('render checkboxes', () => {
+      expect(filterGroupInteractor.checkboxes().length).to.equal(3);
+    });
+
+    it('render checkbox label', () => {
+      expect(filterGroupInteractor.checkboxes(0).hasLabelElement).to.be.true;
+      expect(filterGroupInteractor.checkboxes(0).label).to.equal('Books');
```
maximdidenkoepam: We need to use translations value instead of plain text. Please fix it in all cases.

annamelnyk: We do not need to use translations in `FilterGroups` component, because there are used mocked values.

PR #1127, opened 2019-12-17
UICHKIN-146: move countries data to stripes-components

MaksymDryha:
Part of [UICHKIN-146](https://issues.folio.org/browse/UICHKIN-146)
 ## Purpose
To have a single place across all modules where country names and codes are stored
## Approach
In the `util/countries.js` file, there is a list of country codes, exported globally. The names of the countries are listed in `en.json`. When one needs to render a country name, they can do it in the following way:
```js
import { countryCodes, countries, countriesByCode } from '@folio/stripes-components';

intl.formatMessage({id: `stripes-components.countries.US`}),  // any of the country codes listed in the above imported variables
```

## TODO
Refactor address `AddressFieldGroup` in `stripes-smart-components` to use country names from `stripes-components`


zburke:
It absolutely puts an expectation on the stripes-components translators to translate the names of every country, but those values have to come from somewhere. At present, we only present English names for countries and that's clearly wrong.

PR #1133, opened 2019-12-20
UIORGS-131 Update table layouts in edit forms to remove redundant headings

NikitaSedyx:
purpose: https://issues.folio.org/browse/UIORGS-131

approach: new property to display fields heading only one + styles (SS is attached)

doc has been updated
test has been added

ss


JohnC-80:
There needs to be some sort of accessibility note with this that the headings will need id's and the fields will need corresponding `aria-labelledby` attributes applied if this technique is used. Also, if this is to be a prevalent pattern to follow, we should include an example in documentation that includes the accessibility implementation.

NikitaSedyx:
@JohnC-80 IMO accessibility is part of fields and labels definition, I did it with `aria-label`, it's just a heading - not a part of this component

agreed, I'll add an example for this case

zburke:
Do you have a before and after screenshot? Sorry; it isn't clear to me from the ticket what the problem is (there is no "expected results, actual results" comparison), nor from this PR how that problem is addressed. 

Based on @JohnC-80's comment, I am concerned that this makes managing accessibility correctly a manual process that is the responsibility of the developer. Certainly it is possible for a developer to do correctly, but one of the major benefits of using components is to be certain that it is already handled correctly and automatically.

NikitaSedyx:
@zburke we have it now - repeatable labels, and our PO and UX team requested to avoid them
ss

as for accessibility, in this component it's controlled manually by developer because everything is passed in props
if you know how to make it better please advice

NikitaSedyx:
@JohnC-80 I've got you incorrectly, yep, I'll update readme with note, thanks!

lib/RepeatableField/RepeatableField.js
```diff
@@ -82,7 +95,11 @@ class RepeatableField extends Component {
                 </div>
                 {
                   typeof onRemove === 'function' && (
-                    <div className={css.repeatableFieldRemoveItem}>
+                    <div
+                      className={
+                        `${css.repeatableFieldRemoveItem} ${headLabels && css.repeatableFieldRemoveUnlabeledItem}`
```
ViktorSoroka: What would be if `headLabels` in `headLabels && css.repeatableFieldRemoveUnlabeledItem` is falsy? I wonder whether something is set in that case to the class value. If so, I would recommend proceeding with the ternary option.

 `headLabels ? css.repeatableFieldRemoveUnlabeledItem : ''`

NikitaSedyx: actually nothing critical, but yep, it's better

NikitaSedyx:
@JohnC-80 please take a look, thanks

NikitaSedyx:
@zburke and it's react final form

PR #1140, opened 2020-01-03
Upgrade Babel from v6 to v7.8

doytch:


doytch:
This PR's tests cannot pass until https://github.com/folio-org/stripes-core/pull/780 is merged. When building the tests, the dependency on stripes-core brings in the old stripes-core w/ `stage-2` presets that don't work under Babel 7.

PR #1146, opened 2020-01-10
STCOM-633 Horizontal scrollbar when columnWidths add up to 100%...

JohnC-80:
# Problem
See title... the width calculation was based on a full width value rather than the value minus the margins, hence the horizontal scroll.

## Approach
 Adjusted the width being calculated in the column width calculation.

### Before:
image

### After:
image

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -103,10 +103,14 @@ export function initColumnsFromData(item, headerMeta, rowMeta) {
 
 function getPixelColumnWidths(columnWidths, width) {
   const columnsFromProps = {};
+  // account for margin of rows (5px in CSS... 5 * 2 = 10).
+  // this will prevent percentage width that sum to 100% values from exceeding
+  // actual width and causing horizontal scroll. - STCOM-633
+  const calculatedWidth = width - 10;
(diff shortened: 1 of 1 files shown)


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -103,10 +103,14 @@ export function initColumnsFromData(item, headerMeta, rowMeta) {
 
 function getPixelColumnWidths(columnWidths, width) {
   const columnsFromProps = {};
+  // account for margin of rows (5px in CSS... 5 * 2 = 10).
+  // this will prevent percentage width that sum to 100% values from exceeding
+  // actual width and causing horizontal scroll. - STCOM-633
+  const calculatedWidth = width - 10;
```
zburke: I see you use `rowMargin` below for this calculation. Is that not available here? Magic numbers always scare me, even when there are helpful comments explaining where they come from. The CSS value could change from 5 to 6 and then `calculatedWidth` would be wrong again.

JohnC-80: yeah... I'm going to try and add a regression test for this case with 4 25% columns, like my storybook screenshots in the descriptions just for the sake of this potentially changing. As for the variable, I'm thinking I may just need to make a `const` at the top of the file and use it in the places where it's needed...

PR #1148, opened 2020-01-13
STSMACOM-247-Convert `<Multiselection>` to use `<Popper>`

JohnC-80:
# Issues
Aside from the general inefficiencies reported from overlay components that depended on `tether.js` for positioning, cases where the overlay is actually cut-off by the borders of the window are another problem. See [STSMACOM-247](https://issues.folio.org/browse/STSMACOM-247)

## Approach
Convert the component to use (well, prefer) `<Popper>`. The `tether` prop is still available - and, if used, the component will revert to the tether-based implementation. This implementation makes use of `<Popper>`'s `modifiers` prop to "drop-up" in cases where the overlay would be cut off by the borders of the screen.
To smoothly handle the transition, I created a top-level boolean: `useLegacy` that I could pass down through the tree. Then I created a new sub-component to work as the container for the options list (`<OptionsListWrapper>`).

### Screenshot
multiselection

#### Before: Dropdown list cut off at bottom of window.
image

#### After: Dropdown list 'drops' upward.
image


zburke:
Reviewers are happy; tests pass. Are we good to merge, @JohnC-80?

PR #1159, opened 2020-01-30
STCOM-574: Add test coverage

annamelnyk:
# Description
Add test coverage for `<ExportCSV>` component
# Story
https://issues.folio.org/browse/STCOM-574
# Screenshot
image


lib/ExportCsv/exportToCsv.js
```diff
@@ -17,10 +17,13 @@ function triggerDownload(csv, fileTitle) {
       link.setAttribute('download', exportedFilename);
       link.style.visibility = 'hidden';
       document.body.appendChild(link);
-      window.setTimeout(() => {
-        link.click();
-        document.body.removeChild(link);
-      }, 50); // Need to debounce this click event from others (Pane actionMenuItems dropdown)
+      // this check is for testing browser to prevent file downloading
+      if (window.location.port !== '9876') {
```
VictorSoroka1: I would rather have another way the check whether there is a testing mode at the moment. 

`process.env.NODE_ENV === 'test'`

PR #1164, opened 2020-02-07
STCOM-637 Curb code that's prone to cause memory leaks in pane/paneset.

JohnC-80:
Storing class instances (this) or refs in state can accumulate since these objects (and their heirarchies of children) stick around in memory after they're re-rendered/unmounted.

This accumulation happens even in production builds of react, and can cause slow-down over long periods of use. This is a symptom of react after 16.8-ish - the path/problem is roughly documented in 
https://blog.discordapp.com/discord-react-memory-leak-565c89763e8

Panes/panesets, since they wrap great portions of the UI are prone to draw heavy memory since they contain so many children.

## Approach
Refs/class instances were mostly stored as convenience to access methods. Rather than store the entire class, an object with only the necessary properties was stored.
ex before:
```
state: { paneManager: this }
```
ex after: 
```
state: {
  paneManager: {
    handleClose: this.handleClose,
    registerPane: this.registerPane,
    removePane: this.removePane,
    getContainer: this.getContainer,
  }
```


JohnC-80:
Thanks guys.. will merge after tomorrow's demos

PR #1166, opened 2020-02-13
Added 'subValue'-prop for <KeyValue>

rasmuswoelk:
## Changes
- Added `subValue`-prop for `<KeyValue>`.
- Updated readme
- Updated examples
- Updated readme

image


JohnC-80:
Minor - does `sub` work best as a prop-name here? Maybe `subValue` or `subtext` as something more specific?

rasmuswoelk:
@JohnC-80 Personally, I think it's fine. But I can change it to subValue.

PR #1169, opened 2020-02-14
Add `bodyTag` prop for `<ConfirmationModal>`

mariia-aloshyna:
## Changes
- Added `bodyTag`-prop for `<ConfirmationModal>`
- Updated examples
- Updated readme
- Updated tests


lib/ConfirmationModal/ConfirmationModal.js
```diff
@@ -64,9 +67,9 @@ const ConfirmationModal = (props) => {
       size="small"
       footer={footer}
     >
-      <p className={css.message}>
```
annamelnyk: @mariia-aloshyna it looks great what you have done. I am just wondering if it could be easier replace `p` with regular `div` to make this console error message dissapear? What do you think, would it be worked?

mariia-aloshyna: You are right, it could be easier, but if some string will be passsed to the message prop (as it is done in the test) then it will be better to render this message inside the <p> tag.

annamelnyk: Better to render message inside `p` tag rather than `div`? Better in what sense, like semantic or what is the difference?

mariia-aloshyna: Yes, I mean semantic difference, `p` tag is more suitable for describing some text (paragraph), in turn, `div` tag behaves like a wrapper.

lib/ConfirmationModal/tests/interactor.js
```diff
@@ -2,10 +2,15 @@
...
 import ButtonInteractor from '../../Button/tests/interactor';
 
+const tagName = selector => computed(function () {
+  return this.$(selector).tagName.toLowerCase();
+});
+
 export default interactor(class ModalFooterInteractor {
   confirmButton = scoped('[data-test-confirmation-modal-confirm-button]', ButtonInteractor);
   cancelButton = scoped('[data-test-confirmation-modal-cancel-button]', ButtonInteractor);
+  bodyTagName = tagName('[class^="message"]');
```
rasmuswoelk: Can you add a data-test attribute for the modal body and use it here? It could be useful for other tests – both in stripes-components and in individual modules.

lib/ConfirmationModal/ConfirmationModal.js
```diff
@@ -9,6 +9,7 @@ import ModalFooter from '../ModalFooter';
 import css from './ConfirmationModal.css';
 
 const propTypes = {
+  bodyTag: PropTypes.string,
```
rasmuswoelk: I wonder if this prop should be called `bodyElement` instead and maybe support passing React components as well? That way you could pass a custom component directly to replace the body element entirely instead of having to pass it as a child to the Modal.

In that case, the prop types could be:

`bodyElement: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),`

JohnC-80:
I like this. Having it default to `<p>` still encourages brevity - but there may be instances where more is required to display in these brief modals. Having the more versatile `bodyElement` prop as @rasmuswoelk mentions would be a positive enhancement.

lib/ConfirmationModal/ConfirmationModal.js
mariia-aloshyna: Hi @rasmuswoelk Apologize for the late reply. If I got you right, you propose to replace part of code where `props.message` renders with `bodyElement`, but why don't devs pass React component to `props.message`? In this case, the React component will be just wrapped in a tag (bodyTag) as it is now.

PR #1173, opened 2020-02-17
STCOM-635 AccordionStatus Component

JohnC-80:
With this component holds toggle state for all accordions within an `<AccordionSet>`. It can be used as an alternative to maintaining accordion state and onToggle() handlers within an App.

Simply:
```
  <AccordionStatus>
     <ExpandAllButton />
     <AccordionSet>
        <Accordion label="AccordionStatus_one">
           <p>first content</p>
         </Accordion>
         <Accordion label="AccordionStatus_two">
            <p>second content</p>
          </Accordion>
      </AccordionSet>
    </AccordionStatus>
```

The `<AccordionStatus>` will set up entries corresponding with each `id` prop, generating unique identifiers as-needed.
The `<ExpandAllButton>` can be used without any props, as it will inherit the necessary `onToggle` and `accordionStatus` props from the `AccordionStatus` context.


lib/Accordion/AccordionStatusContext.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+export const AccordionStatusContext = React.createContext();
+
+export const withAccordionStatus = Component => {
+  const WrappedComponent = props => (
+    <AccordionStatusContext.Consumer>
+      {status => (
+        <Component
+          accordionStatus={status && status.status}
```
VictorSoroka1: With babel 7 I assume it is feasible now. Would it be possible to use that?
```suggestion
          accordionStatus={status.?status}
```

lib/Accordion/ExpandAllButton.js
```diff
@@ -3,46 +3,52 @@ import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
 import expandAll from './expandCollapseAll';
 import Button from '../Button';
+import { withAccordionStatus } from './AccordionStatusContext';
+
+// returns true if more keys in status are set to false than true;
+const majorityCollapsed = (status) => {
```
rasmuswoelk: This function could be memoized to prevent it from running on each re-render. Lodash has a utility function for this purpose I belive.

rasmuswoelk: This function could be memoized to prevent it from running on each re-render. Lodash has a utility function for this purpose I believe.

lib/Accordion/AccordionStatus.js
```diff
@@ -0,0 +1,57 @@
...
+        }}
+      >
+        {typeof children === 'function'
+          ? children({
+            status: this.state,
+            setStatus: this.setStatus,
+            onToggle: this.onToggle
+          })
+          : children}
+      </AccordionStatusContext.Provider>
```
rasmuswoelk: Could these values be moved into a constant since they are being passed in as arguments in two places?

lib/Accordion/ExpandAllButton.js
JohnC-80: Thanks! I'll look into some memoization for it. It doesn't strike me as a very expensive operation, but these little things can pile up...

lib/Accordion/AccordionStatus.js
JohnC-80: Good idea.

JohnC-80: Updated.

lib/Accordion/AccordionStatusContext.js
JohnC-80: Thanks Victor! Time to start using these nice, new things...  updated.

lib/Accordion/ExpandAllButton.js
rasmuswoelk: No, probably not. But yes.. they sure do pile up :)

PR #1186, opened 2020-03-02
Fix always-on deprecation message and inputRef warning in `<MultiSelection>`.

JohnC-80:
Despite the good intentions of the `deprecated` function from `prop-types-extra`, it doesn't account for the prop having a `defaultProps` entry - in that case, it always displays whether the prop is used or not.. moved the error to the constructor where `useLegacy` is determined.

Also, `<OptionsListWrapper>` was being given an `inputRef` prop that it didn't need to pass along to its rendered elements.

These should clear up a couple of console annoyances.


lib/MultiSelection/MultiSelection.js
```diff
@@ -124,6 +120,9 @@ class MultiSelection extends React.Component {
       <FormattedMessage id="stripes-components.multiSelection.defaultEmptyMessage" />;
 
     this.useLegacy = !isEqual(props.tether, MultiSelection.defaultProps.tether);
+    if (this.useLegcy) {
+      throw new Error('MultiSelection: tether prop will be removed in future releases. Use "modifiers" prop instead.');
+    }
```
zburke: `this.useLegacy`: leg-a-cy. 

Also, if we're `throw`ing in the constructor, won't that actively prevent you from using those props? That is, it's unusable in _this_ release, ain't it?

JohnC-80: *derp... vowels... are books would be so much shorter without them, narf! I'll fix... and warn....

PR #1197, opened 2020-03-10
MCL offset paging

JohnC-80:
## Problem 
MCL doesn't currently behave well when given an array with gaps in the indices (indices with null values.) Loaders for previously null items don't get their positions updated correctly when new items are loaded in. The `scrollToIndex` functionality doesn't really work since the DOM elements aren't quite big enough.

## Approach
* Expand the logic that checks for changes so that specific types of changes are reported - `dataChangedOrLess` returns an object with `changed` and `changeType` properties.
Within MCL this can be used to determine if new data is loaded in or sorting just happened - i.e. whether or not the list should scroll back to top when the array of the contentData is the same, but the data has changed.
* An additional state field has also been added so that loaders rendering after the extent of the contentData will have their positions updated when new data is received.
* If scrollToIndex is used, the ScrollContainer will have a minheight set to `scrollToIndex * minRowHeight + viewportHeight`


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -74,19 +74,32 @@ function comparison(obj, oth) {
...
-    const notEqual = !isEqualWith(previous[i] || {}, current[i], comparison);
-    if (notEqual) return true;
+  // find a suitable interval...
+  const interval = Math.min(4, Math.ceil(previous.length / 6));
+  for (let i = 0; i < previous.length - 1; i += interval) {
+    const notEqual = !isEqualWith(previous[i] || {}, current[i] || {}, comparison);
+    if (notEqual) {
+      return {
+        changed: true,
+        changeType: previous[i] === null ? 'new data' : 'difference'
```
ryandberger: Should throw these strings into top-level constants.

PR #1208, opened 2020-03-17
UIEH-832 Added ariaLabel prop to <NoValue>

BogdanDenis:
## Purpose
- Give `<NoValue>` component users ability to pass custom text to `aria-label` attribute.


mkuklis:
Nice work @BogdanDenis!

PR #1210, opened 2020-03-17
AccordionSet: registerAccordion should account initialStatus

JohnC-80:
Bug: initialStatus prop of AccordionSet doesn't work.

Reason: it's not accounted for when the accordion registers with the AccordionSet.

This PR fixes that.


JohnC-80:
Thanks @rasmuswoelk ! going to wait for a look from @Baroquem  since he reported the issue and is making use of the feature....

PR #1211, opened 2020-03-18
Add note about color contrast

annamelnyk:
# Description
Add note about WCAG color contrast checker issue, which is actually NOT an issue
# Issue
https://issues.folio.org/browse/UICHKOUT-603
# Screenshots
Screenshot_4 (1)


VictorSoroka1:
Once you are here, I noticed a duplication of the word `announce` [here](https://github.com/folio-org/stripes-components/blame/363308eb6ef6b57a89b2a8ba53753898646fb471/docs/AccessibilityDevPrimer.md#L27). Could you please delete one of them?

docs/AccessibilityDevPrimer.md
```diff
@@ -75,6 +75,8 @@ Color is a great tool for meaning - but it should never be the *only* indicator
 
 Color contrast is an important aspect to ensure that the webpage is legible, even for users with color-blindness or other sight disabilities. WCAG 2.0 AA standard requires that elements on a page have a sufficient color contrast of 4.5:1 for normal text and 3:1 for large or bold text.([1.4.3](https://www.w3.org/TR/WCAG21/#contrast-minimum)) Check out the links below for a color contrast checking tool.
 
+**Not the issue**: When the contrast checker tool checks the icons which have a default/placeholder background color and a text color but there aren't actually any text – it's rendering an icon, so there's no issue.
```
JohnC-80:
```suggestion
**Known false positive**:  When the contrast checker tool checks the icons which have a default/placeholder background color and a text color with no visible text.
```
This could probably be the start of a list - we're bound to collect these false positives for reference.

PR #1220, opened 2020-03-25
STCOM-649 Refactor away from componentWillReceiveProps

ryandberger:
- This is a first pass to move as much code out of the deprecated React lifecycle `componentWillReceiveProps` as possible.
- When moving code into `componentDidUpdate` the general strategy is for `nextProps` to go to `this.props` and `this.props` to go to `prevProps` since we have moved to the other side of the `render` lifecycle.
- Remaining code was attempted to move, but all attempts thus far have resulted in failing tests. More investigation is needed into how we handle these other cases, or if there are some actions that aren't needed anymore, which will also require modifying some of the test assertions.
- Also note, that state is being set in some cases for `componentDidUpdate`. While not ideal, I'm open to suggestions here if there is a better path forward.

### DIFF
--- a/lib/SRStatus/SRStatus.js
+++ b/lib/SRStatus/SRStatus.js
@@ -6,6 +6,7 @@ class SRStatus extends React.Component {
   static propTypes = {
     ariaLive: PropTypes.oneOf(['polite', 'assertive', 'off']),
     message: PropTypes.string,
+    rest: PropTypes.object,
   }
 
   static defaultProps = {
@@ -24,8 +25,7 @@ class SRStatus extends React.Component {
     };
   }
 
-  shouldComponentUpdate(nextProps)
-  {
+  shouldComponentUpdate(nextProps) {
     if (nextProps.message !== undefined &&
       nextProps.message !== '') {
       return true;
@@ -35,7 +35,7 @@ class SRStatus extends React.Component {
     }
     return false;
   }
-  
+
   componentDidUpdate(prevProps) {
     if (prevProps.message === this.props.message) {
       if (this.willUpdateRepeats) {
--- a/lib/SRStatus/SRStatus.js
+++ b/lib/SRStatus/SRStatus.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import isEqual from 'lodash/isEqual';
 
 class SRStatus extends React.Component {
   static propTypes = {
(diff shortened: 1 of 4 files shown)


lib/SRStatus/SRStatus.js
```diff
@@ -23,16 +25,24 @@ class SRStatus extends React.Component {
     };
   }
 
-  componentWillReceiveProps(nextProps) { // eslint-disable-line react/no-deprecated
+  shouldComponentUpdate(nextProps) {
```
ryandberger: This extra lifecycle became necessary since `message` not changing needs to be accounted for as long as it isn't `undefined` and `this.willUpdateRepeats` is `true`.

JohnC-80: The most typical use of this component is for people to set a ref and use ref.sendMessage directly... that being the way, I think we need to account for state changes in `shouldComponentUpdate` as well. I hate that ShouldComponentUpdate is such a sticky slope, but this is definitely the reason why PureComponent exists (cursory glance, but maybe that's what we should use?)

lib/Selection/SingleSelect.js
```diff
@@ -148,19 +150,34 @@ class SingleSelect extends React.Component {
...
+        this.setState((curState) => {
+          const newState = Object.assign({}, curState);
+          newState.cursoredIndex = curState.selectedIndex;
+          newState.cursoredValue = curState.selectedValue;
+          newState.cursoredLabel = curState.selectedLabel;
+          return newState;
+        });
+      }
+    }
+  }
```
JohnC-80: I wonder if we could just always perform this sync in a `gDSFP` - I'm not sure if the displayed status is really important for this - mainly just that the selectedIndex is set...

lib/Timepicker/TimeDropdown.js
```diff
@@ -79,24 +81,24 @@ class TimeDropdown extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) { // eslint-disable-line react/no-deprecated
+    if (!nextProps.visible && nextProps.visible !== this.props.visible) {
+      this._focusTimeout = setTimeout(() => { this.props.mainControl.current.focus(); }, 20);
+    }
+  }
```
JohnC-80: I feel like since handling focus is a DOM-related activity, ComponentDidUpdate is probably the best place for this logic.

lib/Timepicker/Timepicker.js
```diff
@@ -135,29 +137,48 @@ class Timepicker extends React.Component {
...
+        }
+
+        return {
+          presentedValue: null,
+          timeFormat: this._timeFormat,
+          timeString: newTimeString,
+        };
+      });
+    }
+  }
```
JohnC-80: this has potential for `gDSFP` as well.... context probably should be coming through in props via an HOC...

lib/SRStatus/SRStatus.js
ryandberger: I don't think I can go the PureComponent route since it seems that an existing use case is for someone to be able to pass the same `message` value multiple times. PureComponent would just ignore those values. I will include state changes though!

lib/Timepicker/TimeDropdown.js
ryandberger: I'll give it a go.

JohnC-80:
Closing this guy since these updates are/have been handled elsewhere...

PR #1223, opened 2020-03-26
UIU-1539: Add possibility to click to set Associated Service Point for Fee/Fine Owner

annamelnyk:
# Description
**Settings --> Users --> Fee/Fine Owners**: Unable to use mouse to set Associated Service Point for Fee/Fine Owner.
# Approach
Looks like `OverlayContainer` is not a parent for Settings --> Users --> Fee/Fine Owners, so I changed to a parent component `ModuleContainer`.
# Issue
https://issues.folio.org/browse/UIU-1539
# Screenshot
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 6.1.1 (IN-PROGRESS)
 
 * Fix issue with `initialStatus` prop on Accordions not working.
+* Fix bug with impossibility to use mouse to set Associated Service Point for Fee/Fine Owner. Refs UIU-1539.
 
 ## [6.1.0](https://github.com/folio-org/stripes-components/tree/v6.1.0) (2020-03-16)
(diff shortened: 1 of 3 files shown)


JohnC-80:
@annamelnyk OverlayContainer functions as a z-index safe place for these kinds of controls. While it does need a landmark role associated with it for accessibility purposes, it has a more functional purpose/convention across the component library, so I'm not sure this is the change we want... I believe the best solution would be to set an attribute of `.multiSelectMenu` to `pointer-events: all`. I believe this was a regression from Popper conversion and applies to cases where a portal is used (most now render the popper locally)

annamelnyk:
@JohnC-80, that is unexpected solution for me, `pointer-events: all` fixed current issue, I have killed bunch of time on investigation why click does not work while I found "solution" with `portal`, and the funniest is that there were no thoughts to try to fix it with styles. Thanks.

PR #1225, opened 2020-03-27
UIU-1537: Change style

annamelnyk:
# Description
 Settings>Users>Fee/Fine Owner: Associated Service Point that has a long name overpaps `Save`, `Cancel` buttons 
# Issue
https://issues.folio.org/browse/UIU-1537
# Screenshots
**Before**

image

**After**

image

### DIFF
--- a/lib/MultiSelection/MultiSelect.css
+++ b/lib/MultiSelection/MultiSelect.css
@@ -164,7 +164,6 @@
   display: flex;
   justify-content: space-between;
   align-items: center;
-  white-space: nowrap;
   list-style: none;
 
   &.valueChipSelected {


doytch:
Does this handle tag names that_are_very_long_but_don't_contain_white_space? If not, do we want it to? 

I've run into that issue with MCL columns and haven't been able to just say "don't call it that" to some users.

annamelnyk:
@doytch to handle such cases should be added style `word-break: break-word`, I guess it is a good point to handle it too. @JohnC-80 what do you think?

JohnC-80:
I'm happy to accept this as an initial change an iterate into a solution that will meet what @doytch is seeing... this component does have some style issues when it's in tight space which would be nice to resolve... I've got a storybook version look like this: 
image
But it's full of min-width settings which has mysteriously wonderful effect on flex-box behavior...

JohnC-80:
The ultimate thing that's nice to get working is an `overflow-wrap: break-word` property on the element containing the text... but to get this to behave appropriately, a guiding width of some kind needs to be given to flexbox, which `min-width` suffices nicely for... so the nice solution for these sorts of problems will probably be setting the `overflow-wrap: break-word;` property and then setting a min-width up the element heirarchy until all of the flex elements have the guidance they need to make the wrappin' happen.... this seems heavy and it's only been a cursory stab, but I'll try to commit something soonish that will fix this component's ailments - unless you'd like to work at this based on my description, @annamelnyk Thank you!

annamelnyk:
@JohnC-80, would not be better to make tags on full width?
image

PR #1226, opened 2020-03-30
STCOM-657 Resizeable Panes

JohnC-80:
## Pane resizing - User-level layout control.
This update allows users to resize panes by dragging the inner edges of groups of panes.

## App-level layout control
Apps can now control layout of panes beyond their defaultWidth props via the `onLayout`  and `initialLayouts` prop. [See paneset documentation for details](https://github.com/folio-org/stripes-components/tree/stcom-657-resizeable-panes/lib/Paneset#onlayout)

## Screenshots
resizepanes2
resizepanes3

## Approach
Each non-nested/root paneset renders a resize manager that renders resize handles for all panes within the pane hierarchy. Handles are invisibly rendered at pane edges, and appear on hover. The resize cursor appears if a user drags on an edge to indicate where the new edge of the pane will end up.
Internally, the paneset maintains a cache of widths so that a user's resizing is not reset if panes are change or removed and re-added.


doytch:
So stoked to see progress on this! 

BTW, I figure this may be annoying for you to test given your Windowsity, but [there's a weird bug on Safari with this](https://imgur.com/a/8qHbkon).

JohnC-80:
Thank you for bringing that to my attention, @doytch ! I'll have to put some code in place to properly disable the selection in Safari...

VictorSoroka1:
Tested on folio-testing and found buggy behavior when it comes to the shrinking third pane in settings. At some point, it becomes cut from the right side. It applies to the majority but not all of the third panes. Please take a look.

chrome-capture

VictorSoroka1:
And another onе when there are more than two panes and the Layout is opened.
chrome-capture (1)

doytch:
The Settings example looks like an example of what happens when a Pane incorrectly calculates its width to be `NaN`. I noticed these sometimes during testing and [specifically added exclusion logic to the persistence component for this.](https://github.com/folio-org/stripes-smart-components/pull/783/files#r413992812)

I'm not sure what's happening in the second example. You said Layout, but do you mean `<Layer>`? I'm not sure why there'd be resizing since it appears there's only a single Pane 🤔

PR #1230, opened 2020-04-01
Suggested style changes to Multiselect, MulticolumnList

annamelnyk:
# Description
Suggested style changes to improve components layout:

- wrapping of long words for Multiselect and MulticolumnList cells
- make filtered items on full width in Multiselect
- make MulticolumnList cells border single height 

# Screenshots
**Before**
image
image

**After**
image

image


lib/MultiColumnList/MCLRenderer.css
```diff
@@ -163,16 +163,28 @@
...
+    content: '';
+    position: absolute;
+    right: 0;
+    top: 5px;
+    bottom: 5px;
+    width: 1px;
+    background-color: var(--color-border-p2);
+  }
 
   &:last-child {
```
rasmuswoelk, line 189: The border should be removed on the last child.

image

lib/MultiColumnList/MCLRenderer.css
```diff
@@ -163,16 +163,28 @@
...
   text-align: left;
+  word-break: break-word;
+  position: relative;
+
+  &::after {
+    content: '';
+    position: absolute;
+    right: 0;
+    top: 5px;
+    bottom: 5px;
```
rasmuswoelk, line 184: We have a variable that has this value (5px): `--gutter-static-one-third`

lib/MultiSelection/MultiSelect.css
```diff
@@ -159,6 +159,7 @@
   display: inline-block;
   border: 1px solid var(--color-border);
   background-color: #efefef;
+  width: 100%;
```
rasmuswoelk, line 162: This is not ideal because we won't be taking advantage of the available space in most cases:

image

rasmuswoelk: I think the word-break alone is enough:

image

lib/MultiColumnList/MCLRenderer.css
```diff
@@ -163,19 +163,34 @@
...
+
+  &::after {
+    content: '';
+    position: absolute;
+    right: 0;
+    top: var(--gutter-static-one-third);
+    bottom: var(--gutter-static-one-third);
+    width: 1px;
+    background-color: var(--color-border-p2);
+  }
```
rasmuswoelk, line 187: Taking another look at this made me realize that this change will be applied to all MultiColumnList's. So while it will work fine for EditableList's specifically, it won't play nice with basic MCL's:

image

These borders should go from top to bottom without a gap.

rasmuswoelk: Would it maybe be possible to make these changes for the EditableList specifically instead?

lib/MultiColumnList/MCLRenderer.css
```diff
@@ -186,7 +201,6 @@
 .mclScrollable {
   flex-grow: 1;
   flex-shrink: 0;
-  overflow: auto;
```
rasmuswoelk, line 189: Why was this removed?

lib/MultiColumnList/MCLRenderer.css
JohnC-80: Could we A) break out borders/modifications into their own classes, keeping the more functional cell class intact and b) provide a prop that might allow you to set a custom class in addition to the existing classes? It seems like it would add the flexibility you need for the use-case of EditableList, @annamelnyk

annamelnyk:
@rasmuswoelk, @JohnC-80 I fixed comment and added `borderWithGaps` prop, setted to `false` by default, and to `true` for Editable components. Please take a look for [this](https://github.com/folio-org/stripes-smart-components/pull/765) also, it should be merged with this PR.

lib/MultiColumnList/MCLRenderer.css
annamelnyk: It shoud not be removed, I've tested one more time and moved it back

rasmuswoelk:
> @rasmuswoelk, @JohnC-80 I fixed comment and added `borderWithGaps` prop, setted to `false` by default, and to `true` for Editable components. Please take a look for [this](https://github.com/folio-org/stripes-smart-components/pull/765) also, it should be merged with this PR.

Hi @annamelnyk 

The `MultiColumnList` is already pretty bloated with a lot of different props so I think we need a solution that is a bit more versatile and can be re-used for other purposes as well.

 @JohnC-80 and I discussed a possible approach for this. We ended up with a prop that could be used to apply additional CSS classes for columns in an MCL. This prop could then be used to apply the custom CSS for the EditableList in stripes-smart-components since it's the only place currently (afaik) that we need borders with gaps. E.g.:

```js
className = {`${css.mclCell} ${showOverflow} ${props.getCellClasses ? props.getCellClasses(css.mclCellStyle) : css.mclCellStyle}`}
```

So the `getCellClases`-prop would receive the default style for the column so that it could either be extended:

```js
getCellClasses={defaultClass => `${defaultClass} ${css.myCustomClass}`
```

... or completely overwritten:
```js
getCellClasses={() => css.myCustomClass}
```

The idea is to separate the functional classes from the visual style classes so `css.mclCell` would still contain the functional CSS whereas e.g. a `css.mclCellStyle`-class would contain the default style for the cell – in this case, it would be the border, min-height, and line-height (if I remember correctly).

I hope this makes sense. Otherwise, feel free to write to me on Slack or here in the PR. I'm sorry for the delayed response to this and the additional work needed to achieve such a little change.. Almost nothing is simple in a Stripes project :-D

And @JohnC-80 if you have any additions or corrections then feel free to comment too 👍

JohnC-80:
@rasmuswoelk 's  comment is exactly what we need for this, @annamelnyk  - MCL already has many props - if we merge this as-is and eventually need something to affect padding, background-color, whathaveyou,  we would need another prop and another prop and another - so allowing the instance level to supply a function for setting the aesthetic class is most ideal.

annamelnyk:
@rasmuswoelk @JohnC-80 please take a look to changes, I am not sure that current implementation is what you expected. I was doubt if I should pass to Editable list prop `getCellClasses` as function, but I don't see how can it be done if all styles are in `MCLRenderer.css`. [Changes in EditableList](https://github.com/folio-org/stripes-smart-components/pull/765)

PR #1232, opened 2020-04-01
UIEH-869: Introduced `isDisabled` property to `NavListItem`

BogdanDenis:
## Purpose

Introduce a new `isDisabled` property to`NavListItem` to be able to render navigation links with disabled state

## Screenshots

ezgif-6-82055282bef4

### DIFF
--- a/lib/NavListItem/NavListItem.js
+++ b/lib/NavListItem/NavListItem.js
@@ -9,7 +9,7 @@ import Link from 'react-router-dom/Link';
 import css from './NavListItem.css';
 import NavListSectionContext from '../NavListSection/NavListSectionContext';
 
-const NavListItem = forwardRef(({ children, className, isActive, ...rest }, ref) => {
+const NavListItem = forwardRef(({ children, className, isActive, isDisabled, ...rest }, ref) => {
   let Element = 'button';
 
   if (rest.href) {
@@ -33,6 +33,7 @@ const NavListItem = forwardRef(({ children, className, isActive, ...rest }, ref)
                 css.NavListItem,
                 { [css.isActive]: active },
                 { [css.striped]: striped },
+                { [css.isDisabled]: isDisabled },
                 className
               )
             }
@@ -50,6 +51,7 @@ NavListItem.propTypes = {
   children: PropTypes.oneOfType([PropTypes.element, PropTypes.string, PropTypes.array]),
   className: PropTypes.string,
   isActive: PropTypes.bool,
+  isDisabled: PropTypes.bool,
 };
 
 NavListItem.displayName = 'NavListItem';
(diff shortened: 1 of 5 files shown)


JohnC-80:
Thank you @BogdanDenis - I think we may have to note in documentation with this that, while this does apply a disabled styling to the NavListItem, the children of the item will still need to have a `disabled` prop or something applied so that they're out of tab order (unreachable by tabbing through with the keyboard)

JohnC-80:
Since the props are passed down to the rendered element, `disabled` alone as a prop should work for buttons.. for anchors, the style can be applied... despite the mouse events being turned off, this will still need something to account for keyboard navigation - perhaps an `aria-disabled` attribute added to anchors in case of this prop since 'disabled' doesn't work there.

JohnC-80:
on second thought, if a disabled prop is applied, the control can just render a button with aria-disabled (and no href) which should announce fine for screen readers.... the style shouldn't be necessary, then as 'disabled' has an effect on button styling.

zburke:
@JohnC-80, As I noted on the ticket, the Folio way is usually to hide things you can't use/don't have access to. 

Although normally we think of that in terms of permissions (e.g. if you don't have "create Foo" permission you don't see the "New Foo" button) here I view it as the same thing: you haven't initialized the Foo, so you don't have "edit Foo" capability, so you don't see the "Edit Foo" link.

JohnC-80:
Goodness knows our UI's could certainly use 'less'... It's definitely better to completely hide the items in the case of permissions... I'm not sure that completely covers the cases for reasons to disable a navlist item, though so I'm fine with adding the capability to the generic component since it's not a boatload of effort. I do wish people held more tightly to that mantra... too many times I've had to root-cause cases down to us dangling lack of permissions in front of people.

lib/NavListItem/NavListItem.js
```diff
@@ -9,7 +9,7 @@ import Link from 'react-router-dom/Link';
 import css from './NavListItem.css';
 import NavListSectionContext from '../NavListSection/NavListSectionContext';
 
-const NavListItem = forwardRef(({ children, className, isActive, ...rest }, ref) => {
+const NavListItem = forwardRef(({ children, className, isActive, isDisabled, ...rest }, ref) => {
```
rasmuswoelk, line 12: I think we should try to lean towards the default attribute naming of `disabled` instead of `isDisabled` in this case. Then the prop could be applied directly onto the nav item when it renders a `<button>` and we would achieve native browser behavior.

BogdanDenis:
This change is no longed needed anymore due to changed requirements.
However, if you think that this would be a good addition then I'll be happy to update the PR

PR #1238, opened 2020-04-13
STCOM-670: expose filter delimiter

mkuklis:
https://issues.folio.org/browse/STCOM-670

In order to correctly handle the issue described in: https://issues.folio.org/browse/STSMACOM-329

We need to replace current filter delimiter (`,`) and expose it via stripes-components.

This PR does just that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Fix issue with `initialStatus` prop on Accordions not working.
 * Fix bug with impossibility to use mouse to set Associated Service Point for Fee/Fine Owner. Refs UIU-1539.
 * Introduce a new filter config property called `operator`. Refs STCOM-662.
+* Expose `FILTER_GROUP_SEPARATOR` and `FILTER_SEPARATOR` for splitting/joing filters. Refs STCOM-670.
 * a11y improvements for form components and update primary color. Refs STCOM-658.
 
 ## [6.1.0](https://github.com/folio-org/stripes-components/tree/v6.1.0) (2020-03-16)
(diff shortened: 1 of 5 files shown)


JohnC-80:
Pardon my ignorance - is there significance to it being ALL_CAPS ? It really does shout at me from the code... if that was the intent, good job!

mkuklis:
@JohnC-80 I'm happy to change that. I initially exposed a function called `getFilterDelimiter` but that felt like an overkill. Would `filterDelimiter` work better?

zburke:
```
#ifdef MY_FIRST_LANGUAGE_WAS_C
        printf("Sorry for shouting at the JavaScript kids!");
        printf("p.s. K&R braces and %d character tab stops forever!!!!!", 8);
#endif
```

zburke:
Any reason not to
```
export const FILTER_GROUP_SEPARATOR = '.';
```
at the same time?

JohnC-80:
I don't really have a preference, personally - but it may be something we want to adopt with consistency. If it's a system-wide/top level name we can probably justify all caps with that - the convention just may need to be in a document somewhere or something.  The lack of blending in with the rest of the local code is good for such entities, I think.

zburke:
Should we close this in light of folio-org/ui-users/pull/1258? Or change `\t` to `,` and then merge it on the principle that future work would want to take advantage of the constant?

doytch:
> Or change `\t` to `,` and then merge it on the principle that future work would want to take advantage of the constant?

I'm in favour of this. I would've used the constant in my code had it been available.

mkuklis:
Ok guys I exposed both `FILTER_GROUP_SEPARATOR ` and `FILTER_SEPARATOR`.  The `FILTER_SEPARATOR` should be set back to `,`.

PR #1242, opened 2020-04-16
STCOM-618: Added 'centerContent'-prop for <Pane>

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/STCOM-618

## Changes
- Added `centerContent `-prop for `<Pane>` that adds a max-width and centers the content of the pane
- Added a class name for `<Layout>` to make it possible to apply the same CSS for custom use-cases
- Added example under `<Pane>` and updated readme

## Screenshot
image


JohnC-80:
Perhaps just 'centerContent' ?- More exciting names came to mind... "deadCenter," "middleEarth," "fullMalcolm," (Has Malcolm in the Middle aired over there?) I'd say that it's got a clarity advantage over the catchy-sounding 'containerize' - I can see it making the association with 'centering the content' at larger widths. We could explain in detail in the docs.

rasmuswoelk:
@JohnC-80 Fine by me :) I would prefer `fullMalcolm` because I used to love that show when I was a kid but `centerContent` will do.

PR #1248, opened 2020-04-23
STDTC-8: Allowed `to`, `href` and `labelStrings` props to be passed as function to `defaultRowFormatter`

VictorSoroka1:
## Purpose

The idea is to extend [defaultRowFormatter](https://github.com/folio-org/stripes-components/blob/master/lib/MultiColumnList/defaultRowFormatter.js) to make `to`, `href` also as functions to accept `rowData.id` as a parameter. Usually, upon clicking on row item there is a need to navigate to a specific item route and for this, the `id` is needed. Because of existing limitation, other apps usually need to define their own row formatted. ([example](https://github.com/folio-org/ui-users/blob/master/src/views/UserSearch/UserSearch.js#L217)). Additionally, @doytch suggested having a similar possibility for `labelStrings`.

### DIFF
--- a/lib/MultiColumnList/defaultRowFormatter.js
+++ b/lib/MultiColumnList/defaultRowFormatter.js
@@ -1,29 +1,53 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
+import { isFunction } from 'lodash';
 
 const propTypes = {
   cells: PropTypes.arrayOf(PropTypes.element),
-  labelStrings: PropTypes.arrayOf(PropTypes.string),
+  labelStrings: PropTypes.oneOf([PropTypes.arrayOf(PropTypes.string), PropTypes.func]),
   rowClass: PropTypes.string,
+  rowData: PropTypes.object,
   rowIndex: PropTypes.number,
   rowProps: PropTypes.object,
   rowWidth: PropTypes.number,
 };
 
-function defaultRowFormatter({ rowIndex, rowClass, cells, rowProps, labelStrings }) {
+function defaultRowFormatter(row) {
+  const {
+    rowIndex,
+    rowClass,
+    rowData,
+    cells,
+    rowProps: initialRowProps,
+    labelStrings: initialLabelStrings,
+  } = row;
+
   // Default row element
+  const rowProps = { ...initialRowProps };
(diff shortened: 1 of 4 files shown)


lib/MultiColumnList/defaultRowFormatter.js
```diff
@@ -1,30 +1,52 @@
...
   // Render an anchor tag if an "href"-prop is provided
   if (rowProps.href) {
     Element = 'a';
+
+    if (isFunction(rowProps.href)) {
+      rowProps.href = rowProps.href(rowData.id);
+    }
   }
 
+  const labelStrings = isFunction(rowProps.labelStrings)
```
VictorSoroka1, line 46: @doytch, did you mean this way you need `labelString` stuff?

doytch: Sort of but not quite. [Here's an example of a custom rowFormatter I have right now](https://github.com/folio-org/ui-erm/blob/d3a33951db15adca300aa4554eeedc0321e57daa/src/components/views/Agreements.js#L297). Technically I'm setting `data-label` there instead of `aria-label` but the point is the same.

This implementation wouldn't work because:
- I get `[Object object]` for the column headers. Not a big deal since I suppose I can switch to `formatMessage()`.
- It's not possible for me to selectively exclude certain indices of `labelStrings` unless I count on magic numbers (eg, I know that "Coverage" is gonna be at index 5). Some of our columns simply can't/shouldn't be reduced to text so I want to be able to build up my label manually from `rowData` itself.
  -  I do understand why some people might want to _not_ build up from `rowData` and use something else though, so maybe just passing along [the entire `row` object that's passed to the `rowFormatter`](https://github.com/folio-org/ui-erm/blob/d3a33951db15adca300aa4554eeedc0321e57daa/src/components/views/Agreements.js#L289) would be safest and most future-proof.

VictorSoroka1: `rowProps.labelStrings(row)`?

doytch: I think that'd be perfect 👍

VictorSoroka1: Done.

PR #1250, opened 2020-04-27
MCL: Remove labelstrings from DOM element

doytch:
Need to remove this from the object we'll spread onto element we're creating to prevent React from complaining with `console.error`s.

We could do some fancy `filter()` and list of keys that get excluded...but given that it's just the single solitary key for now I wanted to keep it simple and obvious.


VictorSoroka1:
Good catch.

PR #1251, opened 2020-04-27
STCOM-673, STCOM-674 Pane resize/layer bugfix

JohnC-80:
## Problem
See [STCOM-673](https://issues.folio.org/browse/STCOM-673)
Pane resize handles rendering above full width/height non-Pane siblings. Any module using a Layer was affected... some modules use a Layer-like component - this should cover that as well.

## Approach
Using a mutationObserver to see when children are added/removed from the paneset. Added children are compared against the full container - if they overlap, the resize handles are hidden and the overlapping element's info is saved. As elements are removed, it checks those elements for a match to the saved overlapping element's info. The resize handles are restored when the element is added then removed...


zburke:
Not sure what the STCOM-647 reference in the title here is all about. Maybe that's the wrong project key?

lib/Paneset/Paneset.js
```diff
@@ -105,6 +107,15 @@ class Paneset extends React.Component {
 
   getContainer = () => this.container.current;
 
+  getTopmostContainer = () => {
+    const { paneset, isRoot } = this.props;
+    if (!isRoot && paneset) {
+      return paneset.getTopmostContainer();
+    } else {
```
VictorSoroka1, line 114: Else can be omitted.

lib/Layer/Layer.js
```diff
@@ -24,13 +25,19 @@ class Layer extends React.Component {
     container: componentOrElement,
     contentLabel: PropTypes.string,
     enforceFocus: PropTypes.bool,
+    inRootSet: PropTypes.bool,
```
VictorSoroka1, line 28: Could you please extend the readmee for the component with the information about `inRootSet ` and `resizer` props.

lib/Layer/Layer.js
```diff
@@ -24,13 +25,19 @@ class Layer extends React.Component {
...
       getContainer: PropTypes.func,
+      getTopmostContainer: PropTypes.func,
       handleClose: PropTypes.func,
       registerPane: PropTypes.func,
       removePane: PropTypes.func,
     }),
+    resizer: PropTypes.shape({
+      resume: PropTypes.func,
+      suspend: PropTypes.func,
+    }),
```
VictorSoroka1, line 40: Seems like this one should be required, right as I see in the `handleOpen` the `resizer.resume();` is used without checks.

lib/Paneset/Paneset.js
```diff
@@ -56,11 +57,12 @@ class Paneset extends React.Component {
         registerPane: this.registerPane,
         removePane: this.removePane,
         getContainer: this.getContainer,
+        getTopmostContainer: this.getTopmostContainer,
       },
       panes: [],
       layoutCache: props.initialLayouts,
       style: initStyle,
-      changeType: 'init',
+      changeType: 'init', // eslint-disable-line react/no-unused-state
```
VictorSoroka1, line 65: Could you please clarify whether comment is needed? Fur curiosity I tried without it and seems like it works just fine.
```suggestion
      changeType: 'init',
```

lib/Paneset/PaneResizeContainer.js
```diff
@@ -44,14 +113,16 @@ const PaneResizeContainer = ({ isRoot, children, onElementResize }) => {
   useEffect(() => { // eslint-disable-line consistent-return
     if (isRoot) {
       window.addEventListener('resize', debouncedUpdate);
+
       return function cleanup() {
+        if (observer.current) observer.current.disconnect();
```
VictorSoroka1, line 118:
```suggestion
        observer.current?.disconnect();

```

lib/Paneset/PaneResizeContainer.js
```diff
@@ -1,30 +1,100 @@
...
+    if (isRoot) {
+      const observerOptions = {
+        childList: true,
+      };
+
+      if (parentElement) {
+        observer.current = new MutationObserver(observerCallback);
+        observer.current.observe(parentElement, observerOptions);
+      }
+    }
```
VictorSoroka1, line 96: Should we have some kind of cleanup logic there as well? In case if some of the props changes `[parentElement, isRoot, observerCallback]`, the new MutationObserver will be created without destroying the previous. I that there is the cleanup logic on the unmounting only.

lib/Paneset/PaneResizeContainer.js
```diff
@@ -1,30 +1,100 @@
...
-  )).current;
+  const observer = useRef();
+
+  const observerCallback = useRef(
+    (mutationList) => {
+      setBlocking(curBlocker => {
+        let blocker = curBlocker;
+        mutationList.forEach((m) => {
+          if (m.type === 'childList') {
+            if (m.addedNodes.length > 0) {
```
VictorSoroka1, line 30: This check can be ommited, as in case the node list is empty forEach would not throw an error.

lib/Paneset/PaneResizeContainer.js
```diff
@@ -1,30 +1,100 @@
...
+                  blocker = {
+                    id: n.id,
+                    class: n.className,
+                    style: n.getAttribute('style')
+                  };
+                }
+              });
+            }
+
+            if (m.removedNodes.length > 0 && typeof blocker === 'object') {
```
VictorSoroka1, line 54:
```suggestion
            if (typeof blocker === 'object') {

```

lib/Paneset/PaneResizeContainer.js
```diff
@@ -1,30 +1,100 @@
...
+
+  const observerCallback = useRef(
+    (mutationList) => {
+      setBlocking(curBlocker => {
+        let blocker = curBlocker;
+        mutationList.forEach((m) => {
+          if (m.type === 'childList') {
+            if (m.addedNodes.length > 0) {
+              // check against full container bounds
+              m.addedNodes.forEach((n) => {
```
VictorSoroka1, line 32: Would it be just enough to stop when at least the one is found? It seems like in the current situation only the last (in case if there are multiple full-width ones) will be stored.

```suggestion
              m.addedNodes.some((n) => {
```

lib/Paneset/PaneResizeContainer.js
JohnC-80: I actually attempted this previously and eslint complained about it being a statement rather than an assignment...

JohnC-80: Even sonarcube complained at me!

lib/Paneset/PaneResizeContainer.js
JohnC-80: Yeah, good idea..

lib/Paneset/PaneResizeContainer.js
JohnC-80: Cool.. I've typically used `.some` for it's boolean return, but I suppose it is like a self-terminating forEach.

lib/Paneset/PaneResizeContainer.js
VictorSoroka1: Ah, got it. It is disappointing as we started to use this feature very much now. So feel free to use it in the way you prefer more. Let's touch this maybe on stripes arch.

lib/Paneset/PaneResizeContainer.js
VictorSoroka1: Yeah, but in this case, it is needed to return some bool in order for it to understand whether it needs to continue. Maybe the easier way would be to use `.find` which will return the first matched entry.

JohnC-80: So there was another slight gotcha - The `addedNodes` and `removedNodes` are NodeLists... had to use `Array.from()`.

PR #1256, opened 2020-05-04
STSMACOM-256: Unregister Accordion from AccordionSet when unmounted

BogdanDenis:
## Purpose
Implementation of drag and drop for Custom Fields requires dragged Custom Field accordions to be rendered via Portals for proper positioning on the page. Reason for this is that `react-beautiful-dnd` adds `position: fixed` to accordions, but some ancestors in the DOM tree have relative positioning, so `fixed` is not relative to the window.
When an `<Accordion>` component is unmounted `<AccordionSet>` does not remove it's Id from list of all accordions, so we have to remove it so list of accordions contains correct information

## Issue
[STSMACOM-267](https://issues.folio.org/browse/STSMACOM-267)


JohnC-80:
Awesome! Thank you @BogdanDenis this is in the queue to tackle. How does this work with the `AccordionStatus` component? Will it also remove the accordion from there?

BogdanDenis:
> Awesome! Thank you @BogdanDenis this is in the queue to tackle. How does this work with the `AccordionStatus` component? Will it also remove the accordion from there?

I'm not sure what you mean by 'remove the accordion', could you explain please?

JohnC-80:
@BogdanDenis Sorry, I mean to say "unregister" the accordion from the `<AccordionStatus>` component since a state will be tracked there.
The common use case within apps is to have an `<AccordionStatus>` wrapping somewhere outside of the the `<AccordionSet>` - in that case, `<AccordionSet>` catches the registration and `<AccordionStatus>` handles the open/closed state and can be used to set up patterns like an expand/collapse without requiring a local state.

BogdanDenis:
@JohnC-80 Got it now, thank you!
I've checked this functionality in ui-invoices, where `<AccordionStatus>` is used a lot. Unmounting of an accordion doesn't remove it's status from state of `<AccordionStatus>`.

PR #1264, opened 2020-05-08
AccordionSetContext: Prevent overwrite w/ undefined

doytch:
When using an `AccordionStatus`, you shouldn't pass `open` and `onToggle` to your `Accordion`s because they'll get it from this context.

However, sometimes you're not in control of rendering your `Accordion` directly, like in the case of `NotesSmartAccordion`. In these cases, they always pass in `open` and `onToggle`, even if it's undefined. The fact that they passed in an `undefined` value is different from the _prop itself_ being undefined, so spreading `this.props` here _overwrites_ the open/onToggle from context with `undefined`. This breaks the expand/collapse-all of the notes accordion.

We can fix this by nullish coalescing to either the passed-in value or the context value.


doytch:
I would hope that the future developer who is messing around with that logic reads this PR, but yeah, let's leave a comment there just in case.

And hello, Future Developer! Thank you for investigating the origin of that logic. You are a boon to the project and I applaud your conscientiousness.

PR #1265, opened 2020-05-08
react-hot-loader is a direct dep

zburke:
`react-hot-loader` is not provided by the platform, hence it should not
be a peer-dep. Resolves the following build warning:
```
warning " > @folio/stripes-components@7.0.0" has unmet peer dependency "react-hot-loader@^4.0.0".
```


zburke:
I _thought_ I had already done this. Why couldn't I find #1263 when I opened this??!?

PR #1267, opened 2020-05-13
Updated MultiSelection css for small size of input field

Vladyslav-Velytskyi:
## Purpose
 For `MultiSelection` which size is less than 180px, selectionToggleButton starts to get out from input field block. It's hepend when we try to resize the pane which has `MultiSelection`
image

 or when we add this component on edit page with a small grid layout (example [CustomField - MultiSelect on User record edit page](https://github.com/folio-org/stripes-smart-components/pull/808))
image

The most accurate way to solve the problem so as not to change the structure of the HTML is to add the css property `max-width`

## Screenshots
image
image

### DIFF
--- a/lib/MultiSelection/MultiSelect.css
+++ b/lib/MultiSelection/MultiSelect.css
@@ -23,7 +23,7 @@
 
 .multiSelectControlGroup {
   display: flex;
-  flex: 2 0 90%;
+  flex: 1;
   flex-wrap: wrap;
 }
 


lib/MultiSelection/MultiSelect.css
```diff
@@ -22,6 +22,7 @@
 }
 
 .multiSelectControlGroup {
+  max-width: calc(100% - 22px);
   display: flex;
   flex: 2 0 90%;
```
rasmuswoelk, line 27: I think a simpler solution to this would be to set `flex: 1;` so that `.multiSelectControlGroup` takes up any remaining space.

Vladyslav-Velytskyi: Thanks, your suggestion looks better. Done.

PR #1275, opened 2020-05-20
MCL Columnwidth hints

JohnC-80:
For more flexibilty in column sizing, This PR allows developers to provide an object with `min` and `max` keys...
MCL will discard any widths below the `min`, and will truncate column widths that are greater than the `max`
Values are in pixel...
ex:
```
this.columnWidths = {
      active: '10%',
      title: { min: 50, max: 100 }
    };

...
<MultiColumnList
  columnWidths = { this.columnWidths }
/>
```

## Additionally

The last column should fill the remaining width - setting the style to `{ min-width: columnWidth, flex: 'auto' }`


zburke:
Replaced by #1378.

PR #1280, opened 2020-06-01
STSMACOM-340: User Record. Custom Fields. MultiSelection. Incorrect validation work for required field

BogdanDenis:
## Purpose
Show validation error on submit when a required `<MultiSelection>` field is empty.

## Approach
Currently, `<UserForm>` relies on `<input required>` attribute and native browser validation to block form submission and showing of error messages when a required input is empty.
These changes add an `<input>` element that stores current `<MultiSelection>` component's value (formatted as a comma separated list). This `<input>` is hidden from users and cannot be accessed via mouse or keyboard.

## Screenshots

image

image


lib/MultiSelection/MultiSelection.js
```diff
@@ -384,6 +389,14 @@ class MultiSelection extends React.Component {
               disabled,
             };
 
+            const inputProps = {
+              ariaLabelledBy,
+              id: `multiselect-input-${uiId}`,
+              required,
+              getInputProps,
+              value: Array.isArray(value) ? value.map(item => item?.value).join(',') : value,
```
JohnC-80, line 397: There isn't exactly a guarantee that the values are all objects or contain a `value` property - you can make use of the function from 'itemToString' prop to shore this up... See the example in the Readme: https://github.com/folio-org/stripes-components/tree/master/lib/MultiSelection#example

PR #1281, opened 2020-06-02
UIEH-899: Settings > eholdings > KB credentials page | UI updates

BogdanDenis:
## Purpose
Change submit KB form button text from 'Save & close' to 'Save'

Related PR in `ui-eholdings`: https://github.com/folio-org/ui-eholdings/pull/1063

## Issue
[UIEH-899](https://issues.folio.org/browse/UIEH-899)

### DIFF
--- a/translations/stripes-components/en.json
+++ b/translations/stripes-components/en.json
@@ -32,6 +32,7 @@
   "addressSection": "Address Section",
   "cancel": "Cancel",
   "submit": "Submit",
+  "save": "Save",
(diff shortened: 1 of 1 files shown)


BogdanDenis:
> You can add `Save` to stripes-components if you want, but `stripes-core.button.save` [already exists in stripes-core](/folio-org/stripes-core/blob/master/translations/stripes-core/en.json#L23).

Thanks, @zburke , I'll use that instead

PR #1283, opened 2020-06-03
document final-form's need for type="radio"

zburke:
Follow-up to #1282 which documented this need for checkboxes.

@JohnC-80, do you know if we need to do something special/similar when `<Field>` receives `component={RadioButtonGroup}`?

### DIFF
--- a/lib/RadioButton/readme.md
+++ b/lib/RadioButton/readme.md
@@ -62,5 +62,6 @@ import { RadioButton } from '@folio/stripes/components';
     value="ripe"
   />
 ```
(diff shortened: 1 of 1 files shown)


zburke:
I think so. From https://final-form.org/docs/react-final-form/types/FieldProps#type

> If set to `"checkbox"` or `"radio"`, React Final Form will know to manage your values as a checkbox or radio button respectively. Results in a `checked` boolean inside the `input` value given to your render prop.

MikeTaylor:
:thumbsup:

PR #1284, opened 2020-06-04
Timepicker timezone documentation fix

MikeTaylor:
This was claimed to default to 'UTC' but actually the prop itself has no default value, and the behaviour of the Timepicker defaults to functioning within the timezone of the prevailing locale (`props.intl.timeZone`).

Added a note to say that passing `timeZone='UTC'` allows you to treat the time as an absolute time of day.


MikeTaylor:
@JohnC-80 Any idea why this one-line documentation change is suffering from test errors?

MikeTaylor:
> @JohnC-80 Any idea why this one-line documentation change is suffering from test errors?

Well, whatever the cause was, it's gone away. Just another testing mystery.

PR #1287, opened 2020-06-05
Textfield end-controls behavior in that other browser we use half the time...

JohnC-80:
Firefox inherently sets a min-width on inputs which causes them to not shrink smaller than a certain size. This change overrides that so that icons/endControls aren't shoved out of their container... Timepicker used for example... Datepicker/Textfield also affected....
One less thing we'll have to fix later...

## Before
image

## After 
image


doytch:
Wait so, what's the change here? I only see the storybook code, and nothing in the usage is jumping out at me.

PR #1288, opened 2020-06-06
avoid setState in unmounted components

zburke:
A promise may return after a component has unmounted. In this case, it
should avoid calling `setState`.

### DIFF
--- a/lib/PasswordStrength/PasswordStrength.js
+++ b/lib/PasswordStrength/PasswordStrength.js
@@ -68,7 +68,7 @@ class PasswordStrength extends Component {
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


lib/PasswordStrength/PasswordStrength.js
```diff
@@ -63,16 +63,22 @@ class PasswordStrength extends Component {
   }
 
   componentDidMount() {
+    this._isMounted = true;
     this.initPasswordStrength();
   }
 
+  componentWillUnmount() {
+    this._isMounted = true;
```
JohnC-80, line 71:
```suggestion
    this._isMounted = false;
```
false here, ya?

PR #1289, opened 2020-06-08
Ensure TextField and TextArea always have their <label> hooked up to the <input>

doytch:
Per [ERM-898](https://issues.folio.org/browse/ERM-898) and [ERM-899](https://issues.folio.org/browse/ERM-899), textfields and textareas fail axe a11y checks when an `id` is not provided. 

Use the/create a unique id and use that for the label's `for` and input's `id` attributes.

Per [ERM-900](https://issues.folio.org/browse/ERM-900), axe complains that `NoValue` sets an invalid `role` attribute. The `role` is unnecessary here so I've removed it.

### DIFF
--- a/lib/NoValue/NoValue.js
+++ b/lib/NoValue/NoValue.js
@@ -14,8 +14,6 @@ function NoValue({ ariaLabel = null }) {
     <FormattedMessage id="stripes-components.noValue.noValueSet">
       {message => (
         <span
-          /* eslint-disable-next-line jsx-a11y/aria-role */
-          role="text"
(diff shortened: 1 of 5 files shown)


JohnC-80:
A place where `role="text"` might be appropriate - 
```
<p>
  Who you going to call?
  <img src="ghostbusters-logo.jpg" alt="GhostBusters!" role="text">
</p>
```
Which is, generally an example of not-so-great a11y practice in the first place (text within images is a NO.)

Essentially, it will make the screen-reader believe the element is something it isn't.

PR #1290, opened 2020-06-08
Textfield/input not shrinking in flex-row in FF...

JohnC-80:


MikeTaylor:
Fantastic! Many thanks, @JohnC-80!

PR #1294, opened 2020-06-15
STCOM-614 localize currency values

zburke:
Fixes [STCOM-614](https://issues.folio.org/browse/STCOM-614)

### DIFF
--- a/lib/CurrencySelect/CurrencySelect.js
+++ b/lib/CurrencySelect/CurrencySelect.js
@@ -1,10 +1,20 @@
 import React from 'react';
-import { currenciesOptions } from '../../util/currencies';
+import { useIntl } from 'react-intl';
+import { filteredCurrencies } from '../../util/currencies';
 
 import Select from '../Select';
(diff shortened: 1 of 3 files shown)


lib/CurrencySelect/CurrencySelect.js
```diff
@@ -1,10 +1,20 @@
...
 
 import Select from '../Select';
 
-const CurrencySelect = props => (
-  <Select {...props} dataOptions={currenciesOptions} />
-);
+const CurrencySelect = props => {
+  const intl = useIntl();
+
+  const currenciesOptions = filteredCurrencies.map(c => {
```
aliaksei-chumakou, line 10: maybe it's a good idea to use `useMemo()` here? it looks like static data

zburke: If you want to add that, it'd be great! I'm not even supposed to be working on this right now, so am unlikely to tweak it further.

PR #1295, opened 2020-06-15
STCOM-291 implement CountrySelection

zburke:
Provide a country-picker component.

Refs [STCOM-291](https://issues.folio.org/browse/STCOM-291)

### DIFF
--- a/lib/CountrySelection/CountrySelection.js
+++ b/lib/CountrySelection/CountrySelection.js
@@ -1,0 +1,20 @@
+import React from 'react';
+import { useIntl } from 'react-intl';
+import countries from '../../util/countries';
+
+import Selection from '../Selection';
+
+const CountrySelection = props => {
+  const intl = useIntl();
+
+  const sortedCountries = countries.map(c => {
+    return {
+      label: intl.formatDisplayName(c.alpha2, { type: 'region' }),
+      value: c.alpha2,
+    };
+  }).sort((a, b) => a.label.localeCompare(b.label));
+
+  return <Selection {...props} dataOptions={sortedCountries} />;
+};
+
+export default CountrySelection;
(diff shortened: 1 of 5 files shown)


lib/CountrySelection/CountrySelection.js
```diff
@@ -0,0 +1,20 @@
...
+
+const CountrySelection = props => {
+  const intl = useIntl();
+
+  const sortedCountries = countries.map(c => {
+    return {
+      label: intl.formatDisplayName(c.alpha2, { type: 'region' }),
+      value: c.alpha2,
+    };
+  }).sort((a, b) => a.label.localeCompare(b.label));
```
doytch, line 15: [useMemo](https://reactjs.org/docs/hooks-reference.html#usememo)
```suggestion
  const sortedCountries = useMemo(() => (
    countries
      .map(c => ({
        label: intl.formatDisplayName(c.alpha2, { type: 'region' }),
        value: c.alpha2,
      })
      .sort((a, b) => a.label.localeCompare(b.label))
  ), countries)
```

zburke:
ESLint wants 
```
  const sortedCountries = useMemo(() => (
    countries
      .map(c => ({
        label: intl.formatDisplayName(c.alpha2, { type: 'region' }),
        value: c.alpha2,
      }))
      .sort((a, b) => a.label.localeCompare(b.label))
  ), [intl]);
```
Does that make sense to you @doytch, @aliaksei-chumakou, that `intl` would be a required dependency but `countries` is _not_?

doytch:
Mmm, no that wouldn't make sense. You want to memoize based on the value of `countries` so that if that array changes, it'll recompute the translated-and-sorted list.

Looking at the parse errors I'm pretty sure I probably [messed up a bracket here - it should be one extra closing paren](https://github.com/folio-org/stripes-components/pull/1295/commits/052903967c0a447ac249684ca676e65a5961ad0e#diff-b0145d7be8b79331fa954adc95ec6d20R15). I didn't run that code, I just wrote the suggestion in Github 🙃

PR #1297, opened 2020-06-16
UIU-1674: fix multiselect carat not aligned when there is not enough space

MaksymDryha:
Ticket:  https://issues.folio.org/browse/UIU-1674

## Before
image
image
image

## After
image
image
image


rasmuswoelk:
LGTM 👍 

I noticed that this also fixes an issue where the placeholder is being cut off:

image

PR #1299, opened 2020-06-16
UIOR-522 fix aria-owns attribute value

aliaksei-chumakou:
https://issues.folio.org/browse/UIOR-522
Accessibility Error: ARIA attributes must conform to valid values


aliaksei-chumakou:
https://github.com/folio-org/stripes-components/pull/1299/files?diff=unified&w=1

PR #1302, opened 2020-06-17
STSMACOM-362: Settings | Users | Custom fields | Ability to re-order options

BogdanDenis:
## Purpose
Custom Fields settings uses `<MultiColumnList>` component to render it's options and we need to support changing order of these options via drag-and-drop.
Similar functionality was previously implemented in `ui-requests` to order request queues.
So as we are beginning to have similar implementations of drag-and-drop across several applications I believe we should consider moving MCL related drag-and-drop code to shared `stripes-components` repository.

## Approach

`<SortableList>` is a wrapper for `<MultiColumnList>` component. It uses `react-beautiful-dnd` library to extend it with drag-and-drop functionality. `<SortableList>` defines drag-and-drop context around `<MultiColumnList>` and passes a custom `rowFormatter` to MCL, which itself acts as a wrapper around each row to make them draggable within a single MCL.

#### TODOS and Open Questions

- Failed tests. These are caused by `react-beautiful-dnd` printing warnings that contain an emoji (👷). Warnings are printed because `<SortableList>` is unmounted before `react-beautiful-dnd` has had a chance to announce it's screen reader messages. I wonder if anybody knows how we can filter/disable warnings in tests console output. @mkuklis I believe you've worked on implementing drag-and-drop in `ui-requests`? Have you had these issues?

### DIFF
--- a/lib/SortableList/DraggableRow.js
+++ b/lib/SortableList/DraggableRow.js
@@ -1,0 +1,65 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+
+import css from './DraggableRow.css';
+
+const getItemStyle = (draggableStyle) => {
+  return {
+    userSelect: 'none',
+    ...draggableStyle,
+  };
+};
+
+const propTypes = {
+  cells: PropTypes.arrayOf(PropTypes.element),
+  provided: PropTypes.object,
+  rowClass: PropTypes.string,
+  rowIndex: PropTypes.number,
+  snapshot: PropTypes.object,
+};
+
+const DraggableRow = ({
+  snapshot,
+  provided,
+  rowIndex,
+  rowClass,
+  cells,
+}) => {
+  const usePortal = snapshot.isDragging;
+  const classNames = [rowClass];
+
+  if (usePortal) {
+    classNames.push(css.DraggableRow);
+  }
+
+  const Row = (
+    <div
+      id={`row-${rowIndex}`}
+      data-test-draggable-row
+      ref={provided.innerRef}
+      {...provided.draggableProps}
+      {...provided.dragHandleProps}
+      key={`row-${rowIndex}`}
+      className={classNames}
+      role="row"
+      tabIndex="0"
+      style={getItemStyle(
+        provided.draggableProps.style,
+      )}
+    >
+      {cells}
+    </div>
+  );
+
+  if (!usePortal) {
+    return Row;
+  }
+
+  const container = document.getElementById('ModuleContainer');
+  return ReactDOM.createPortal(Row, container);
+}
+
+DraggableRow.propTypes = propTypes;
+
+export default DraggableRow;
--- a/lib/SortableList/DraggableRow.js
+++ b/lib/SortableList/DraggableRow.js
@@ -58,7 +58,7 @@ const DraggableRow = ({
 
   const container = document.getElementById('ModuleContainer');
   return ReactDOM.createPortal(Row, container);
-}
+};
 
 DraggableRow.propTypes = propTypes;
 
--- a/lib/SortableList/tests/SortableList-test.js
+++ b/lib/SortableList/tests/SortableList-test.js
@@ -13,11 +13,11 @@ describe('SortableList', () => {
   const onDragEndFake = sinon.fake();
   const onDragStartFake = sinon.fake();
   const onDragUpdateFake = sinon.fake();
-  
+
   beforeEach(async () => {
     await mountWithContext(
       <>
-        <div id="ModuleContainer"></div>
+        <div id="ModuleContainer" />
         <SortableList
           onDragEnd={onDragEndFake}
           onDragStart={onDragStartFake}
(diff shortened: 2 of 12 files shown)


JohnC-80:
Are there plans to use this in places besides Custom Fields? If the smart component needs it sooner rather than later, it might be quicker to keep it there for a bit... If it's here, it has to be prepared to be used by all FOLIO UI modules and handle situations that are beyond the scope of the story it's created to serve...

Aside from that -

I'm not sure the name "SortableList" really tells me this is a 'drag and drop' implementation since "Sort" is such a general term. "DNDList" might be better. Just a minor suggestion.

lib/SortableList/DraggableRow.js
```diff
@@ -0,0 +1,65 @@
...
+  rowIndex,
+  rowClass,
+  cells,
+}) => {
+  const usePortal = snapshot.isDragging;
+  const classNames = [rowClass];
+
+  if (usePortal) {
+    classNames.push(css.DraggableRow);
+  }
```
Vladyslav-Velytskyi, line 34: In this module we have `classnames` package, please use it ([documentation](https://www.npmjs.com/package/classnames)), it could help omit `const classNames = [rowClass]`

lib/SortableList/DraggableRow.js
```diff
@@ -0,0 +1,65 @@
...
+      {cells}
+    </div>
+  );
+
+  if (!usePortal) {
+    return Row;
+  }
+
+  const container = document.getElementById('ModuleContainer');
+  return ReactDOM.createPortal(Row, container);
```
Vladyslav-Velytskyi, line 60: Please add empty string before `return`

lib/SortableList/DraggableRow.js
```diff
@@ -0,0 +1,65 @@
...
+      ref={provided.innerRef}
+      {...provided.draggableProps}
+      {...provided.dragHandleProps}
+      key={`row-${rowIndex}`}
+      className={classNames}
+      role="row"
+      tabIndex="0"
+      style={getItemStyle(
+        provided.draggableProps.style,
+      )}
```
Vladyslav-Velytskyi, line 49: This could be written in one line

lib/SortableList/tests/SortableList-test.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import { describe, beforeEach, it } from '@bigtest/mocha';
```
Vladyslav-Velytskyi, line 2: Please, split values in separate lines

JohnC-80:
@mkuklis please have a look at this and see if your other dnd work could easily switch to using this component. Thanks!

lib/SortableList/DraggableRow.js
```diff
@@ -0,0 +1,65 @@
...
+    return Row;
+  }
+
+  const container = document.getElementById('ModuleContainer');
+  return ReactDOM.createPortal(Row, container);
+};
+
+DraggableRow.propTypes = propTypes;
+
+export default DraggableRow;
```
JohnC-80, line 65: Could this be combined/wrap a call to `defaultRowFormatter` so that mark-up changes could possibly stay in sync?

BogdanDenis:
@JohnC-80 thanks for your detailed review. That's a good point that in `stripes-components` this component should be able to handle different cases. I think I'm going to move this to `stripes-smart-components` for now

lib/SortableList/tests/SortableList-test.js
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  it('should render correct number of rows', () => {
+    expect(sortableList.rows().length).to.equal(2);
+  });
+
+  describe('when lifting a row', () => {
+    beforeEach(async () => {
+      await sortableList.liftRow();
+    });
```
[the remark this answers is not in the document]
BogdanDenis: @JohnC-80 thanks, combining the actions in one method fixed the unmounting problem

lib/SortableList/DraggableRow.js
BogdanDenis: @JohnC-80 good idea! Made these changes and they'll be in PR in `stripes-smart-components`

BogdanDenis:
Decided to move to `stripes-smart-components`
New PR https://github.com/folio-org/stripes-smart-components/pull/840

mkuklis:
Thank you for your work on this @BogdanDenis. It looks like this would still work well with what has been done in ui-requests. I haven't seen any issues described in your TODO but the tests I wrote in ui-requests were also kind of limited:

https://github.com/folio-org/ui-requests/blob/master/test/bigtest/interactors/sortable-list.js#L18-L58

mkuklis:
This feels like it should actually belong to stripes-components. Perhaps it needs to be a bit more polished and as @JohnC-80 mentioned probably renamed.

JohnC-80:
This one will stick in my brain as a thing that's up for migration if I hear of anyone asking for a drag-and-drop MCL.
I think the path we could take in moving it into this library would be to try it out with some of the scenarios that MCL has in storybook. This would be just to figure out caveats and whether it works with some of the other features that MCL has since it is essentially an MCL wrapper and people would want to use it interchangeably with their regular MCL. If there are caveats, we just document them if they aren't complete dealbreakers.

JohnC-80:
Plus, if the component itself doesn't quite work out, I think we might at least document it as a 'pattern' so that people can implement their own to include features that suit their needs... lots of ways to improve things for future developers seeking to meet similar requirements....

PR #1305, opened 2020-06-19
UX-371: Update disabled/read-only styles for RadioButton and Checkbox

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/UX-371

## Changes
- Updated disabled/read-only styles for RadioButton and Checkbox
- Updated RadioButton error/warning styles to match Checkbox error/warning styles
- Removed the lock icon from read-only RadioButton and Checkbox

## Note
I had to disable the `no-descending-specificity`-rule for the CSS for both RadioButton and Checkbox. I tried making it work but it's pretty much impossible to fix the warnings unless you move CSS around in an illogical way that would do more harm than good.

## Before
image

image

## After
image

image

### DIFF
--- a/lib/Checkbox/Checkbox.css
+++ b/lib/Checkbox/Checkbox.css
@@ -11,7 +11,6 @@
   color: var(--color-text);
   line-height: var(--line-height);
   margin-bottom: 0;
-  cursor: pointer;
 
   &.fullWidth {
     width: 100%;
@@ -168,7 +167,9 @@
  */
 .disabled,
 .readOnly {
-  cursor: default;
+  & .inner {
+    cursor: default;
+  }
 
   & .checkboxIcon,
   & input:checked + .checkboxIcon {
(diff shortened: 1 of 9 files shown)


lib/Checkbox/Checkbox.css
```diff
@@ -73,15 +73,74 @@
  */
 .input {
   position: absolute;
-  z-index: 10;
+  z-index: -1;
   opacity: 0;
-  cursor: pointer;
+  pointer-events: none;
```
aliaksei-chumakou, line 78: @rasmuswoelk this line along with `z-index: -1;` makes user not able to select a Checkbox in ui-plugin-find-po-line. Is it the issue how we use Checkbox? https://github.com/folio-org/stripes-acq-components/blob/master/lib/PluginFindRecord/PluginFindRecordModal.js#L193
here is the screen with button _Add_ to call the plugin https://folio-testing.aws.indexdata.com/invoice/view/58a09c81-da98-46a1-bffa-85545b40acb7

rasmuswoelk: Fix merged in 👍

PR #1307, opened 2020-06-23
STCOM-702 Accessibility Error: ARIA role must be appropriate for the element (Selection)

AndreiShumski2:
**Ticket:**
https://issues.folio.org/browse/STCOM-702

**Changes:**
remove `role="listbox"` for `<input>` as no needed
add `role="combobox"`, aria-controls, aria-expanded

### DIFF
--- a/lib/Selection/SelectList.js
+++ b/lib/Selection/SelectList.js
@@ -253,7 +253,7 @@ class SelectList extends React.Component {
                   {ariaLabel => (
                     <input
                       type="text"
-                      role="listbox"
+                      role="textbox"
                       aria-label={ariaLabel}
                       placeholder={placeholder}
                       onChange={this.props.onFilter}
--- a/lib/Selection/SelectList.js
+++ b/lib/Selection/SelectList.js
@@ -254,6 +254,9 @@ class SelectList extends React.Component {
                     <input
                       type="text"
                       aria-label={ariaLabel}
+                      role="combobox"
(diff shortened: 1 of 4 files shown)


lib/Selection/SelectList.js
```diff
@@ -253,7 +253,7 @@ class SelectList extends React.Component {
                   {ariaLabel => (
                     <input
                       type="text"
-                      role="listbox"
+                      role="textbox"
```
aliaksei-chumakou, line 256: do we need any role here? according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role 

> Whenever possible, rather than using this role, use an <input> element with type="text", for single-line input, or a <textarea> element for multi-line input.

lib/Selection/SelectList.js
```diff
@@ -253,7 +253,6 @@ class SelectList extends React.Component {
                   {ariaLabel => (
                     <input
                       type="text"
-                      role="listbox"
                       aria-label={ariaLabel}
```
JohnC-80, line 256: Thank you @AndreiShumski2 !
NVDA acts the same either way...
JAWS, is very different - it doesn't consider the list.

With this in mind, could you try 
```suggestion
                      aria-label={ariaLabel}
                      role="combobox"
                      aria-controls={`sl-container-${this.props.id}`}
                      aria-expanded="true"
```
axe *may still complain about the `<ul>` needing a label...  it can have the `aria-label`

PR #1308, opened 2020-06-23
STCOM-703: Filter headings are not descriptive when using a screenreader

annamelnyk:
# Description
When using a screenreader `Filter headings` was reading as `object object filter list`
# Issue
https://issues.folio.org/browse/STCOM-703
# Screenshots
**Before**
image

**After**
image


JohnC-80:
For testing internationalized components, we have a `mountWithContext` helper function that will wrap the contents in an `<IntlProvider>` component so tests don't have to import/use that component directly... Also, the entire UI of FOLIO is wrapped in an `<IntlProvider>` instance with a dynamically set locale/language within stripes core, so the individual components shouldn't need it themselves...

lib/FilterGroups/FilterGroups.js
```diff
@@ -281,29 +282,31 @@ const FilterGroups = (props) => {
...
-            header={FilterAccordionHeader}
-            disabled={disableNames[group.name]}
-            // If any of the filters start with this group's name, then we should display the 'clear all'
-            // button for this accordion.
-            displayClearButton={Object.keys(filters).some(filter => filter.startsWith(group.name))}
-            onClearFilter={onClearFilter}
-          >
-            <FilterGroup
-              key={`${group.label}-${index}`}
+          <IntlProvider locale="en">
```
JohnC-80, line 285: We shouldn't have to add this here since the whole FOLIO UI is wrapped in one at a high level within stripes-core...

annamelnyk: Thanks @JohnC-80, your review helped me to find why tests were failed.

PR #1314, opened 2020-06-24
MultiSelect: Add `data-open` attr for ease of testing

doytch:
Sure, this looks dumb. 

But @MikeTaylor stumbled onto [this bit of information just now](https://github.com/folio-org/stripes-testing/blob/master/doc/nightmare.md#be-careful-with-multiselection) and I gotta admit, I don't love that [we have to write `:not([hidden])` here](https://github.com/folio-org/ui-inventory/pull/828/files#diff-85c813b2a76a618951639af498e3f0c9R65).

I don't want to break existing functionality obviously, so I'm just adding this attribute to be able to write selectors in a more lexically cogent manner.

### DIFF
--- a/lib/MultiSelection/MultiSelectResponsiveRenderer.js
+++ b/lib/MultiSelection/MultiSelectResponsiveRenderer.js
@@ -11,7 +11,7 @@ const MultiSelectResponsiveRenderer = (props) => {
     return (
       ReactDOM.createPortal(
         (
-          <div className={css.mobileBackdrop} hidden={!props.isOpen}>
+          <div className={css.mobileBackdrop} hidden={!props.isOpen} open={props.isOpen}>
             <div className={css.mobileContainer}>
               <MultiSelectOptionsList {...props} />
             </div>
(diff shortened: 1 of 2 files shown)


zburke:
@doytch, can you explain how this helps given `open` is not a [DOM attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes)?

doytch:
@zburke, that list of attributes contains those have a standardized value format or purpose. You can throw any attribute you want on a DOM element. It can be `foobar-i-love-folio-this-much="a lot"` for all the browser cares.

For this purpose, given the absence of a semantically-appropriate attribute, I think we'd want something that is specifically _not_ on that list to make sure we don't collide with anything else.

VictorSoroka1:
Since it is not in the specification, I would suggest data-open so we can be sure that it won't collide.

zburke:
> I also see warnings from React when DOM element provided with an attribute which is not in the specification.

Yeah, React gets noisy on the console about this kind of thing. `data-open` works for me; good suggestion @VictorSoroka1.

doytch:
Good point. For some reason I missed this warning...

Updated!

MikeTaylor:
Yes, `data-open` is better. (Though that the HTML people were thinking in using such a generic prefix as `data-` for private extensions, I can't imagine. At least RFC 1822's `X-` prefix for non-standard email headers make it clear that they are extensions. I mean, isn't _everything_ data?)

Anyway: go ahead and merge, say I.

PR #1327, opened 2020-07-09
Extend getCellClass and getHeaderCellClass callbacks for more flexibl…

aditya-matukumalli:
…e styling, refs ERM-953, STCOM-718

- Added in `rowData` and `column` name params to `getCellClass` callback for styling specific cells in the MCL.
- Added a similar callback for the headers called `getHeaderCellClass` tthat passes back the column name which lets us add custom classes to specific columns in the MCL


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -1284,18 +1288,25 @@ class MCLRenderer extends React.Component {
...
 
+      const defaultHeaderCellClass = this.getHeaderClassName(header, isSortHeader, isInteractive);
+
       let headerStyle;
+      let headerCellClass;
+
       if (headerWidth) {
         headerStyle = { flex: `0 0 ${headerWidth}px`, width: `${headerWidth}px` };
+        headerCellClass = typeof getHeaderCellClass === 'function' ?
+          getHeaderCellClass(defaultHeaderCellClass, header) : defaultHeaderCellClass;
```
rasmuswoelk, line 1304: Customizing the header cell class should be possible regardless of the header cell having a width or not.

aditya-matukumalli: Reason for this I believe is because MCL needs to get its measuring done first with the basic, functional 
`headerStyle = { flex: '0 0 auto'}`  and then apply other styles?

So 

1. MCL initially renders and applies the default style attributes/classes - merely functional stuff.
2. MCL measures cells to eventually get a columnWidth/headerwidth. It uses the default style/class to accomplish this
3. MCL applies the custom classname as well as the derived headerwidth. 

Since this happens after the measurement is done, the extra styles from `getHeaderCellClass  shouldn't disrupt the functionality of MCL.

JohnC-80: It's tough to dictate the use-case in a web-app that's all-things-but-an-OS running in a browser.... but Ideally, people aren't applying custom classnames that interfere with the functionality... conditionally rendering behind width just shores that up at the cell level so I'm onboard with it. Customizations that may affect width should probably be safely made within `columnMapping` . At the cell level, This is really just to cover purely aesthetic styles - ie background-color.

aditya-matukumalli: I tweaked the behavior such that the app just has the ability to extend the default header class instead of completely overwriting it.

PR #1332, opened 2020-07-16
STCOM-614 leverage currency options to a hook for re-use

aliaksei-chumakou:
we have to leverage logic piece into a reusable hook and export it, so currency options could be easily imported in any module from one place.


VictorSoroka1:
Please update Changelog.

PR #1337, opened 2020-07-22
STCOM-703: Remove dash from formatMessage at FilterAccordionHeader label

annamelnyk:
# Description
Add check for `label` value to remove `-` from formatMessage to prevent console errors.
# In scope of task
https://issues.folio.org/browse/STCOM-703


lib/Accordion/headers/FilterAccordionHeader.js
```diff
@@ -62,11 +62,12 @@ class FilterAccordionHeader extends React.Component {
       contentId,
       headingLevel,
       intl: {
-        formatMessage,
+        formatMessage
       },
     } = this.props;
     const clearButtonVisible = displayClearButton && typeof onClearFilter === 'function';
-    const labelText = formatMessage({ id: `${label?.props?.id}` }) || '-';
+    const labelPropsId = label?.props?.id ?? '';
```
aliaksei-chumakou, line 69: IMHO, there is no too much sense in `?? ''`, since you need actual id here and you check if it's truthy later. There is no particular case where this empty string makes the difference. What do you think of 
```suggestion
    const labelPropsId = label?.props?.id;
```

PR #1339, opened 2020-07-22
Datepicker patch

JohnC-80:
Resolves odd behavior with back-spacing out current input field values.

Approach: 
1) keep state updated when something valid is present
2) only update the formatted value when the incoming value is different.

Additionally, this cleans up the behavior in the storybook story for Datepicker.

Task list
- [x] fixed issue with clear button not appearing
- [x] fixed issue with backspacing
- [x] fixed issue calendar 'cursor' staying in sync with month changes. (a11y)
- [x] issue with focus returning to datepicker when other inputs are selected.
- [x] add usePortal prop to render the calendar into a portal to avoid it being cut off by overflow-hidden containers.


aditya-matukumalli:
@JohnC-80 I am seeing an issue with the `DatePicker` rendered inside a `Card` component. Find the screenshot below
image

This seems to stem from the fact that the `Card` class has got an `overflow: hidden` set on it and that seems to cut off the calendar from displaying.

JohnC-80:
Thanks for reporting @aditya-matukumalli will add to to-do and fix..

PR #1348, opened 2020-08-04
expose EndOfList

aliaksei-chumakou:


NikitaSedyx:
mb it makes sense to export `MultiColumnList` components and export them from lib?

PR #1349, opened 2020-08-04
disable SearchField if loading is true

aliaksei-chumakou:
@JohnC-80 recently I discovered that user can hit search with some filter entered, but search might take time. So user clears the filter, clicks search again, some results are returned, but first one response comes after and rewrites results with the previous search. So maybe it's a good idea to disable fields until previous request is fulfilled, what do you think?


NikitaSedyx:
:+1:

PR #1350, opened 2020-08-04
STCOM-725 Support AutoSuggest field in react-final-form

NikitaSedyx:
https://issues.folio.org/browse/STCOM-725

onStateChange is used for redux-from and onInputChange for final-form

### DIFF
--- a/lib/AutoSuggest/AutoSuggest.js
+++ b/lib/AutoSuggest/AutoSuggest.js
@@ -52,6 +52,7 @@ const propTypes = {
   validationEnabled: PropTypes.bool,
   value: PropTypes.string,
   valueKey: PropTypes.string,
+  withFinalForm: PropTypes.bool,
 };
 
 class AutoSuggest extends React.Component {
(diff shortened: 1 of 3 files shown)


lib/AutoSuggest/readme.md
```diff
@@ -17,4 +17,5 @@ items | List of items to display | | Yes
 onChange | Callback called when the value changes | |
 renderOption | Callback that renders the item in the dropdown | `item => item.value` |
 renderValue | Callback that render the item in the input field | `item => item.value` |
-valueKey | The key in the item object to use as the value. | `"value"`
\ No newline at end of file
+valueKey | The key in the item object to use as the value. | `"value"`
+withFinalForm | toggle form time: tru - final-form, false - redux-from (default) | |
```
AndreiShumski2, line 21: `true`

lib/AutoSuggest/AutoSuggest.js
```diff
@@ -52,6 +52,7 @@ const propTypes = {
   validationEnabled: PropTypes.bool,
   value: PropTypes.string,
   valueKey: PropTypes.string,
+  withFinalForm: PropTypes.string,
```
aliaksei-chumakou, line 55: isn't it bool?

NikitaSedyx: oops, yes

PR #1351, opened 2020-08-04
STCOM-726: Create/Edit User Record: Multi-select custom field does not have a label

BogdanDenis:
## Purpose
- Set correct label for MultiSelection's hidden value input so it no longer violates accessibility guidelines

## Issue
(STCOM-726)[https://issues.folio.org/browse/STCOM-726]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Increase test coverage to 80% | Editor. Refs STCOM-660.
 * Increase test coverage to 80% | FilterPaneSearch. Refs STCOM-689.
 * Datepicker refactor: numerous fixes in accessibility, internationalization, usability. Addresses STCOM-325, STCOM-606, STCOM-684, STCOM-640, STCOM-577, STCOM-315, STCOM-653, STCOM-639, STCOM-470, STCOM-641.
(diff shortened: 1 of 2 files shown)


JohnC-80:
@BogdanDenis This is a more complex issue with the general approach we have with this... once HTML5 Validation happens and the focus is trasferred to the hidden-not-hidden input, the interaction gets a bit stuck: Hopefully you're able to see this: https://folio-project.slack.com/files/U7XR660LD/F017GQRCMRT/multi-select-bug.mov

BogdanDenis:
@JohnC-80 the link you sent doesn't work for me... but I think I understand what the issues is. After HTML validation happens when I try to search an option only one character is displayed in the search field, right?
Users - FOLIO

JohnC-80:
Exactly... so I'm wondering if we *could... put an `onFocus` handler on the 'hidden' field to send focus directly to the filter field... seems like focus is going directly to that anything-except-hidden field when the validation triggers. Give it a shot?

BogdanDenis:
@JohnC-80 sure, I'll look into this issue

BogdanDenis:
@JohnC-80 I've created a separate ticket for this issue and we're going to address it in a sprint or two
https://issues.folio.org/browse/STCOM-728
Do you have any objections if I merge this PR now?

PR #1361, opened 2020-08-17
STCOM-728: Fixed odd behavior on MultiSelect on HTML5 validation

rasmuswoelk:
When the `required`-prop was set to `true` and the HTML5 validation kicked in on form submit, the wrong input field was focused and this caused some odd behavior where only one character was shown at a time when filtering the options.

With this change, the focus will be moved from the hidden input value field and over to the visible filter input which fixes the odd behavior.

## Ticket
https://issues.folio.org/browse/STCOM-728

## Changes
- Moving focus from hidden value input to the visible filter input on validation error
- Moved focus to the control element on mobile because the filter input is hidden (and opened by tapping the MultiSelect) on mobile but the keyboard would still show which caused some even more odd behavior.

## Notes
I tested this using the Mac screen reader and NVDA/JAWS on Windows. While it fixes the issue described in the ticket, the focus shifting also removes the "Please fill in this field"-message that is read out by default by all three screen readers when using native HTML5 validation.

I am not sure if there's a way to preserve that message for screen readers while shifting the focus that quickly.

Also, I must admit that this change feels a bit "hacky" but I don't see any other way of doing it without finding a different way to do HTML5 validation that doesn't include having a hidden value input. But I am open to suggestions!


JohnC-80:
The only circumstance where the value field is focused is if there's a validation error - given that, we could possibly use the SRStatus already used by MultiSelect to relay a similar message.

rasmuswoelk:
> The only circumstance where the value field is focused is if there's a validation error - given that, we could possibly use the SRStatus already used by MultiSelect to relay a similar message.

Good idea! I will give it a shot.

PR #1368, opened 2020-08-24
STCOM-492 : Fixed operator from = to == in order to improve performance

dutr5288:
The "*=*" syntax is translated to *~ regex* search on database side and is *very slow*. Since filters search the exact key, we should use "*==*" syntax that translates to *LIKE* (equivalent to exact in most cases) and that is way faster. For example in location filtered instance search, we see performance improved about *30x*.

**Ticket**

[https://issues.folio.org/browse/STCOM-492](https://issues.folio.org/browse/STCOM-492)

**Changes**

Updated operator from "*=*" to "*==*". Please see the commit for the change.


zburke:
Does this handle array search as noted in #927 (where we rolled back the first implementation of STCOM-492 in #925)? 

> Given the current state of CQL-to-SQL translation in cql2pgsjon-java, detailed in [CQLPG-81](https://issues.folio.org/browse/CQLPG-81), #925 inadvertently broke searching for array fields. Once query translation is updated, we'll consider reinstating this PR. 

I see that CQLPG-81 is closed now. Are we certain that we can filter by string-values (filter items by status, of which they have one) and array-values (filter items by language, of which they have many) when this PR is in place?

zburke:
Note also the performance issue may be substantially mitigated by [STCOM-662](https://issues.folio.org/browse/STCOM-662) since filters that _know_ they want `==` have the ability to specify it. Granted, that's opt-in, and it would be nice if we could make this efficient by default.

dutr5288:
@zburke hmm. this is what i see in the commit of STCOM-662 : 
`          const { operator = '=' } = group;
`
if we use it as the standard way of implementing the operator that i would rather using '*=*' instead of '*==*'.

zburke:
@dutr5288, my point was that the last time we changed the default from `=` to `==` we broke a bunch of things and it isn't clear to me from the description in this PR that the change here is any different than what was done in #925, even though the codebase has evolved since then on both the front-end and back-end. 

Before we merge this change, I would like for there to be some investigation, and documentation, into how things have evolved such that this change will not lead to same set of the problems that required us to rollback #925 in #927. Sound good?

zburke:
As noted in the comments on [STCOM-492](https://issues.folio.org/browse/STCOM-492), "Some anecdotal testing suggests this will work just fine now, but more thorough testing would be prudent." and that testing hasn't happened.

zburke:
The comments above notwithstanding, we are going to move ahead with this change now because delaying it until the next major release could mean waiting six months or more. We know this will break some filters; we'll fix them. The alternative to small number of broken filters that we can easily fix with a small UI patch is a large number of exceptionally inefficient filters that cannot be fixed without backend work and a reindexing of the database.

PR #1371, opened 2020-08-25
STRIPES-693 Textfield clear not working on iOS.

JohnC-80:
It appears that the clear button is disappearing in the focus state transition before the click handler is called. Placing a tiny timeout here can resolve.

This is despite the fact that focus/blur handlers actually are applied to the parent element of the input/clear button...


lib/TextField/TextField.js
```diff
@@ -273,10 +273,15 @@ Use the "inputRef" prop instead to obtain a ref to the input.`);
     const { currentTarget, relatedTarget } = event;
 
     if (!(relatedTarget && currentTarget.contains(relatedTarget))) {
-      this.setState({
-        focused: false
+      // delay focus stay setting for a whole tick. This is intended to keep the clear button around long enough for its
+      // click event to fire on iOS devices.
+      setTimeout(() => {
```
ViktorSoroka07, line 278: `setTimeout` now causes warning in tests like. Can something be done to avoid such?

```
ERROR: 'Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in %s.%s', 'the componentWillUnmount method', '
    in TextField (created by ForwardRef(FormField(TextField)))
    in ForwardRef(FormField(TextField)) (created by ForwardRef(Field))
    in ForwardRef(Field)
    in div (created by Layout)
    in Layout
    in div
    in div
    in Unknown
    in Unknown (created by List)'
```

rasmuswoelk: Maybe store it in a variable and clear the timeout on unmount?

PR #1375, opened 2020-08-27
fix modal backdrop ignoring clicks when closeOnBackgroundClick is passed

MaksymDryha:


rasmuswoelk (approved):
While testing this, I realized that the backdrop couldn't be clicked at all, so I pushed a fix for it.

Can you maybe test it to see if it works as expected?

Approved 👍

aditya-matukumalli (approved):

ryandberger:
Since the CLA is the only thing preventing this from getting merged, I'm going to merge as Admin.

PR #1376, opened 2020-08-27
Upgrade devDep of react-intl to 5.0.0

JohnC-80:


zburke (approved):
Ooof the arg passed to the render-props is a hot mess, but that ain't your fault. This LGTM!

PR #1379, opened 2020-08-31
STCOM-659 provide FormattedDate, FormattedTime

zburke:
ISO-8601 is ambiguous about how timezones in timestamps should be
formatted. In short, the spec says a timezone can be formatted as an
offset like `+HHMM` or `+HH:MM`, but the W3C opted to only require
browsers to implement the latter. Unfortunately, our backend APIs chose
to implement the former.

Some browsers handle both, and for them we are grateful. Others, like
Safari, or Chrome on iOS, do not, and for them we provide
`<FormattedDate>` and `<FormattedTime>` to provide that missing colon.

So much work for two little dots, eh? Sheesh.

Refs [STCOM-659](https://issues.folio.org/browse/STCOM-659)

### DIFF
--- a/lib/FormattedDate/FormattedDate.js
+++ b/lib/FormattedDate/FormattedDate.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { FormattedDate as FT } from 'react-intl';
+import { FormattedDate as FD } from 'react-intl';
 import iso8601Timestamp from '../../util/iso8601Timestamp';
 
 /**
--- a/lib/FormattedDate/FormattedDate.js
+++ b/lib/FormattedDate/FormattedDate.js
@@ -25,7 +25,7 @@ import iso8601Timestamp from '../../util/iso8601Timestamp';
 const FormattedDate = props => {
   const { value, ...rest } = props;
   const tweakedValue = iso8601Timestamp(value);
-  return <FT value={tweakedValue} {...rest} />;
+  return <FD value={tweakedValue} {...rest} />;
 };
 
 FormattedDate.propTypes = {
--- a/lib/FormattedDate/FormattedDate.js
+++ b/lib/FormattedDate/FormattedDate.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { FormattedDate as FT } from 'react-intl';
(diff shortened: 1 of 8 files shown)


lib/FormattedDate/FormattedDate.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedDate as FT } from 'react-intl';
```
JohnC-80, line 3:
```suggestion
import { FormattedDate as FD } from 'react-intl';
```

lib/FormattedDate/FormattedDate.js
```diff
@@ -0,0 +1,35 @@
...
+ * It turns out that Chrome will usually handle timestamp strings without the
+ * colon, but not on iOS. Ditto for Firefox. Safari always chokes on them.
+ *
+ * @see https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC
+ * @see https://www.w3.org/TR/NOTE-datetime
+ */
+const FormattedDate = props => {
+  const { value, ...rest } = props;
+  const tweakedValue = iso8601Timestamp(value);
+  return <FT value={tweakedValue} {...rest} />;
```
JohnC-80, line 28:
```suggestion
  return <FD value={tweakedValue} {...rest} />;
```

zburke:
In d5abd0f, swapping out `-` for `/` makes sense because it's probably just creating a string to pass to `new Date()`, i.e. you're at the mercy of the [browser's `Date.parse(:string:)` impl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse). Many browsers choose to accept `m/dd/yyyy` even though it's not part of the ISO-8601 standard but because it's the US convention and, as we all know, the universe revolves around the US. But of course there is some variation in exactly what formats different browsers choose to handle, which is exactly the issue here (except WRT how timezone offsets are expressed, rather than how date-part separators are expressed). 

IOW, I don't think `-` vs `/` is relevant here, unless you think we should get into more robust prop checking in the util function. Given this is mainly expected to handle date strings retrieved from the backend which we expect to be consistently formatted, I was aiming for speed (simple parsing based on length and the position of `+`) over robustness (some big regex). What do you think?

PR #1387, opened 2020-09-03
Datepicker: delay/cancel blur events

JohnC-80:
A few have reported issues with Redux-form/ RFF validation on Datepicker. The new behavior of Datepicker moves focus to the calendar widget when it is summoned. This can cause R-F's `onBlur` handler to trigger unwantedly, causing premature validations and potentially messing up the state of Redux-form.
This PR saves an onBlur trigger for once focus leaves the entire control (setTimeout instanciated on blur, canceled if focus happens within the component)

### DIFF
--- a/lib/Datepicker/Calendar.js
+++ b/lib/Datepicker/Calendar.js
@@ -85,6 +85,7 @@ const propTypes = {
   intl: PropTypes.object,
   locale: PropTypes.string,
   onBlur: PropTypes.func,
+  onFocus: PropTypes.func,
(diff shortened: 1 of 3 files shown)


aliaksei-chumakou (approved):
To me it's hard to review this solution and to say what could go wrong, because the whole component looks pretty complicated and trying to fulfill 2 form engines (redux and finalForm)
So I approve to not hold on potential issues for datepickers with redux-form across FOLIO

PR #1388, opened 2020-09-07
Fix SingleSelection interactor

BogdanDenis:
## Purpose
`<SingleSelection>`  interactor throws errors when used in tests outside of `stripes-components`. This is caused by options list being rendered outside of component's markup through a portal, which causes app tests to fail when selecting `li:nth-of-child` elements

## Issue
[UIEH-908](https://issues.folio.org/browse/UIEH-908)


zburke (approved):
Thanks; this looks more robust now. 

I've had the same kind of trouble in the end-to-end tests with components that render differently, or to different parts of the DOM, at different resolutions. That is a tricky problem to track down until you realize what you're looking for. Nice work!

PR #1392, opened 2020-09-10
STCOM-753 log errors caught by ErrorBoundary

zburke:
Log errors in `<ErrorBoundary>` when handlers for Rollbar, Sentry, or
BugSnag are configured via stripes-config in `stripes.config.js` in the
`errorLogging` property.

Refs [STCOM-753](https://issues.folio.org/browse/STCOM-753), [STCOR-455](https://issues.folio.org/browse/STCOR-455)


zburke:
This is inarguably unambiguously the wrong approach to handling different logging providers and will be closed with extreme prejudice. As discussed at #stripes-architecture, this should be handled via a `stripes.actsAs` type, maybe a new type such as `error-logger`, or maybe an existing type such as `handler` and a new event-type.

PR #1395, opened 2020-09-16
STCOM-709 refactor SingleSelect away from componentWillReceiveProps

NikitaSedyx:
purpose: https://issues.folio.org/browse/STCOM-709 componentWillReceiveProps is deprecated

approach: getDerivedStateFromProps is used
approach is not ideal because component itself should be reimplemented, memoization is more suitable approach here, but can't be used


aliaksei-chumakou (approved):

JohnC-80 (approved):
Looks good! Thanks @NikitaSedyx !

mkuklis (approved):

PR #1398, opened 2020-09-17
Continued datepicker fixes.

JohnC-80:
## Problem
Datepicker calendar displays the incorrect month when viewed in an EU timezone

The date string used to construct the months list didn't had a misformatted date string. I simple adjusted the formatting to include all 2 digits of the necessary month and raised the date to a number out of the month range.

Additionally, updated the weekday list (locked down the timezone) so the week days - "Sun, Mon, Tues etc" would be consistent.

### Other included work -
-  Expanding the formats considered in the default parser. Some apps use `moment.now.format()` to generate default values - this creates a string that's a slightly different format from `moment.now.toISOString()`.

## Before
image

## After
image


zburke:
> Expanding the formats considered in the default parser. Some apps use `moment.now.format()` to generate default values - this creates a string that's a slightly different format from `moment.now.toISOString()`.

Do we just need clarity here about the acceptable formats in our documentation -- "thou shalt not use ...", or even our own utility function, implemented in terms of our preferred `moment` function, so that when the time comes to [refactor away from `moment`](https://issues.folio.org/browse/STRIPES-699) it's easier?

JohnC-80:
@zburke At the moment (har) I'm apt not to tell anyone how they should/shouldn't use moment, since that would fall into the things we remove when we do refactor away from moment....
plus, refactoring away from moment shouldn't be incredibly difficult - the line of exchange between Datepicker and apps is a string value... either end could refactor away from moment and neither would care.

PR #1403, opened 2020-09-22
STCOM-758 add nonInteractive prop to TextField

aliaksei-chumakou:
https://issues.folio.org/browse/STCOM-758
Guys, here is the another requirement from the ticket - to display text field on the form in non-interactive mode as KeyValue component.

### DIFF
--- a/lib/TextField/TextField.js
+++ b/lib/TextField/TextField.js
@@ -4,6 +4,7 @@ import { FormattedMessage } from 'react-intl';
 import uniqueId from 'lodash/uniqueId';
 import classNames from 'classnames';
 
+import KeyValue from '../KeyValue';
 import formField from '../FormField';
 import Label from '../Label';
 import nativeChangeField from '../../util/nativeChangeFieldValue';
@@ -92,6 +93,9 @@ class TextField extends Component {
      * String of preset styles to apply to textfield. possible values: noBorder
      */
     inputStyle: PropTypes.string,
+
+    isNonInteractive: PropTypes.bool,
+
     /**
      * The content for the input's label.
      */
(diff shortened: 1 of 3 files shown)


ViktorSoroka07 (approved):

AndreiShumski2 (approved):

uladislausamets (approved):

JohnC-80:
Why do this here rather via conditional rendering at the App level? If we do this for TextField, then for consistency's sake it would need to be done for all of our individual form controls....

It might be better to create a wrapping component that with either render the KeyValue *or* the Component/Control that's necessary.
The API would look something like: (Just using EditSwitch as temporary name)

```
<EditSwitch interactive component={ TextField } />
```

zburke:
That's a nice API proposal, @JohnC-80, and it encapsulates all the logic in a single component instead of spreading it across every form component.

aliaksei-chumakou:
it might be a good suggestion, but in fact we need this to be implemented exactly for every form control, because non-interactive view is different. For example, Datepicker - it's not a standard value of datefield. The same for Selection. And so on.

lib/TextField/TextField.js
```diff
@@ -92,6 +93,9 @@ class TextField extends Component {
      * String of preset styles to apply to textfield. possible values: noBorder
      */
     inputStyle: PropTypes.string,
+
+    isNonInteractive: PropTypes.bool,
```
mariia-aloshyna, line 97: Consider adding a description for this prop for consistency, as it's done above for other props.

mariia-aloshyna (approved):

aliaksei-chumakou:
@JohnC-80  here is the context of requirement https://ux.folio.org/docs/guidelines/ux-patterns/hiding-vs-disabling-elements-ux-pattern/

PR #1408, opened 2020-09-29
STCOM-714: Expose the ability to remove margins on RepeatableFields' legend

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/STCOM-714

## Changes
- Added `className`-prop that allows for adding a custom class name for the root element
- Added `hasMargin`-prop which defaults to `true`.

## Notes
Ideally, the component should not have any vertical margin by default so that it automatically aligns with other sibling components in e.g. a form, but since most implementations out there currently expect the margin to be present, I set it to true so we don't "break" any layouts.

Personally, I would recommend to set the default to false in the future or even removing the option entirely as it should be handled by parent components or via. custom styling.

For now, developers can set `hasMargin` to `false` if they want to disable the default margin.


MikeTaylor (changes requested):
Looks good!

MikeTaylor (approved):

JohnC-80 (approved):

MikeTaylor:
Hey, @rasmuswoelk  — just to let you know that this does solve my problem in https://openlibraryenvironment.atlassian.net/browse/PR-723

Many thanks.

PR #1417, opened 2020-10-13
STCOM-743: Refactor <Popover> to use <Popper>

rasmuswoelk:
## Ticket
http://issues.folio.org/browse/STCOM-743

## Changes
- Moved the current `<Popover>`-component into another component: `<LegacyPopover>` that will be rendered if the deprecated component API is used (array of children). This ensures backward compatibility.
- Added deprecation warnings for deprecated props
- Updated the `<Popover>` to use the `<Popper>` for positioning the overlay
- Updated examples to use the new component API
- Updated the readme `<Popover>`-readme and added tests
- Updated the `<InfoPopover>` to use the new `<Popover>` component API.
- Added a `renderTrigger`-prop for the  `<InfoPopover>`
- Updated the `<InfoPopover>`-readme and added tests


JohnC-80 (approved):
The only nitty thing is the "lagacy" filename.
Is there at least one test case where the legacy code runs? If not, I suppose we could just see about istanbul-ignore-ing it since it is slated for removal.

rasmuswoelk:
@JohnC-80 You are absolutely right.. I did try to add a test for the legacy popover but it didn't render for some reason. I didn't want to spend more time on it since it's soon to be deprecated anyway. However, I went back in and I found out what caused this issue in the tests.

I removed the code that returned null if the `this.popContainer` is null in the render method of the legacy popover and that fixed the issue. The legacy popover still works in Storybook. And I added a simple test to check if it renders when the legacy component API is used. You can see the change in my latest commit. Let me know if you think this change could cause any issues for existing implementations.

ViktorSoroka07:
After this PR merged tests started to fail in stripes-data-transfer-components. Please address this.
https://github.com/folio-org/stripes-data-transfer-components/pull/101
https://jenkins-aws.indexdata.com/job/folio-org/job/stripes-data-transfer-components/job/UIDATIMP-574/1/display/redirect

rasmuswoelk:
@ViktorSoroka07 The legacy popover should still work. I will check up on it.

ViktorSoroka07:
@rasmuswoelk, I think this is local issue to my module as we depended on the implementation details in the interaction. My team will take care of this. Thank you for looking at it.

rasmuswoelk:
Ah. Actually, the legacy popover works but the `<InfoPopover>` use the new component API. I think it's easily fixable by changing the selector in the test interactor from `[data-role="popover"]` to `[data-test-info-popover]`.

PR #1420, opened 2020-10-15
STCOM-671 - Update state of AccordionSet and AccordionStatus when unmounting chil…

JohnC-80:
…d accordions

This PR keeps states in sync with mounted Accordion children by removing the entries of the children when they dismount.


mariia-aloshyna (approved):
I tested this in ui-data-import and everything works perfectly, thank you!

BogdanDenis (approved):
Tested it with Custom Fields, looks good!

PR #1422, opened 2020-10-23
STCOM-756 Accordion css implementation

JohnC-80:
Behavior of accordions in inventory prompted me to re-examine the way that Accordions are implemented - there had to be a better way. 
This uses CSS animation and an inner div to wrap the content. TIL: the visibility property can be "animated" - it just steps, so it's really more in applying the delay.
Also, it's functional....
invent


rasmuswoelk:
I'll take a look at this soon!

zburke (approved):
Functionally, this looks great to me. Nice work!

Are there DOM changes that we should know about if folks have written custom interactors in their own repos?

rasmuswoelk (changes requested):
Nice work overall! It's really a long overdue update for the accordion so I'm looking forward to getting these changes merged in!

The only thing I noticed is the collapse/expand transitions being a bit "janky". It looks like it's due to the margin-bottom being removed/applied on collapse/expand. It kinda jumps a bit just before the transition happens:



I played around with a bit and I could get it to stop jumping by adding that spacing inside the content div. It could be solved by either adding an additional "inner"-element that has that extra margin or in the case below, I added an :after pseudo element to .content that has a height of 1rem. I also tweaked the transition a bit:

```css
.content {
  opacity: 0;
  visibility: hidden; /* inputs within will not be focusable */
  transition: transform 0.1s ease 0.5s, opacity 0.1s linear 0.5s, visibility 0.2s linear 0.5s;

  &.expanded {
    opacity: 1;
    visibility: visible;
    transform: translateY(0);
    transition: transform 0.15s ease-in-out 0.2s, opacity 0.15s ease-in-out;
  }

  &:after {
    content: '';
    height: 1rem;
    display: block;
    width: 100%;
  }
}
```

Updated transition:

JohnC-80:
@rasmuswoelk aw, you don't like the fade-in of content? Just kidding.. it's gone. Thanks for the adjustments!

rasmuswoelk (approved):

lib/Accordion/Accordion.css
```diff
@@ -67,20 +67,43 @@
...
  */
-.content {
-  max-height: 0; /* hidden */
+
+.content-wrap {
+  max-height: 0;
   overflow: hidden;
-  transition: max-height 0.25s ease;
-  width: 100%;
+  z-index: 1;
```
BogdanDenis, line 76: Hi @JohnC-80, we've found one issue with updated Accordion in eHoldings application - part of content is cut off and disappears under an accordion that is below
image

I was able to partly fix is by applying these styles:
```
.content-wrap.expanded {
  overflow: visible;
}
```
This fixes the cutoff issues, but still something wrong with z-indexes
image

Adding
```
.content-wrap {
  z-index: 2;
}
```
This makes z-index equal to that of `<Headline>` element, which is defined in `interactionStyles.css:26`
image

Now I'm stuck at this point - both `Tags` and `Holding status` are accordions which have `z-index: 2`, `multiSelectMenu` has `z-index: 9999`.
Could you help me with this please?

BogdanDenis: Also, we have a component that displays a list of items which is a "infinite scroll" type of list
https://folio-snapshot.dev.folio.org/eholdings/packages/19-6581?searchType=packages&q=ebsco
Here we should be able to scroll through all of the items, but the scroll isn't working now
You can see how it should work here
https://folio-goldenrod.dev.folio.org/eholdings/packages/19-6581?searchType=packages&q=ebsco

lib/Accordion/Accordion.js
```diff
@@ -35,257 +35,124 @@ const propTypes = {
...
+    updateOpen(current => !current);
+  }).current;
 
-  setContentRef = (ref) => {
-    this.content = ref;
-    if (this.props.contentRef) {
-      this.props.contentRef(ref);
-    }
-  }
+  const onToggle = typeof openProp === 'undefined' ? uncontrolledToggle : onToggleProp;
```
BogdanDenis, line 95: `openProp` here seems to be an undeclared variable

yelyzavetakhokhlova:
Hey @JohnC-80, we also have issues with content being cut off in data export accordions. ☹️
image

PR #1428, opened 2020-11-04
STCOM-771: Set modalRoot position to absolute to allow for modals to stack

rasmuswoelk:
## Issue
https://issues.folio.org/browse/STCOM-771

## Changes
- Set .modalRoot position to absolute to allow for modals to stack.

## Notes
This change will allow for multiple modals to appear at the same time. Before this change, only one modal would be visible at a time, so when the second modal opened, the first modal would disappear and then reappear once the second modal was closed. 

In general, we should avoid having multiple modals open at once but there are edge cases – such as when the "You have unsaved changes"-modal opens while another modal is visible, where the new behaviour is better than the previous because it's less confusing for the user.

image


yelyzavetakhokhlova:
Please, update the changelog

yelyzavetakhokhlova (approved):

PR #1433, opened 2020-11-09
STCOM-775: Provide useDateFormatter and useTimeFormatter hooks

yelyzavetakhokhlova:
[[STCOM-775]](https://issues.folio.org/browse/STCOM-775) Provide `useDateFormatter` and `useTimeFormatter` hooks

## Purpose
Currently, there are `FormattedDate` and `FormattedTime` components in stripes-components which allow handling dates with different formats, i.e. 2020-03-24T17:59:57.369+0000 and 2020-03-24T17:59:57.369+00:00
However, there are use cases when such info should be provided as a string type (e.g. aria-labels). For this `useDateFormatter` and `useTimeFormatter` were created.


ViktorSoroka07 (approved):

hooks/tests/useDateFormatter-test.js
```diff
@@ -0,0 +1,16 @@
+import {
```
aliaksei-chumakou, line 1: IMO, the point of unit tests to put them along with the tested code. Should we discuss this? cc @ViktorSoroka07

ViktorSoroka07: Do you mean removing `tests` folder?

aliaksei-chumakou: in case it's jest tests - yes. This dir came from bigtest, where we put tests under `tests` dir. With jest/rtl we use another pattern. 
If it's bigtests - then it's fine to keep as is

ViktorSoroka07: Yeah, it is Bigtest. So I suggest leaving as per the established approach in the project. Regarding jest, makes sense. I will think about this approach in data-export project. Currently, we use the `test` directory there as well.

aliaksei-chumakou (approved):

JohnC-80 (approved):
I like these alot! Thank you!

yelyzavetakhokhlova:
> I like these alot! Thank you!
Thanks, John!

zburke (approved):
Ooooh, good thinking! 

To provide better parity with the existing components and the underlying [formatjs.io](https://formatjs.io/docs/react-intl/api#formatdate) functions, what if we name the hooks `useFormatDate()` and `useFormatTime()`?

rasmuswoelk (approved):

yelyzavetakhokhlova:
> Ooooh, good thinking!
> 
> To provide better parity with the existing components and the underlying [formatjs.io](https://formatjs.io/docs/react-intl/api#formatdate) functions, what if we name the hooks `useFormatDate()` and `useFormatTime()`?

Done.

PR #1438, opened 2020-11-11
STCOM-778 Dropdown only uses focusHandler.open if trigger is focused

JohnC-80:
### Before
Dropdown would just look at its open prop and feel like it would need to move focus when state changes occurred outside...
DD-before
### After
Dropdown only moves focus to the menu if the trigger is clicked/focused. The focus changes here were done with the keyboard, so that still works...
DD-after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Implement info callout. Refs STCOM-776.
 * Set modalRoot position to absolute to allow for modals to stack. Refs STCOM-771.
 * Provide `useDateFormatter` and `useTimeFormatter` hooks. Refs STCOM-775.
+* Fix Dropdown moving focus when external state changes. Fixes STCOM-778.
 
 ## [8.0.0](https://github.com/folio-org/stripes-components/tree/v8.0.0) (2020-10-05)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v7.0.1...v8.0.0)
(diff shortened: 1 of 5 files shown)


zburke (approved):
YES PLEASE! Nice work!

zburke:
Hmmm, it appears this change broke three focus-related tests in stripes-core ([#1](/folio-org/stripes-core/blob/ad9f2cc9f5dd7b15ce624e948d6c5edfb9f9c109/src/components/MainNav/AppList/tests/AppList-test.js#L74-L76), [#2](/folio-org/stripes-core/blob/ad9f2cc9f5dd7b15ce624e948d6c5edfb9f9c109/src/components/MainNav/AppList/tests/AppList-test.js#L122-L124), [#3](/folio-org/stripes-core/blob/ad9f2cc9f5dd7b15ce624e948d6c5edfb9f9c109/src/components/MainNav/AppList/tests/AppList-test.js#L131-L133)). 

When I debug the tests manually in the browser with `describe.only` on the block in question, the expectations are met, e.g. `document.activeElement.id` _does match_ the id of the first menu, so I'm not sure why the test fails, or how to change it to succeed. Any thoughts, @JohnC-80? This is preventing a few PRs @mkuklis and I have been working on. Mine are just cleaning up tests and some console warnings, but https://github.com/folio-org/stripes-core/pull/946 is a legit bug.

PR #1442, opened 2020-11-17
UX-397: Add design tab in Storybook

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/UX-397

## Changes
- Added `storybook-addon-designs`-package and set up a Figma design for the `<Button>` component for testing purposes.

## Try it out
https://ux.folio.org/storybook/?path=/story/button--basic-usage

## Screenshot
image


JohnC-80 (approved):
Kinda curious about anything not-so-obvious as to what we might actually do with this.. Just a way to keep design docs in-sync/consolidated?

rasmuswoelk:
@JohnC-80 I think Filip wants to experiment with providing additional design documentation and probably also UX guidelines for components – starting with the Button. 

Similar to what ux.folio.org/docs is for but it's not getting updated too often. The idea is to make it easier to maintain I guess :)

PR #1444, opened 2020-11-25
PaneHeader performance tweak

rasmuswoelk:
This PR features a few performance updates for the PaneHeader. I noticed that some improvements could be made here while I was working on the columns selector ticket for ui-users.

## Changes
- Wrapped `hasActionMenu` in `useMemo` to prevent it from running on each re-render.
- Wrapped PaneHeader in `memo` to hopefully save a few re-renders there as well


ryandberger (approved):
Nice find! Looks good!

PR #1449, opened 2020-12-04
Add centralized hotkeys, Storybook story for Commander

JohnC-80:
As we move closer to hotkeys implemetations across FOLIO, it makes some sense to centralized the default key handlers for the system. This way, module developers won't have to write them all from scratch and can augment them if needed...

Shared keys are document in this section: https://github.com/folio-org/stripes-components/blob/centralize-default-hotkeys/lib/Commander/README.md#step-1-base-configuration

Improvements to HasCommand include support for functional components and updating handlers that need to access properties of state.


lib/Commander/README.md
```diff
@@ -54,6 +54,12 @@ const commands = [
 
 export default commands;
 ```
+
+A centralized set of FOLIO-wide shortcut defaults is availabl. Simply import... it contains basic key mapping for commands named **save, new, edit, search, expandAllSections, collapseAllSections, and duplicateRecord"**
```
aditya-matukumalli, line 58:
```suggestion
A centralized set of FOLIO-wide shortcut defaults is available. Simply import... it contains basic key mapping for commands named **save, new, edit, search, expandAllSections, collapseAllSections, and duplicateRecord"**
```

aditya-matukumalli (approved):
LGTM!

lib/Commander/README.md
JohnC-80: 👍 thanks! I updated on another commit. Habits from a system that would invalidate someone's review if the suggestion was commited >.<

PR #1452, opened 2020-12-09
STCOM-785: interactors update - checkbox

jbolda:
This upgrades the `Checkbox` to work with the new interactors. This requires the new interactor as PRed in https://github.com/folio-org/stripes-testing/pull/78.

/cc @JohnC-80

### DIFF
--- a/lib/Checkbox/tests/Checkbox-test.js
+++ b/lib/Checkbox/tests/Checkbox-test.js
@@ -3,15 +3,15 @@ import { describe, beforeEach, it } from 'mocha';
 import { expect } from 'chai';
 
 import { Field } from 'redux-form';
+import { Checkbox as Interactor } from '@folio/stripes-testing';
 import { mount, mountWithContext } from '../../../tests/helpers';
 import TestForm from '../../../tests/TestForm';
 import Harness from '../../../tests/Harness';
 
 import Checkbox from '../Checkbox';
-import { Checkbox as CheckboxInteractor } from '@folio/stripes-testing';
 
 describe('Checkbox', () => {
-  const checkbox = CheckboxInteractor('My label');
+  const checkboxWithLabel = Interactor('My label');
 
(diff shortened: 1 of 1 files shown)


lib/Checkbox/tests/Checkbox-test.js
```diff
@@ -8,10 +8,10 @@ import TestForm from '../../../tests/TestForm';
 import Harness from '../../../tests/Harness';
 
 import Checkbox from '../Checkbox';
-import CheckboxInteractor from './interactor';
+import { Checkbox as CheckboxInteractor } from '@folio/stripes-testing';
```
cowboyd, line 11: The convention we're rolling with is to just call it `Interactor`

lib/Checkbox/tests/Checkbox-test.js
```diff
@@ -27,72 +27,61 @@ describe('Checkbox', () => {
...
         />
       );
-      await checkbox.focusInput();
+      await checkbox.focus();
     });
 
-    it('has an id on the input element', () => {
-      expect(checkbox.id).to.equal('checkbox-test');
-    });
+    it('has an id on the input element', async () =>
```
cowboyd, line 33: Since an `async function` is just a function that returns a promise, most of the assertions, like this one, can be converted into a 1-liner:

```js
it('has an id on the input element', () => checkbox.has({ id: 'checbox-test' }));
```

cowboyd (approved):
Looks good, maybe convert to 1-liners where appropriate

JohnC-80 (approved):

lib/Checkbox/tests/Checkbox-test.js
jbolda: Ah, I thought someone mentioned we needed async/await, I didn't even try it without. Makes sense to me.

PR #1453, opened 2020-12-09
STCOM-785: interactors update - radio button

jbolda:
This upgrades the `RadioButton` to work with the new interactors. This requires the new interactor as PRed in https://github.com/folio-org/stripes-testing/pull/79.

/cc @JohnC-80


JohnC-80 (approved):

jbolda:
Looks like I missed the redux form test. I will finish that up on Monday.

PR #1457, opened 2020-12-11
STCOM-790: Fix inconsistent label font sizes

rasmuswoelk:
Added `--font-size-form-label` CSS variable and used it to replace the hardcoded and generic variables in form components. This ensures that the label font-size is consistent across the form components.


Dmitriy-Litvinenko (approved):

Dmitriy-Litvinenko:
@rasmuswoelk I approve current pull request, but we also need get approval from people that responsible for current repository

PR #1458, opened 2020-12-14
STCOM-756 Accordion css implementation

JohnC-80:
Fixes on #1422


BogdanDenis:
Thanks @JohnC-80, looks much better now!
Only problem that I see is left is in eHoldings Titles list - the majority of the list should be hidden until you scroll to the bottom of the page. Then an additional scroll appears.
This particular functionality has caused a lot of pain to support and eventually we want to come up with a different way t oshow this data. But until then we have to support this, so...

I made a few changes locally to fix scrolling issues:
```
.content-wrap.expanded {
  min-height: 0; // for some reason overflow: hidden does not actually hide content unless you apply this :/
}

.content {
  // other styles
  margin: 0 0 1em 0;

  /* &::after { ... } */ //removed this to fix issue with wobbly scroll (gif attached). Any particular reason to use after over a margin?
}
```

eHoldings - EBSCO Chemical Engineering Collection_ India - FOLIO

You can't see it on the gif, but I removed `content::after` in dev tools and added a bottom margin and this fixed the issue.

What do you think about these changes?

BogdanDenis (approved):

zburke (approved):
LGTM

PR #1467, opened 2020-12-21
STCOM-783 After the browser window is minimized the pane is not adjusted so the content on the righthand side is cut

uladislausamets:
## Purpose

https://issues.folio.org/browse/STCOM-783


lib/Paneset/PaneResizeContainer.js
```diff
@@ -104,21 +98,21 @@ const PaneResizeContainer = ({ isRoot, children, onElementResize, parentElement
         setActiveHandle(null);
       }
     }
-  }, [dragging, activeHandle, onElementResize, cursorX, isRoot]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [dragging, activeHandle, onElementResize, cursorX, isRoot, containerRect.current]);
```
aliaksei-chumakou, line 102: could you please check this issue, it looks like depending on refs might cause troubles https://github.com/facebook/react/issues/14387

ViktorSoroka07:
I checked it from the branch and works perfectly. Thanks @uladislausamets. I do share concern from @aliaksei-chumakou  though. Aside from that, looks great.

lib/Paneset/PaneResizeContainer.js
uladislausamets: @aliaksei-chumakou @ViktorSoroka07 I removed ref dependency and it works the same now

aliaksei-chumakou (approved):

ViktorSoroka07 (approved):

JohnC-80 (approved):
I like the approach of using hooks/memoization vs debouncing. Looks good!

PR #1468, opened 2020-12-22
STCOM-785: interactors update: text-field

jgnieuwhof:
## Motivation
`TextField` is one of the critical path interactors listed in https://issues.folio.org/browse/STCOM-785.

Depends on: https://github.com/folio-org/stripes-testing/pull/90

## Approach


jbolda:
@JohnC-80 I think this one should be all set. It seems we can rerun the tests, and hopefully it is good to go.

JohnC-80 (approved):

PR #1470, opened 2020-12-23
STCOM-785: interactors update: textarea

jgnieuwhof:
## Motivation
`TextArea` is one of the critical path interactors listed in https://issues.folio.org/browse/STCOM-785.

Depends on: https://github.com/folio-org/stripes-testing/pull/91

## Approach


jbolda:
@JohnC-80 This one should be set as well.

JohnC-80 (approved):

PR #1481, opened 2021-01-05
STCOM-785: interactors update: MultiColumnList

jgnieuwhof:
## Motivation
`MultiColumnList` is one of the critical path interactors listed in https://issues.folio.org/browse/STCOM-785.

Depends on: https://github.com/folio-org/stripes-testing/pull/95

## Approach

### DIFF
--- a/lib/MultiColumnList/tests/MultiColumnList-test.js
+++ b/lib/MultiColumnList/tests/MultiColumnList-test.js
@@ -36,11 +36,6 @@ const data5 = generate(5);
 const data100 = generate(100);
 const bigWidthData = generate(30, generator);
 
-const log = x => {
-  console.log(x);
-  return x;
-};
-
 const columnCount = (dataSet) => Object.keys(dataSet[0]).length;
 const rowCount = (dataSet) => dataSet.length;
 const value = (dataSet, [x, y]) => Object.values(dataSet[x])[y];
@@ -65,7 +60,7 @@ const cumulativeCellWidth = async (mcl) => {
 
 let headerClicked = false;
 
-describe.only('MultiColumnList', () => {
+describe('MultiColumnList', () => {
   const mcl = Interactor();
 
   describe('column width ranges', () => {
--- a/lib/MultiColumnList/tests/MultiColumnList-test.js
+++ b/lib/MultiColumnList/tests/MultiColumnList-test.js
@@ -1,18 +1,19 @@
 import React from 'react';
 import { StaticRouter } from 'react-router';
-import { describe, beforeEach, before, it } from '@bigtest/mocha';
-import { Interactor } from '@bigtest/interactor';
+import { describe, beforeEach, before, it } from 'mocha';
 import sinon from 'sinon';
 import { expect } from 'chai';
 import faker from 'faker';
+import {
+  MultiColumnList as Interactor,
+  MultiColumnListCell as CellInteractor
+} from '@folio/stripes-testing';
 
 import { mountWithContext } from '../../../tests/helpers';
-
 import MultiColumnList from '../MultiColumnList';
 import { dataChangedOrLess } from '../MCLRenderer';
 import MCLHarness from './MCLHarness';
 import DimensionCache from '../DimensionCache';
-import MultiColumnListInteractor from './interactor';
 import * as handlers from '../defaultHandlers';
 import convertToPixels from '../convertToPixels';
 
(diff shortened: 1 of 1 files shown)


lib/MultiColumnList/tests/MultiColumnList-test.js
```diff
@@ -29,14 +30,41 @@ const generator = () => (
...
+  await mcl.perform(el => { nColumns = el.querySelectorAll(['[class*=mclHeader-]']).length; });
+
+  const widths = await Promise.all([...Array(nColumns)].map(async (_, i) => {
+    let width = 0;
+    await mcl.find(CellInteractor({ xy: [0, i] })).perform(el => { width = el.offsetWidth; });
+    return width;
+  }));
+
+  return widths.reduce((sum, width) => sum + width, 0);
+};
```
jgnieuwhof, line 62: @cowboyd / @jbolda - this function seems to highlight a couple of points for discussion.

1. It'd be nice to have a `perform`-ish function that returns a value. The "declaring a mutable variable to set it in a promise" pattern is a bit unfortunate.

2. The `cumulativeCellWidth` function should be a filter on the interactor. The reason it's here is because it clutters the `MultiColumnList` interactor, which is intended for application testing, rather than component library testing. This has come up before, using `perform` as an escape hatch when we don't want to clutter the interactor with another filter. This might point to the need to have two sets of interactors - one for testing applications that consume the component library, and one for testing the component library itself. The component library interactors would be a superset of the application interactors. Thoughts?

lib/MultiColumnList/tests/MultiColumnList-test.js
```diff
@@ -194,80 +200,77 @@ describe('MultiColumnList', () => {
...
+
+        it('shifts the data down another row', async () => {
+          await mcl.find(CellInteractor({ xy: [2, 0] })).has({ content: data5[0].name });
+        });
       });
     });
   });
 
-  describe('interaction with a non-virtualized list', () => {
+  describe.only('interaction with a non-virtualized list', () => {
```
jgnieuwhof, line 267: @cowboyd / @jbolda - this is where I left off, do we have a pattern with the new interactors for handling scrolling?

Should it be an action on the interactor? Or some global mouse / scroll interactor like `keyboard`?

cowboyd: Your intuition feels right, but let's de-risk by keeping it scoped to this one interactor for now.

lib/MultiColumnList/tests/MultiColumnList-test.js
```diff
@@ -194,80 +200,77 @@ describe('MultiColumnList', () => {
...
+        await mcl.find(CellInteractor({ xy: [1, 0] })).has({ content: data5[0].name });
       });
 
-      it('renders data in appropriate cell', () => {
-        expect(mcl.rows(1).cells(0).text).to.equal(dataName);
+      describe('click a header again', () => {
+        beforeEach(() => mcl.clickHeader('name'));
+
+        it('shifts the data down another row', async () => {
+          await mcl.find(CellInteractor({ xy: [2, 0] })).has({ content: data5[0].name });
```
cowboyd, line 261: rather than specifying everything by, index would this be better as combination of 

```js
await Cell({ column: 'name', row: 0}).has({ content: data5[0].name });
```

Also, I'm thinking that though it might be tedious, but using a fixture here, or at least overriding a specific instance of the generated data would make the test case so much clearer:

```js
await Cell({ column: 'name', row: 0}).has({ content: 'Bob Dobalina' });

jgnieuwhof: Great call on using the column name rather than index, that's a big legibility improvement.

Re: the data, I also have a strong preference away from generated data in tests (unless explicitly testing randomness / load). Was hoping to avoid re-engineering the test, but am happy to do so if that's appropriate.

jbolda:
Got through the remainder of the tests punting on a few (see the `.skip`). Should be able to finish up the remainder in the near term here.

lib/MultiColumnList/tests/MultiColumnList-test.js
jbolda: Chatting with @cowboyd, we ended up adding a `scrollBy` to the interactor. This should be resolved.

lib/MultiColumnList/tests/MultiColumnList-test.js
jbolda: I refactored out the `xy` to `row` and `column`. Switching `column` to the string version is a bit more work, but the last thing outstanding that we want to change on the interactor I believe.

jbolda:
This is ready to go, but I am unable to mark it ready. /cc @jgnieuwhof maybe can only do it?

cowboyd:
> This is ready to go, but I am unable to mark it ready. /cc @jgnieuwhof maybe can only do it?

It looks that way. As long as you preserve the commit history, you can create your own fork, then fetch this branch from @jgnieuwhof's repo, and then pushing that branch to your own fork, and closing old PR, reopening new one. All of the commit history is preserved so all parties get credit and life is great!

cowboyd:
@jgnieuwhof @jbolda @JohnC-80 

This should be good to close out right now that it has been superceded by what's now in main?

jgnieuwhof:
👍

PR #1483, opened 2021-01-06
UIEH-1017: Fix a11y errors on Modal and ConfirmationModal components

BogdanDenis:
## Description
A few ui-eholdings tests are failing due to accessibility issues [ARIA dialog and alertdialog nodes must have an accessible name](https://dequeuniversity.com/rules/axe/4.1/aria-dialog-name).
This PR is aimed at adding missing `aria-label` attribute to wrapping element of `<Modal>` and `<ConfirmationModal>` components

### Related PRs:
https://github.com/folio-org/stripes-smart-components/pull/979

## Issues
[UIEH-1017](https://issues.folio.org/browse/UIEH-1017)


lib/Modal/Modal.js
```diff
@@ -145,6 +145,7 @@ const Modal = forwardRef((props, ref) => {
                 enforceFocus={enforceFocus}
                 restoreFocus={restoreFocus}
                 renderBackdrop={renderBackdrop}
+                aria-label={rest['aria-label'] || props.ariaLabel}
```
JohnC-80, line 148: line 152 below is applying this prop to the WrappingElement should be fulfilling this so long as the prop is being applied to the Modal component... if it isn't working and placing this here is the true fix, please remove line 152. Thanks!!

JohnC-80 (changes requested):

lib/Modal/Modal.js
BogdanDenis: Applying `aria-label` to `WrappingElement` still gives an error, because that element doesn't have `role="dialog"`, but `<OverlayModal>` does

zburke: Thanks, @BogdanDenis, this explanation makes sense. Can you please remove Line 152 as @JohnC-80 requested?

BogdanDenis: @zburke sorry, I missed that I had to remove that line. Done now

zburke (approved):
Thanks for cleaning this up! Getting ARIA attributes correct can be really tricky and I really appreciate your attention to detail here.

PR #1484, opened 2021-01-08
STRIPES-721: Upgrade to react-redux 7

doytch:
This is part two of the PRs required to upgrade to react-redux 7 in Stripes.

I'm not sure if we'll be able to get CI builds passing in every repo during the process, but if it's possible, it'll probably be with PR merges in this order:

- [ ] [stripes-connect](https://github.com/folio-org/stripes-connect/pull/163)
- [ ] [stripes-core](https://github.com/folio-org/stripes-core/pull/985) and [stripes-components](https://github.com/folio-org/stripes-components/pull/1484)
- [ ] [stripes-smart-components](https://github.com/folio-org/stripes-smart-components/pull/981) and [stripes-form](https://github.com/folio-org/stripes-form/pull/135)
- [ ] [stripes](https://github.com/folio-org/stripes/pull/114)
- [ ] platforms


zburke:
Regrettably, this seems like a consistent failure: 
```
FAILED TESTS:
  AutoSuggest in redux-form
    rendering
      entering an existing value
        selecting an option
          ✖ sets the value of the input to the selected option
            Chrome 87.0.4280 (Linux 0.0.0)
          AssertionError: expected '' to equal 'book'

      + expected - actual

      +book
```

zburke:
On the one hand, this is a consistent test failure specific to redux-form that I don't really understand looking at the code. On the other, I don't see any uses of `<AutoSuggest>` in redux-form based repos across snapshot-core: 
```
$ grep -rl AutoSuggest . --exclude-dir node_modules | egrep -v "(test|artifacts|CHANGELOG)" 
./organizations/src/Organizations/OrganizationForm/OrganizationContactInfoForm/ContactInfoFormGroup/Urls.js
./organizations/src/Organizations/OrganizationForm/OrganizationContactInfoForm/ContactInfoFormGroup/PhoneNumbers.js
./organizations/src/Organizations/OrganizationForm/OrganizationContactInfoForm/ContactInfoFormGroup/EmailAddresses.js
./organizations/src/common/components/AddressInfo/AddressInfo.js
./tenant-settings/src/settings/LocationLocations/DetailsField.js
./stripes-components/index.js
./stripes-components/README.md
./stripes-components/lib/AutoSuggest/index.js
./stripes-components/lib/AutoSuggest/AutoSuggest.js
./stripes-components/lib/AutoSuggest/readme.md
./stripes-components/lib/AutoSuggest/stories/AutoSuggest.stories.js
./stripes-components/lib/AutoSuggest/stories/BasicUsage.js
./stripes-acq-components/lib/FieldAutoSuggest/FieldAutoSuggest.js
./stripes-acq-components/lib/FieldAutoSuggest/index.js
./stripes-acq-components/lib/FieldAutoSuggest/FieldAutoSuggestFinal.js
./stripes-acq-components/lib/index.js
```
@aditya-matukumalli, @doytch, any thoughts on the best way to proceed here? We could just deprecate `<AutoSuggest>` for redux-form, but that's pretty ugly, given that we normally provide 6-12 mos of warning that a component feature is deprecated before removing it. Alternatively, we need to fix `<AutoSuggest>` to make it function under `redux-form` `v8`.

doytch:
I haven't really been a fan of `AutoSuggest` ever since it's functionality kind of got superseded by `Selection`, so I honestly wouldn't be bothered by removing support for it w/ redux-form. I know the timing isn't _great_, but yeah, ain't nobody using it and really, nobody should be adding AutoSuggest w/ redux-form to their code anyway.

doytch:
But yeah, I poked at this for an hour yesterday and got nowhere with it. I don't really understand why this breaks.

aditya-matukumalli:
@zburke I spent some time on this this past hour but with no luck. This seems like some issue with how Downshifts internal state is working with the new redux form state. I didnt find any issues reported on their Github though. I havent used this component in a while and personally dont mind deprecating the redux form version here as its not being used elsewhere.

aditya-matukumalli:
Okay. So spent some more time here and was able to isloate the problem to this [line](https://github.com/folio-org/stripes-components/blob/master/lib/AutoSuggest/AutoSuggest.js#L120). 

So, the combination of _onChange_ on [L119](https://github.com/folio-org/stripes-components/blob/master/lib/AutoSuggest/AutoSuggest.js#L119) and _selectedItem_  on [L122](https://github.com/folio-org/stripes-components/blob/master/lib/AutoSuggest/AutoSuggest.js#L122) is whats binding the redux-forms value back to the selectedItem value. 

This value is getting overridden by the [onStateChange](https://github.com/downshift-js/downshift#onstatechange) callback which gets passed the inputValue as `undefined` (value from the internal state of Downshift) when the input changes, and thats what is failing the test.

Look how we are currently by-passing the RFF verison by passing a `withFinalForm` prop. Similarly ignoring the `onStateChange` prop altogether for Redux form as well solves the issue. (Tests pass, i.e.)

So somethings changed between the redux form upgrades thats causing the `onStateChange` to override the change made by the `onChange`, reasons still unknown to me.

Note: The tests also pass with the previous versions of redux, redux-form, react-redux with `onStateChange` prop taken off. So, maybe it wasnt doing anything all this while and the issue surfaced now that we have upgraded the versions.

zburke:
Wow, @aditya-matukumalli, this is dedicated troubleshooting! What do you think is the right approach here? 

As @doytch noted, ain't nobody using `AutoSuggest` with `redux-form`, and ain't nobody writing _new_ forms with `redux-form`, so maybe we just deprecate `redux-form` use with this component immediately and move on because ain't nobody gonna notice. 

Alternatively, since you figured out that ignoring `onStateChange`, as we do for `withFinalForm`, resolves the test issue, do you think we should simply omit that line? (Sure, the test will pass, but do you remember why you added that line to begin with?) I'm trying to wrap my head around the [Downshift documentation](https://github.com/downshift-js/downshift#onstatechange) but without any comments in our own code about how we use the `<Downshift>` component, I'm not feeling very enlightened.

aditya-matukumalli:
Looking at the code closely again, I honestly dont think `onStateChange` and `onInputValueChange` are even necessary (for either RF or RFF) for AutoSuggest. 
Note: I got the tests working using both RF and RFF `TestForm` versions with neither the `onStateChange` and `onInputValueChange` props.

So, in scenarios where `AutoSuggest` is rendered in a RF or RFF, `onChange` is coming in from the `input.onChange` from the FieldProps. In non-RF or non-RFF scenario, the `onChange` is passed in via `Downshift`. In the former case the `value` is pulled in from the store. In the latter its pulled in from the Downshifts internal state.

Considering there is much ambiguity here, I am inclined towards deprecating the support for Redux-form for the reasons Mark suggested. In any case, I just wanted to get to the bottom of the issue here so that we have some reference for us in the future.

PR #1488, opened 2021-01-14
STCOM-796 correctly set height for expanded accordions

zburke:
Expanded `<Accordion>`s had a `max-height` CSS property that caused
accordions with lots and lots of content to display incorrectly. The
content would overflow its container's bottom margin, overlapping the
following container.

Rather than setting an arbitrary `max-height` property, this PR unsets,
via `revert`, that property altogether, allowing the `max-height`
setting for _unexpanded_ accordions to stay as-is while allowing
_expanded_ accordions to adjust their height to that of their content.

Refs [STCOM-796](https://issues.folio.org/browse/STCOM-796)


ViktorSoroka07:
Did not know about this trick, nice one.

ViktorSoroka07 (approved):

MikeTaylor (approved):
Sweet!

PR #1500, opened 2021-01-28
STCOM-753 pass componentDidCatch arguments to a callback

zburke:
Given the callback function `onError` as a prop, call it from
`componentDidCatch` with the same arguments. While this does not
directly address STCOM-753 by logging the errors, it does allow the
component that instantiated this error boundary to provide a callback
that _does_ log the errors.

Refs [STCOM-753](https://issues.folio.org/browse/STCOM-753), [STCOR-455](https://issues.folio.org/browse/STCOR-455), [STCOR-493](https://issues.folio.org/browse/STCOR-493)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Add `className` prop to `<Accordion>`. Refs UIEH-926.
 * Settings > Panes are off. Refs STCOM-795.
 * Fix overflow in `sr-only` elements. Refs STCOM-801.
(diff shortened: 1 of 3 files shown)


ViktorSoroka07 (approved):

doytch (approved):
Nice! This was always my plan for this component in the future so I'm glad to see it's getting done!

mkuklis (approved):

ryandberger (approved):

PR #1504, opened 2021-02-04
UISEES-27: Update Checkbox component in the stripes-components library to include count in the label

Godlevskyi:
# Purpose
In this PR the new prop `count` is added.
This is needed for the ElasticSearch POC in order to show the count(number of hits) of the filter/facet.

For these the new prop labelInfo was added (labelInfo is more general and could be used not only to show the count)

labelInfo was added as part of the label in order to make it clickable and focusable.

# Screenshots
Checkbox with the count will be used in the CheckboxFilter:

Снимок экрана 2021-02-04 в 10 09 54

With Zero count:
Снимок экрана 2021-02-04 в 15 20 15


lib/Checkbox/Checkbox.js
```diff
@@ -222,6 +244,7 @@ class Checkbox extends React.Component {
               </svg>
             </span>
             {(!vertical && (label || readOnly)) && this.renderLabelText()}
+            {count && this.renderCount()}
```
ViktorSoroka07, line 247: As I understand this would no render `0` (Number.isInteger(0) or number >= 0)?

zburke: Nice catch, @ViktorSoroka07. Indeed, a test like the following fails: 
```
    describe('when the count is 0', () => {
      beforeEach(async () => {
        await mount(
          <Checkbox count={0} />
        );
      });

      it('should render 0', () => {
        expect(checkbox.count).to.equal('0');
      });
    });
```
What is the desired behavior when there are no matches, `0` or nothing?

zburke:
Nice work, @Godlevskyi. Please update the ticket with guidance about what to display when the count is zero so we know whether this can be approved as-is or if it needs to updated. Actually, there should probably be a zero-test like the one in my comment either way so the expectations about display are clear.

Godlevskyi:
Thanks @ViktorSoroka07  and @zburke , actually for the case of facets we would not have the data-options with 0 count.
But I believe you are right we need to show '0' in case it is provided for consistency.
Now it is fixed. Thanks!

lib/Checkbox/Checkbox.js
Godlevskyi: Thanks @ViktorSoroka07 , fixed

ViktorSoroka07 (approved):

lib/Checkbox/Checkbox.css
```diff
@@ -44,6 +44,14 @@
   }
 }
 
+/**
+ * Facet Count
+ */
+.checkBoxCount {
+  margin-left: auto;
+  padding: 0 10px;
```
doytch, line 52: Would it be more appropriate to use [`--gutter-static-two-thirds`](https://github.com/folio-org/stripes-components/blob/master/lib/variables.css#L41) for this instead?

doytch (commented):
This looks slick and all, but I'm not sure I'm really sold on this functionality being in `Checkbox`. 

In my opinion, a `count` is something quite separate from core checkbox functionality. I think a better fit for this would be [`CheckboxFilter` ](https://github.com/folio-org/stripes-smart-components/blob/321bd8305a8196eb464892573f6acca361714e16/lib/SearchAndSort/components/CheckboxFilter/CheckboxFilter.js). My thinking is that I can't really come up with example of UIs where you see a number next to a checkbox and it _isn't_ in the context of a filter. The description in the readme suggests the same.

zburke (approved):
LGTM! Maybe humor me and add one more test to prove this does not change the existing implementation? Your call; not required. 
```
interactor.js
  countIsPresent = isPresent(`.${styles.checkBoxCount}`);

Checkbox-test.js
  describe('a checkbox without a count-prop', () => {
    beforeEach(async () => {
      await mount(
        <Checkbox />
      );
    });

    it('should not render a count element', () => {
      expect(checkbox.countIsPresent).to.be.false;
    });
  });
```

mkuklis (commented):
This looks good but I also agree with @doytch that this should not belong in `Checkbox`. 
It feels more like a styling issue (displaying something on the right side of the checkbox row) which could be potentially used for other things not just `count`. Perhaps renaming it to a more generic thing or moving it to CheckboxFilter could work here.

Godlevskyi:
Thanks guys for your comments!
@zburke, added the negative test case that you suggested.
@mkuklis , @doytch  yeap I would agree that count prop is too specific for the general purpose checkbox. But I need count to be part of the label(clickable, focusable) and do not think that it is a good idea to duplicate the checkbox component just to include count.
I can rename count prop to something like: info or  infoBox or labelInfo. What do you think?

axelhunn:
It's maybe just a label's concern, not the Checkbox's.
The label in Checkbox is `PropTypes.node`, so I believe we could make a wrapper over Checkbox with this specific feature, rather than changing the core component.

mkuklis:
@Godlevskyi I like the `info` or `infoBox` or `labelInfo`.

Godlevskyi:
> It's maybe just a label's concern, not the Checkbox's.
> The label in Checkbox is `PropTypes.node`, so I believe we could make a wrapper over Checkbox with this specific feature, rather than changing the core component.

Thanks @axelhunn, yeap I also considered to pass it through the label prop, but in that case it will be part of the labelText and I believe having the meta data box with the count or what ever should be separated from the label text itself.

axelhunn:
@Godlevskyi Yes, I look closer at the code, and now I think I understand what you're talking about. Basically it's the `label` prop treated as string here.

I still don't feel good with adding such IMO specific thing to the general component.

Maybe refactor `renderLabelText` in a general way, something like that

```js
  renderLabelText = () => {
    const { label, required, readOnly } = this.props;

    const adornments = (
      <>
        {required && <Asterisk />}
        {readOnly && (
          <span className="sr-only">
            <FormattedMessage id="stripes-components.readonly" />
          </span>
        )}
      </>
    );

    if (typeof label === 'function') return label({ required, readOnly, adornments });

    return (
      <span className={css.labelText}>
        {label}
        {adornments}
      </span>
    );
  }

```

And then still make a wrapper for your special case?

axelhunn:
I believe it's a non-breaking change to the API.
On a brief look, I see 2 more places to make similar change: `<Label>` and `<RadioButton>` components, so we could make the API consistent.

doytch:
I like the idea of a functional `label`, but it could get annoying to always ensure that you're applying the adornments the right way. Or if `Checkbox` changes the way the adornments are rendered, then you'd have to make sure that anybody using functional `label` also changes their adornment rendering. 

What about something similar to `TextField`'s `endControl` that we always just plop in? Hell, we could even just call it `endControl` for uniformity across the API.
```
// inside `CheckboxFilter`...
<Checkbox endControl={Number.isInteger(count) && this.renderCount()} />

// inside `Checkbox`
  Checkbox.renderLabelText = () => {
    const { endControl, label, required, readOnly } = this.props;

    return (
      <span className={css.labelText}>
        {label}
        {required && <Asterisk />}
        {readOnly && (
          <span className="sr-only">
            <FormattedMessage id="stripes-components.readonly" />
          </span>
        )}
        {{ endControl && (
          <span className="make-this-pull-right">
            {endControl}
          </span>
         }}
      </span>
    );
  }
```

axelhunn:
@doytch Good idea!
But with `endControl` can we generalize it through `<Label>`?

axelhunn:
BTW, I don't see an icon for read-only state here, like in [`<Label>`](https://github.com/folio-org/stripes-components/blob/a2661a736040131167dcb240f907703cc85854b0/lib/Label/Label.js#L26)
Is it intentionally?

axelhunn:
> it could get annoying to always ensure that you're applying the adornments the right way. Or if `Checkbox` changes the way the adornments are rendered, then you'd have to make sure that anybody using functional `label` also changes their adornment rendering.

We can provide a full service 
```js
    const adornText = text => (
      <span className={css.labelText}>
        {text}
        {adornments}
      </span>
    );

    if (typeof label === 'function') return label({ required, readOnly, adornments, adornText });

    return adornText(label);
```

Godlevskyi:
> I like the idea of a functional `label`, but it could get annoying to always ensure that you're applying the adornments the right way. Or if `Checkbox` changes the way the adornments are rendered, then you'd have to make sure that anybody using functional `label` also changes their adornment rendering.
> 
> What about something similar [to `TextField`'s `endControl`](https://github.com/folio-org/stripes-components/blob/master/lib/TextField/TextField.js#L62) that we always just plop in? Hell, we could even just call it `endControl` for uniformity across the API.
> 
> ```
> // inside `CheckboxFilter`...
> <Checkbox endControl={Number.isInteger(count) && this.renderCount()} />
> 
> // inside `Checkbox`
>   Checkbox.renderLabelText = () => {
>     const { endControl, label, required, readOnly } = this.props;
> 
>     return (
>       <span className={css.labelText}>
>         {label}
>         {required && <Asterisk />}
>         {readOnly && (
>           <span className="sr-only">
>             <FormattedMessage id="stripes-components.readonly" />
>           </span>
>         )}
>         {{ endControl && (
>           <span className="make-this-pull-right">
>             {endControl}
>           </span>
>          }}
>       </span>
>     );
>   }
> ```
Changed count prop to more general labelInfo.
Not the end control as first of all now it is not intended to be control, just some meta information increment for the label.
And as we provide the className for customization it may be positioned not only at the end,

Godlevskyi:
> > it could get annoying to always ensure that you're applying the adornments the right way. Or if `Checkbox` changes the way the adornments are rendered, then you'd have to make sure that anybody using functional `label` also changes their adornment rendering.
> 
> We can provide a full service
> 
> ```js
>     const adornText = text => (
>       <span className={css.labelText}>
>         {text}
>         {adornments}
>       </span>
>     );
> 
>     if (typeof label === 'function') return label({ required, readOnly, adornments, adornText });
> 
>     return adornText(label);
> ```

As for me the label prop is not associated  with the function. And I believe we will need the to teach community to use the functional label.  IMHO for the functional label  we may have renderLabel prop instead.
Anyway, I believe all the approaches mentioned could be considered for the future stripes improvement, but i believe it will take long disussions for the API unification between components, etc
In our case we do not want to unify the API and the extension of the Checkbox component with the generalized labelInfo is an approach to start from to  delver the elasticSearch POC in time. And later on we may dispute on the unification of the stripes components API including the functional labels. IMHO

axelhunn:
> we will need to teach the community to use
We are changing (appending) the API anyway.

Godlevskyi:
> > we will need to teach the community to use
> 
> We are changing (appending) the API anyway.

Yeap, I believe it is a good Idea ti have the functional label which will be unified across the stripes components. 
But, I believe this should be discussed on the stripes architecture meeting. And separate task/scope should be allocated for it.

For now I will extend CheckBox with the labelInfo prop, following the existing FOLIO approach.. And regarding the functional label we can create the stories after the discussion on stripes arch meeting. @axelhunn  could you present func label approach on the next stripes arch meeting?

@zburke , @mkuklis, @doytch, @ViktorSoroka07  what do you think?

PR #1505, opened 2021-02-08
UISEES-33: Replacing existing query search

Dmytro-Melnyshyn:
This PR for ElasticSearch POC.
MultiSelectSearch is a new component for ElasticSearch POC - it is at the initial stage of development.

When a user selects the 'Advanced Search' mode, I need to change the TextField component to the new MultiSelectSearch. 
For this, I need to make changes to the SearchField component of stripes-components.
Since this is a proof of concept, I created a clone of the SearchField component, I named it MultiSelectSearchField and worked with it. 
I expanded it by adding a condition for rendering the desired component (113-139 lines) [https://github.com/folio-org/stripes-components/compare/UISEES-33?expand=1#diff-fa839c4a3fe6f46677966c3d9b48f406424a903348bfb27305dee157b0fb6b50](url)

Wrote two tests into cloned tests to check if the condition works (28 - 45 lines) https://github.com/folio-org/stripes-components/compare/UISEES-33?expand=1#diff-5dda0dc10a42ea40d5f7fd96799e4b6923160cb71c25e7f3fa864c5d916a154d

I expanded the cloned storybook with new properties (isAdvancedSearch, searchButtonRef) https://github.com/folio-org/stripes-components/compare/UISEES-33?expand=1#diff-b029b0edd8fbef10c15aa114289913f10f93a3e2406881adb34f0136d05a9a09

Refs [UISEES-33](https://issues.folio.org/browse/UISEES-33)

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,7 @@ export {
 export { default as ConfirmationModal } from './lib/ConfirmationModal';
 export { default as InfoPopover } from './lib/InfoPopover';
 export { default as SearchField } from './lib/SearchField';
+export { default as MultiSelectSearchField } from './lib/MultiSelectSearchField';
 
 /* specific use */
 export {
--- a/lib/MultiSelectSearchField/MultiSelectSearchField.css
+++ b/lib/MultiSelectSearchField/MultiSelectSearchField.css
@@ -1,0 +1,25 @@
+/**
+ * SearchField
+ */
+
+@import "../variables.css";
+
+/**
+ * Has searchable indexes
+ */
+
+/* Select field */
+.searchFieldWrap .select {
+  /* To make sure it's position is above
+  so the focus style isn't hidden by the input */
+  margin-bottom: -1px;
+
+  &:focus {
+    z-index: 5;
+  }
+}
+
+/* Color search icon on focus */
+.searchFieldWrap .isFocused svg.searchIcon {
+  fill: var(--primary);
+}
--- a/lib/MultiSelectSearchField/MultiSelectSearchField.js
+++ b/lib/MultiSelectSearchField/MultiSelectSearchField.js
@@ -1,0 +1,150 @@
+/**
+ * SearchField
+ *
+ * A universal search field component
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { useIntl } from 'react-intl';
+
+import TextField from '../TextField';
+import Select from '../Select';
+import TextFieldIcon from '../TextField/TextFieldIcon';
+import MultiSelectSearch from './components/MultiSelectSearch';
+import css from './MultiSelectSearchField.css';
+
+// Accepts the same props as TextField
+const propTypes = {
+  ariaLabel: PropTypes.string,
+  className: PropTypes.string,
+  clearSearchId: PropTypes.string,
+  disabled: PropTypes.bool,
+  id: PropTypes.string,
+  inputClass: PropTypes.string,
+  inputRef: PropTypes.object,
+  isAdvancedSearch: PropTypes.bool,
+  loading: PropTypes.bool,
+  onChange: PropTypes.func,
+  onChangeIndex: PropTypes.func,
+  onClear: PropTypes.func,
+  placeholder: PropTypes.string,
+  searchableIndexes: PropTypes.arrayOf(PropTypes.shape({
+    disabled: PropTypes.bool,
+    id: PropTypes.string,
+    label: PropTypes.string,
+    placeholder: PropTypes.string,
+    value: PropTypes.string,
+  })),
+  searchableIndexesPlaceholder: PropTypes.string,
+  searchButtonRef: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
+  selectedIndex: PropTypes.string,
+  value: PropTypes.string,
+};
+
+const MultiSelectSearchField = (props) => {
+  const {
+    className,
+    placeholder,
+    id,
+    ariaLabel,
+    value,
+    onChange,
+    onClear,
+    loading,
+    clearSearchId,
+    searchableIndexes,
+    onChangeIndex,
+    selectedIndex,
+    searchableIndexesPlaceholder,
+    inputClass,
+    disabled,
+    isAdvancedSearch,
+    searchButtonRef,
+    ...rest
+  } = props;
+
+  /**
+   * Search field has searchable indexes dropdown
+   */
+  let searchableIndexesDropdown;
+  const hasSearchableIndexes = Array.isArray(searchableIndexes);
+  const intl = useIntl();
+
+  if (hasSearchableIndexes) {
+    const indexLabel = intl.formatMessage({ id: 'stripes-components.searchFieldIndex' });
+
+    searchableIndexesDropdown = (
+      <Select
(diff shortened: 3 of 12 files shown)


doytch (commented):
I think I'd actually be interested in seeing how the `SearchField` API will change. It seems like the main thing that needs to be changed in the rendering is using `Selection`/whatever instead of `Select` if you want multiselect mode. (Maybe we should consider whether we should always render a `Selection` or not? Or whether we want to update `Select` to have native multiselect? I find the different capabilities of Selection and Select a bit confusing already 😆)

Dmytro-Melnyshyn:
> I think I'd actually be interested in seeing how the `SearchField` API will change. It seems like the main thing that needs to be changed in the rendering is using `Selection`/whatever instead of `Select` if you want multiselect mode. (Maybe we should consider whether we should always render a `Selection` or not? Or whether we want to update `Select` to have native multiselect? I find the different capabilities of Selection and Select a bit confusing already 😆)

MultitSelectSearch is a new component that will be able to accept search queries as a search option - operator - term - boolean operator structure with parenthesis support. The user will be able to make queries similar to the JIRA Advanced search.
examples:
  hrid == "inst000000000018" AND title all "some title" OR ...
  title all = ("title 1" OR "title 2") OR ...

doytch:
> > I think I'd actually be interested in seeing how the `SearchField` API will change. It seems like the main thing that needs to be changed in the rendering is using `Selection`/whatever instead of `Select` if you want multiselect mode. (Maybe we should consider whether we should always render a `Selection` or not? Or whether we want to update `Select` to have native multiselect? I find the different capabilities of Selection and Select a bit confusing already 😆)
> 
> MultitSelectSearch is a new component that will be able to accept search queries as a search option - operator - term - boolean operator structure with parenthesis support. The user will be able to make queries similar to the JIRA Advanced search.
> examples:
> hrid == "inst000000000018" AND title all "some title" OR ...
> title all = ("title 1" OR "title 2") OR ...

OK so I think I misunderstood the point of this then...here's my understanding so tell me where I'm wrong. 😛 

1. At first, I thought the point of this feature was to create a `SearchField`-type component that allowed you to use a multiselect for the field selection. So you could type in `123456` into the TextField; select multiple options, eg, `ID`, `Barcode`, `External system ID`; and it would search for records where `id == 123456 OR barcode == 12346 OR externalSystemId == 123456`.
2. Based on your comment, and now that I've actually run the Storybook (🤦), it appears that there are _no_ visible difference between `SearchField` and `MultiSelectSearchField`. Technically, MSSF renders a textarea rather than an `input`, but they're visually identical. I even couldn't actually insert line breaks despite expecting to be able to.
3. Based on your comment, it seems like the whole point here is for MSSF to allow users to manually type in complex CQL (or CQLish) queries.

Assuming I'm correct on those points (and I'm very uncertain I am!), then I don't really understand what the purpose of this component is. If the purpose of MSSF is to allow users to type in complex queries, why can't they just use a regular `SearchField` who's `value` is parsed by the parent components?

zburke (commented):
I'm with @doytch on this; it isn't clear to me what the purpose of this component is. The difference between "Query search" and, say, "Keyword search" in ui-inventory is in how the value is parsed, not in what component use used to accept that value.

Dmytro-Melnyshyn:
> I'm with @doytch on this; it isn't clear to me what the purpose of this component is. The difference between "Query search" and, say, "Keyword search" in ui-inventory is in how the value is parsed, not in what component use used to accept that value.

The purpose of the MultiSelectSearch component to perform all scenarios of the [story.](https://issues.folio.org/browse/UISEES-26)

If using the TextField component to implement the story I need to add into it dropdown (2-4 scenarios), a new code (5-8 scenarios), `<textarea />` tag - multiline (9 scenario) and pass the parser.

In the future, new stories will appear, expanding the functionality and in order not to add everything to the TextField (Query search) component, MultiSelectSearch is created.

JohnC-80:
@Dmytro-Melnyshyn Thank you for the work! I'm not sure that we should centralize this/publish it immediately - especially if there's more changes/features left to add to the component. I recommend you keep this within the elastic search POC to fulfill its requirements and then we can migrate it to a shared component after. It'll help you move through those requirements quicker than hitting us up for review at every step... but it's cool to expose the progress/eventual capabilities, so thank you for letting us see that! 
From what I see here, the code looks fine (but probably shouldn't allow the textarea to expand multiple lines rather than set its height via style attribute.) 
I would consider a different name for centralizing the component, as 'MultiSelectSearch' is a bit confusing since we already have a MultiSelection component used for a 'tags' like interface as well as 'MultiSelectionFilter' in stripes-smart-components. I could suggest 'MultilineSearch' or even 'ElasticQueryField'. 
Alternatively, as @doytch did hit on a bit, consider if this component could work as the result of additional props (render functions) that could be added to SearchField - (keeping its current common behavior - using a `<select>` and `<TextField>` as default functionality) making it more flexible. This would avoid the whole naming issue altogether, and we can document the setup use for Elastic Search as a pattern/use-case.
Elastic search has strong potential to be 'the new way' to do searches in FOLIO, so I'm excited about seeing the progress here.
To reiterate - working in the POC project will probably help you move much faster and we can see what it will take to centralize the necessary functionality after seeing the results there. Thank you!!

Dmytro-Melnyshyn:
> @Dmytro-Melnyshyn Thank you for the work! I'm not sure that we should centralize this/publish it immediately - especially if there's more changes/features left to add to the component. I recommend you keep this within the elastic search POC to fulfill its requirements and then we can migrate it to a shared component after. It'll help you move through those requirements quicker than hitting us up for review at every step... but it's cool to expose the progress/eventual capabilities, so thank you for letting us see that!
> From what I see here, the code looks fine (but probably shouldn't allow the textarea to expand multiple lines rather than set its height via style attribute.)
> I would consider a different name for centralizing the component, as 'MultiSelectSearch' is a bit confusing since we already have a MultiSelection component used for a 'tags' like interface as well as 'MultiSelectionFilter' in stripes-smart-components. I could suggest 'MultilineSearch' or even 'ElasticQueryField'.
> Alternatively, as @doytch did hit on a bit, consider if this component could work as the result of additional props (render functions) that could be added to SearchField - (keeping its current common behavior - using a `<select>` and `<TextField>` as default functionality) making it more flexible. This would avoid the whole naming issue altogether, and we can document the setup use for Elastic Search as a pattern/use-case.
> Elastic search has strong potential to be 'the new way' to do searches in FOLIO, so I'm excited about seeing the progress here.
> To reiterate - working in the POC project will probably help you move much faster and we can see what it will take to centralize the necessary functionality after seeing the results there. Thank you!!

@JohnC-80 Thanks for the review! Very helpful tips, I fully agree with you. Do you propose to copy all the components I need from the shared components into the ui-inventory-es module (POC module) and work with them, and when all the necessary functionality is completed, then create PRs for the repositories of the shared components? Working on the POC project will definitely help me move faster. Thank you!

JohnC-80:
@Dmytro-Melnyshyn copy them only if you find they're inadequate/need to change them, sure! If you have any questions about whether you can perform a particular functionality with a component, please feel free to reach out to me on slack (j.coburn) If I can't give you a direct answer, I'll loop in folks who can :) Thanks! We'll review the changes after the functional POC is in place - then we can spend time conforming APIs/convention/styling to the rest of the system if necessary.

PR #1509, opened 2021-02-12
Finish conversion from Tether to Popper.

JohnC-80:
We still had 2 lingering usages of the Tether dependency within the library. Complete removal of the Tether dependency will have to wait for further usage analysis, but this change moves the last of our components that were using Tether over to using the Popper.js library.


lib/Selection/SingleSelect.js
```diff
@@ -767,15 +772,26 @@ class SingleSelect extends React.Component {
...
+          </RootCloseWrapper>
+        </TetherComponent>
+      );
+    }
 
     return (
-      <TetherComponent {...mergedTetherProps}>
+      <>
         {uiInput}
-        <RootCloseWrapper onRootClose={this.handleRootClose} ref={this.selectList}>
```
rasmuswoelk, line 775: We still need the RootCloseWrapper when using Popper. The Selection component is not currently closing on outside click or when hitting escape.

rasmuswoelk (changes requested):
Finally.. we get rid of tether for good! We should celebrate 😄 🎊 

Selection changes look good and it works as expected in Storybook. I pushed a minor fix for the example.

When testing the AutoSuggest, I noticed that it does not close on outside click or when you hit escape. This is most likely because the RootCloseWrapper is missing.

lib/Selection/SingleSelect.js
JohnC-80: Updated (finally!)

rasmuswoelk (approved):

PR #1510, opened 2021-02-15
STCOM-809 Implement warning callout

axelhunn:
https://issues.folio.org/browse/STCOM-809
Adding Callout type `warning` , like in MessageBanner, same style

image

### DIFF
--- a/lib/Callout/Callout.css
+++ b/lib/Callout/Callout.css
@@ -12,38 +12,6 @@
   padding: 1.5rem 1rem;
 }
 
-.base {
-  padding: 1rem;
-  border-radius: var(--radius);
-  background-color: #fff;
-  width: 35%;
-  pointer-events: all;
-  position: relative;
-  transition: all 0.2s ease;
-  display: flex;
-  box-shadow: var(--shadow);
-
-  &.error {
-    background-color: color(var(--error, #900) tint(25%));
-    border-color: var(--error);
-    color: #fff;
-
-    & :global .stripes__icon {
-      fill: #fff;
-    }
-  }
-
-  &.success {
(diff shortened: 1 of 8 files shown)


zburke (commented):
The code you added looks good, but can you please explain why all the `.base` style information was removed? The amount of code you added is dwarfed by what you removed. If it was unused cruft or something, that's fine, just please explain what's going on.

doytch (approved):

axelhunn:
> The code you added looks good, but can you please explain why all the `.base` style information was removed? The amount of code you added is dwarfed by what you removed. If it was unused cruft or something, that's fine, just please explain what's going on.

I think it's unused, yes. Changing in doesn't change anything. I tried search usage of `base` in the code - seems nothinng

lib/Callout/Callout.css
```diff
@@ -88,6 +56,11 @@
     background-color: color(var(--success, #060) tint(75%));
```
axelhunn, line 56: I'm not sure about this `color` function
Does it work anywhere except Safari?
https://caniuse.com/css-color-function

image

axelhunn: Maybe just use `--success-fill` here?

axelhunn: Or maybe the opacity in `-fill` colors is too much for the Callouts
image

ViktorSoroka07 (approved):

NikitaSedyx (approved):

lib/Callout/Callout.css
axelhunn: OK, for the `warning` I make the color function, the same as in other types.
I have no idea for now, why it does work in production, but does not in the storybook.

PR #1511, opened 2021-02-17
STCOM-785: switch to use mainline @folio/stripes-testing

cowboyd:
We now have working set of interactors merged onto master, and so we can use that branch now.


taras (approved):
👍

cowboyd:
@taras @JohnC-80 Do we need to do something to update the build process to account for the stripes build changes?

PR #1516, opened 2021-03-01
STCOM-812 Title Record: Search within Packages Modal. Scrolling list of Packages is overlay with Selection filter

BogdanDenis:
## Description
Fix an issue with `<Accordion>` where if you have scrollable content as child of accordion and you try scrolling it using the scrollbar then this content will display below other accordions next to it

## Approach
This issue is caused by elements losing focus when their content is scrolled using a scrollbar. This is possibly a browser issue - I've found Chromium issue that describes similar behavior with status Won't Fix (https://bugs.chromium.org/p/chromium/issues/detail?id=6759)

The solution that I propose is checking when a user has clicked outside of accordion's component tree using a `click` event listener on the document instead of `blur` event. In the handler of `click` we can check whether the event originated inside the component tree or not. And then based on that we can set `z-index` of accordion.

## Screenshots
With bug
eHoldings - Journal - FOLIO (1)

No bug
eHoldings - Journal - FOLIO

## Issues
[STCOM-812](https://issues.folio.org/browse/STCOM-812)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,7 +11,7 @@
   "extends": "@folio/eslint-config-stripes",
   "overrides": [
     {
-      "files": ["lib/**/tests/**", "util/tests/*"],
+      "files": ["lib/**/tests/**", "util/tests/*", "hooks/tests/*"],
       "rules": {
         "func-names": "off",
         "max-classes-per-file": "off",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-components
 
+## [9.1.0] IN PROGRESS
+
+* Fix Accordion content is displayed below other accordions when using scrollbar. Fixes STCOM-812.
+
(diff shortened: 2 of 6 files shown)


Vladyslav-Velytskyi (approved):

zburke (approved):
Your explanation was really helpful, @BogdanDenis, both in terms of explaining the problem and why this solution works. Thank you! 

That said, before merging, please let's wait for approval from @JohnC-80 and/or @rasmuswoelk, who are more familiar with the internals here and are more likely than I to be aware of any gotchas or potential side-effects I am unaware of.

zburke:
(Assuming this _does_ look good, I'll also plan to cherry pick this onto a `v9.0.1` release so it can be part of R1/Iris. I think we had a UIIN ticket for this too but I can't find it now. Does this sound familiar, @mkuklis?)

mkuklis:
@zburke that sounds familiar but I think that problem went away at some point. I can't find a JIRA for it now. Nice work @BogdanDenis!

PR #1521, opened 2021-03-10
STCOM-815: Landmarks must have a unique role or role/label/title (AxeDevTools Error)

IhorKoba:
## Ticket
https://issues.folio.org/browse/STCOM-815

## Changes
- Add `aria-label` attribute to default accordion header of `<DefaultAccordionHeader>` component to fix error about uniqueness of landmarks.

## Before
image

## After
image

### DIFF
--- a/lib/Accordion/headers/DefaultAccordionHeader.js
+++ b/lib/Accordion/headers/DefaultAccordionHeader.js
@@ -19,7 +19,9 @@ const propTypes = {
 };
 
 const defaultProps = {
-  headingLevel: 3
+  headerProps: {
+    headingLevel: 3,
+  }
 };
 
 const DefaultAccordionHeader = (props) => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 * `<ErrorBoundary>` accepts a callback to receive the arguments to `componentDidCatch`. Refs STCOM-753.
 * Increment `@folio/stripes-cli` to `v2`. Refs STCOM-806.
 * Added `refresh`-icon. Refs UX-426.
+* Fix rendering default heading as `h3` tag for accordion of `<DefaultAccordionHeader>` component. Refs STCOM-815
 
 ## [8.0.0](https://github.com/folio-org/stripes-components/tree/v8.0.0) (2020-10-05)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v7.0.1...v8.0.0)
--- a/lib/Accordion/Accordion.js
+++ b/lib/Accordion/Accordion.js
@@ -171,7 +171,6 @@ Accordion.propTypes = propTypes;
 Accordion.defaultProps = {
   className: '',
   header: DefaultAccordionHeader,
-  headerProps: {},
   separator: true,
 };
 


zburke (commented):
I understand the a11y violation needs to be addressed, and this change technically does that, but I'm not sure a value like `accordion-toggle-button-foo` meets the spirit of the [description](https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html#ARIA14-description), "an accessible name". OTOH, maybe something is better than nothing? Dunno.

@JohnC-80, any thoughts here?

IhorKoba:
> I understand the a11y violation needs to be addressed, and this change technically does that, but I'm not sure a value like `accordion-toggle-button-foo` meets the spirit of the [description](https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html#ARIA14-description), "an accessible name". OTOH, maybe something is better than nothing? Dunno.
> 
> @JohnC-80, any thoughts here?

@zburke , I agree , that it would be nice to have some informative name, but as for me in that case it's more important to have correct connection between aria-label and aria-labeledby attributes like there https://dequeuniversity.com/rules/axe/4.1/landmark-unique?application=AxeChrome, and we have a lot of accordions with dynamically generated these names.

zburke:
Would it be better to stick an `id` on the `<div>` on line 80 and point to that with `aria-labelledby` here?

lib/Accordion/headers/DefaultAccordionHeader.js
```diff
@@ -66,6 +66,7 @@ const DefaultAccordionHeader = (props) => {
             autoFocus={props.autoFocus}
             aria-expanded={open}
             aria-controls={props.contentId}
+            aria-label={`accordion-toggle-button-${id}`}
```
JohnC-80, line 69: With regions, most if not all examples show them as labeled with an heading tag.. This is probably best for the regions role as a section of a page. DefaultAccordionHeader has `headingLevel: 3` as a default prop, which should render as an h3 when fed to the Headline component that it renders... (see line 59)

For some reason, this seems to be coming up as 'undefined' once it reaches the Headline.

I believe that once an h3 tag is rendered, rather than a div or other non-semantic elment, it will resolve this issue. In other places I see the tag has been provided in some other way - either ran explicit `headerProps: { headingLevel: 2}` or something like `label={<Headline tag="h3" ... ` `aria-labelledby should work fine on its own for the offending content "region"...

JohnC-80 (changes requested):

lib/Accordion/headers/DefaultAccordionHeader.js
JohnC-80: We shouldn't do this with `aria-label` as it will override any meaningful text that's within the header for assistive technology. As you mentioned, with these id's being generated, it would announce something like "accordion-toggle-button-accordion6" 
- probably not quite explicit enough for an accessibility review.

IhorKoba: Thanks for your comment!

So, for render default heading I thing we need to pass `defaultProps` in this way :
`const defaultProps = {
  headerProps: {
    headingLevel: 3,
  }
};`
instead of
`const defaultProps = {
  headingLevel: 3,
};`
because we’re getting this `headingLevel`  from `headerProps`  on line 46, and remove `headerProps` as an empty object from `<Accordion>`  component.
In that case `DefaultAccordionHeader`  component will be rendered with correct `h3` tag.
I applied that changes locally and now correct heading renders:

image

But when I run Axe check I steel got the same error message:

image

JohnC-80:
Commenting for persistence - So after some discussion about this issue, the fix to have the defaultAccordionHeader render its h3 out-of-box is a necessary one.  So we definitely want to keep that portion of this PR.

The culprit of the axe error is multiple rendered accordions with **Matching visible text labels**.  ie - an accordion named "Information" in the detail and the corresponding "Information" accordion in the edit page. This can happen in any application that uses the `<Layer>` component for rendering an edit view.

I feel that, given the context around the accordion, and the general DOM distance between the matching accordions, there probably won't be much *actual confusion around this from the AT user perspective - this error isn't a dire emergency - NOT having a heading structure, on the other hand, is one that should definitely be resolved (and this PR helps that in its new form. - thank you @IhorKoba !)

Other ways to resolve (aka, right things to do) - 
 - conditional rendering: if using `<Layer>`, hide the detail view's accordions when the edit view is open .
  or
 - use actual routing so that the only elements rendered are those within the form/Layer
  or
 - name the edit view accordions differently (probably more redundant than practical, but not to be left out)

We're going to include this documentation on the accordion component at first for others encountering the issue... and probably find an even more public place to put it with gotchas we run into with axe errors.

JohnC-80 (approved):

PR #1528, opened 2021-03-22
STCOM-794: Add ErrorModal component

SergiyVSergiyenko:
https://issues.folio.org/browse/STCOM-794

## Purpose
Create `ErrorModal` component.

## Approach
These changes are made as part of the concept of creating a small, simple, and rather "dumb" (highly specialized) component in order to display the error in a friendly way.
Therefore, the number of passed properties has been minimized. And as a consequence, in this version, there is **no** possibility:
- to pass your own footer;
- to customize the `Close` button (stylize, change its title);
- to make this modal `dismissible`;
- to have the `closeOnBackgroundClick` prop;
- etc...

It’s not claimed that is the only correct approach. Any suggestions and objections are greatly appreciated. As well as suggestions for the naming of some passed props:

`label`: variants - `heading`, `header`, `title` ...
`content`: variants - `message`, `body` ...
`open`: variant - `isOpen`

For reference, here is a list of the local `ErrorModal` components that were taken as a basis
(name as a link to the local `ErrorModal`):

[ui-users](https://github.com/folio-org/ui-users/blob/master/src/components/ErrorModal/ErrorModal.js)
[ui-inventory](https://github.com/folio-org/ui-inventory/blob/master/src/components/ErrorModal/ErrorModal.js)
[ui-requests](https://github.com/folio-org/ui-requests/blob/master/src/components/ErrorModal.js)
[ui-orders](https://github.com/folio-org/ui-orders/blob/master/src/components/PurchaseOrder/UpdateOrderErrorModal/UpdateOrderErrorModal.js)
[ui-checkin](https://github.com/folio-org/ui-checkin/blob/6ed8c7ff08fa8dc89c81a36ca6f02ba6afec395c/src/Scan.js#L749)
[ui-checkout](https://github.com/folio-org/ui-checkout/blob/master/src/components/ErrorModal/ErrorModal.js)

### Screenshots
ErrorModal

### DIFF
--- a/lib/ErrorModal/tests/ErrorModal-test.js
+++ b/lib/ErrorModal/tests/ErrorModal-test.js
@@ -1,0 +1,50 @@
+import React from 'react';
+
+import { expect } from 'chai';
+import sinon from 'sinon';
+import { describe, beforeEach, it } from '@bigtest/mocha';
+import { mountWithContext } from '../../../tests/helpers';
+
+import ErrorModal from '../ErrorModal';
+import ErrorModalInteractor from './interactor';
+
+describe('ErrorModal', () => {
+  const errorModal = new ErrorModalInteractor();
+  const onClose = sinon.spy();
+
+  const label = 'Something went wrong';
+  const content = 'Here is a detailed message that explains why the error occurred.';
+
+  beforeEach(async () => {
+    await mountWithContext(
+      <ErrorModal
+        open
+        label={label}
+        content={content}
+        onClose={onClose}
+      />
+    );
+  });
+
+  it('the label text should be correct', () => {
+    expect(errorModal.label.text).to.equal(label);
+  });
+
+  it('the content text should be correct', () => {
+    expect(errorModal.content.text).to.equal(content);
+  });
+
+  it('the content tagName should be div', () => {
+    expect(errorModal.bodyTagName).to.equal('div');
+  });
+
+  describe('when clicking the close button', () => {
+    beforeEach(async () => {
+      await errorModal.closeButton.click();
+    });
+
+    it('the onClose callback should be fired', () => {
+      expect(onClose.called).to.be.true;
+    });
+  });
+});
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,9 @@
 ## [9.1.0] IN PROGRESS
 
 * Fix Accordion content is displayed below other accordions when using scrollbar. Fixes STCOM-812. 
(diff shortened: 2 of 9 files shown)


doytch (approved):
I think this approach makes sense. Let's get the basic one in so that we don't accidentally style them differently (`Close` vs `OK` vs `Dismiss`), and if the UX designers truly want more differentiation then it can be added down the road.

mkuklis (approved):
Nice work @SergiyVSergiyenko!

Baroquem (approved):

annamelnyk (approved):

zburke (approved):
This looks great to me! Seeing that long list of local error components really makes clear the value of providing a single component to handle this consistently.

The only question I have is whether we need separate `label` and `aria-label` props, i.e., what are the circumstances when we'd want those values to be different? The example value in the test, `simple-error-modal` is almost certainly wrong. Even if it were changed to `simple error modal`, is that actually better than `Something went wrong`, or maybe, `Error modal: Something went wrong`? 

@JohnC-80, any thoughts on this from an a11y angle?

doytch:
> The only question I have is whether we need separate `label` and `aria-label` props, i.e., what are the circumstances when we'd want those values to be different? The example value in the test, `simple-error-modal` is almost certainly wrong. Even if it were changed to `simple error modal`, is that actually better than `Something went wrong`, or maybe, `Error modal: Something went wrong`?

I'm not even a discount-bin version of John wrt a11y, but I think most people wouldn't pass an `ariaLabel` prop to this component because the label would be read out correctly and be accurate. You generally only want to specify `aria-label` if it's going to be _different_ than the visible label because otherwise the screenreader reads out the same thing twice which is jarring.

There's also the issue of `ariaLabel` **must** be a String, but `label` should support any React `node`.

lib/ErrorModal/tests/ErrorModal-test.js
```diff
@@ -0,0 +1,51 @@
...
+
+  beforeEach(async () => {
+    closed = false;
+
+    await mountWithContext(
+      <ErrorModal
+        open
+        label={label}
+        content={content}
+        onClose={() => { closed = true; }}
```
ViktorSoroka07, line 25: `sinon.spy()` would be a better fit IMO. [Here](https://github.com/folio-org/stripes-components/blob/master/lib/Pane/tests/Pane-test.js#L46) is the usage in Pane-test.

ViktorSoroka07 (approved):

lib/ErrorModal/tests/ErrorModal-test.js
SergiyVSergiyenko: Thank you. Done.

PR #1530, opened 2021-03-29
Add check-in and check-out icons

rasmuswoelk:
https://issues.folio.org/browse/UX-427
https://issues.folio.org/browse/UX-428

image


zburke:
@rasmuswoelk: this check-in icon looks backwards to me, i.e. it's really a left-handed check-out rather than a check-in. Filed [STCOM-826](https://issues.folio.org/browse/STCOM-826)

CC: @JohnC-80

PR #1536, opened 2021-04-07
STCOM-824 set max-width for textarea elements

zburke:
Previously, `max-width` applied only on the container element, not on
the `textarea` element itself. This allowed `textarea`s to be resized
beyond the widths of their containers, which was terrible.

Refs [STCOM-824](https://issues.folio.org/browse/STCOM-824)


mkuklis (approved):
👏

PR #1542, opened 2021-04-16
STCOM-833 Paneset: don't call setState on unmounted components

zburke:
Wrap calls to `setState` in checks to make sure the components is still
mounted.

Refs [STCOM-833](https://issues.folio.org/browse/STCOM-833)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-components
 
+## 9.2.0 (IN PROGRESS)
+
+* `Paneset`: Don't call `setState` on unmounted components. Refs STCOM-833.
+
 ## [9.1.0](https://github.com/folio-org/stripes-components/tree/v9.1.0) (2021-04-08)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v9.0.0...v9.1.0)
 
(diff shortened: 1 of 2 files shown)


mkuklis (approved):
Nice cleanup @zburke!

Baroquem (approved):

JohnC-80 (approved):

JohnC-80:
We could probably handle this differently too by setting the timeouts to variables/refs and clearing them in ComponentWillUnmount or equiv (cleanup function on useEffect with empty dependency array) Conditional is definitely one way to go about it... the checks just end up everywhere and always end up flagged in testing as 'oo, an uncovered branch!'

zburke:
@JohnC-80, yeah, I figured the timeouts were suspicious. If you have a sense of how to repair this with a less blunt method than `if (this._isMounted) ...` I would love to see it. I am not especially familiar with the code but I am familiar with the console noise (and we have so. much. console noise, hence my blunt approach to getting rid of it).

zburke:
Closing in favor of #1545, where timers are handled The Right Way with surgically precise `clearTimeout()` calls, rather than the blanket `_isMounted` conditional here.

PR #1541, opened 2021-04-16
STCOM-832 TextField: don't call setState on unmounted components

zburke:
Wrap calls to `setState` in checks to make sure the component is still
mounted.

Refs [STCOM-832](https://issues.folio.org/browse/STCOM-832)

PR #1545, opened 2021-04-19
WIP: Clear timeouts on unmount

JohnC-80:
ref to #1541 and #1542 
Addresses timeouts in Paneset, Timepicker, Textfield, and MultiSelection

The change ends up a bit lighter as code isn't being wrapped in conditionals... in class-based components, it ends up being simple - set an instance variable, clear it out on unmount...
```
 this.someTO = setTimeout(...
 
 componentWillUnmount() {
  if(this.someTO) {
    clearTimeout(this.someTO);
    this.someTO = null;
  }
 }
```

in functional components, it's a bit weirder - but yeah, functional be like that... (no functional in this PR, but for future ref)

```
// store a ref to hold it (functional version of instance variables)
let someTO = useRef(null).current;

// useEffect with empty dependency array is equiv of didMount -> willUnmount
useEffect(() => {
  return () => {
    if (someTO) {
      clearTimeout(someTO);
      someTO = null;
    }
  }
}, [])

someTO = setTimeout(...
```

### DIFF
--- a/lib/MultiSelection/SelectedValuesList.js
+++ b/lib/MultiSelection/SelectedValuesList.js
@@ -21,6 +21,12 @@ class SelectedValuesList extends React.Component {
     valueLabelId: PropTypes.string,
   }
 
+  componentWillUnmount() {
+    if (this.focusTO) {
+      clearTimeout(this.focusTO);
+      this.focusTO = null;
+    }
+  }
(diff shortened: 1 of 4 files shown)


mkuklis (approved):
Nice work @JohnC-80!

PR #1550, opened 2021-05-04
STCOM-836 CommandList should not warn about restricted bindings

zburke:
There is no need for `<CommandList>` to warn about overriding system
bindings. It is not an actionable warning because there is no way to
remove it without changing the binding, but the decision has already
been made not to change the binding.

Refs [STCOM-836](https://issues.folio.org/browse/STCOM-836)


JohnC-80 (approved):

JohnC-80:
@zburke mind dropping in some edits at https://github.com/folio-org/stripes-components/tree/master/lib/Commander#browser-shortcuts ?

PR #1551, opened 2021-05-04
STCOM-837: Adapt translations remove ui-users dependencies

elsenhans:
https://issues.folio.org/browse/STCOM-837

The translation dependencies to ui-users in KeyboardShortcutModal should be removed


rchr (approved):

zburke (approved):
Good catch. Thanks for fixing this!

PR #1556, opened 2021-05-07
STCOM-833 avoid calling setState after unmounting

zburke:
`<Paneset>` could call `setState()` after unmount, causing a memory
leak.

Refs [STCOM-833](https://issues.folio.org/browse/STCOM-833)


mkuklis (approved):

mkuklis:
👏

PR #1558, opened 2021-05-19
Fix example use of Checkbox within a final-form

MikeTaylor:


zburke (commented):
Can you explain what what wrong and how this fixes it?

MikeTaylor:
Well, sure, but it's just restating in prose what the diff shows:
1. The example was duplicated, with two all-but-identical copies.
2. The copies included a `value` prop, which is rarely if ever used.

JohnC-80 (approved):

PR #1561, opened 2021-05-21
STCOM-830 add 'scroll-to-item' logic

JohnC-80:
This PR aims to give modules a way to restore their scroll position of a current MCL given an 'itemToView' prop, as well as an 'onMarkPosition' handler. In the example image, it retains the item's vertical position across switches in the MCL's width.

scrollItemToView

at first, this will only be compatible with non-virtualized lists, but we'll hopefully expand that soon.

### DIFF
--- a/lib/MultiColumnList/stories/service.js
+++ b/lib/MultiColumnList/stories/service.js
@@ -1,21 +1,21 @@
 import faker from 'faker';
 
-export function syncGenerate(n, start, generator) {
+export function syncGenerate(n, start = 0, generator) {
   const res = [];
-  let i = n;
-  while (i) {
+  let i = start;
+  while (i < n) {
     if (generator) {
       res.push(generator());
     } else {
       res.push({
         active: faker.random.boolean(),
         title: faker.random.words(),
-        index: start + res.length,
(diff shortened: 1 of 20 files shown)


mkuklis (approved):
Nice work @JohnC-80!

lib/MultiColumnList/stories/service.js
```diff
@@ -1,21 +1,21 @@
 import faker from 'faker';
 
-export function syncGenerate(n, start, generator) {
+export function syncGenerate(n, start = 0, generator) {
   const res = [];
-  let i = n;
-  while (i) {
+  let i = start;
+  while (i < n) {
```
zburke, line 6: `for (let i = start; i < n; i++) { ... index: i, ... }` is canonical here. Obvs there's nothing wrong with this, but just, ya know, idioms take less mental energy.

zburke (approved):

lib/MultiColumnList/stories/service.js
JohnC-80: Wayne Campbell said it best: "Sh'yeaah!"

PR #1572, opened 2021-06-09
STCOM-842 re-scale cached sizes to fit...

JohnC-80:
## Purpose
This PR resolves issue [STCOM-842](https://issues.folio.org/browse/STCOM-842) where panesets do not fit the window (either they're too small leaving whitespace or the panes are cropped off.) If a user resizes the panes, travels to a different app, resizes the window, then returns to the app where they resized, the old pane dimensions remain which will not match the resized window (*Phew!)

### Approach
a) use the resize container (a full width div) to judge the real width of the window for the top-level paneset.
b) if an initial layout prop is provided, rescale it before applying it to the panes. - sum the widths, compare to the resize container, rescale the values accordingly...

Math. Wooo 🎏

### DIFF
--- a/lib/Paneset/Paneset.js
+++ b/lib/Paneset/Paneset.js
@@ -372,7 +372,6 @@ class Paneset extends React.Component {
         this.updateLayoutCache(widths, changeType);
         this.resizePanes(this.state.panes, widths);
       }
-      // this.resizePanes(this.state.panes, widths);
     });
   }
 
--- a/lib/Paneset/Paneset.js
+++ b/lib/Paneset/Paneset.js
@@ -354,19 +354,13 @@ class Paneset extends React.Component {
 
       if (toApply && sizesCached === -1) {
         widths = toApply;
-        this.updateLayoutCache(widths, changeType);
-        this.resizePanes(this.state.panes, widths);
       } else if (sizesCached !== -1) {
-        widths = this.state.layoutCache[sizesCached];
-        const adjustedWidths = this.resizeCachedSizesToFit(sizesCached);
-        widths = adjustedWidths || widths;
-        this.updateLayoutCache(widths, changeType);
-        this.resizePanes(this.state.panes, widths);
+        widths = this.resizeCachedSizesToFit(sizesCached) || this.state.layoutCache[sizesCached];
       } else {
         widths = this.calcWidths(this.state.panes);
-        this.updateLayoutCache(widths, changeType);
-        this.resizePanes(this.state.panes, widths);
       }
+      this.updateLayoutCache(widths, changeType);
+      this.resizePanes(this.state.panes, widths);
     });
   }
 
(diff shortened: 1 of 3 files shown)


lib/Paneset/Paneset.js
```diff
@@ -311,13 +355,18 @@ class Paneset extends React.Component {
...
+        const adjustedWidths = this.resizeCachedSizesToFit(sizesCached);
+        widths = adjustedWidths || widths;
+        this.updateLayoutCache(widths, changeType);
+        this.resizePanes(this.state.panes, widths);
       } else {
         widths = this.calcWidths(this.state.panes);
         this.updateLayoutCache(widths, changeType);
+        this.resizePanes(this.state.panes, widths);
       }
-      this.resizePanes(this.state.panes, widths);
```
NikitaSedyx, line 320: looks like it can be unchanged

NikitaSedyx: and now `this.updateLayoutCache(widths, changeType);` is called in all if-else blocks

lib/Paneset/Paneset.js
```diff
@@ -311,13 +355,18 @@ class Paneset extends React.Component {
       if (toApply && sizesCached === -1) {
         widths = toApply;
         this.updateLayoutCache(widths, changeType);
+        this.resizePanes(this.state.panes, widths);
       } else if (sizesCached !== -1) {
         widths = this.state.layoutCache[sizesCached];
+        const adjustedWidths = this.resizeCachedSizesToFit(sizesCached);
+        widths = adjustedWidths || widths;
```
NikitaSedyx, line 362: 3 lines could be `widths = this.resizeCachedSizesToFit(sizesCached) || this.state.layoutCache[sizesCached];`

NikitaSedyx (approved):
nice, thank you!

lib/Paneset/Paneset.js
JohnC-80: Thank you! all DRYed up...

lib/Paneset/Paneset.js
```diff
@@ -14,6 +14,14 @@ const defaultProps = {
   initialLayouts: [],
 };
 
+function sumWidths(key, objectArray) {
+  let sum = 0;
+  objectArray.forEach(o => {
+    sum += parseInt(o[key], 10);
```
zburke, line 20: Maybe `reduce` is more idiomatic? Dunno.
```
return objectArray.reduce((a, i) => parseInt(i[key], 10) + a, 0);
```

lib/Paneset/Paneset.js
zburke: Right, so the question is, which is correct?
* the previous impl when it _wasn't_ called given `sizesCached >= 0`
* the current impl when it is _always_ called, regardless of the value of `sizesCached`

PR #1583, opened 2021-06-29
STCOM-849 wrong day of week

JohnC-80:
resolves https://issues.folio.org/browse/STCOM-849

# Problem(s)
a) FOLIO's hard-coded locales are inconsistently formatted in 2-letter (ex: "ru") and 4-letter (ex: "en-SE") versions. Both forms are valid, but the 2nd 2 letters (region) is particularly important for calendar-rendering since it is necessary to derive the first day of the week.
b) MomentJS stores its i18n info in locale configuration files. If momentJS is asked to apply a locale, and it doesn't have that specific static locale file, it defaults to 'en'. If the locale is set incorrectly, calendar days will render as if 'Sunday' was the first day of the week.

## Approach
I've added a static mapping of 'default' common region codes so that the incoming 2-letter locale can be normalized for Calendar's purposes. 'ru' is adjusted to be 'ru-RU' , 'sv' adjusted to be 'sv-SE' etc...
If we ever need a different region from the default, then a new language/region locale will actually need to be added to stripes. ex: we've got 'ru' - but not 'ru-US'.
Using the adjusted locale, we can get the correct first-day-of-the-week.
Taking that first day of week and offsetting the default 'english' weekday rendering by x days takes care of the calendar offset issue.

Would be awesome if browser Intl API had first-day-of-week info, but it doesn't yet... "yet" could happen 10 years from now... but some recent movement is here - https://github.com/tc39/ecma402/issues/580 This progress after another thread from 2015 https://github.com/tc39/ecma402/issues/6 ... so it'll be another 5 years before browsers actually pick it up XD...

## Future work.
It would be nice to jettison local static files like these in favor (maybe) of unicode CLDR via CLDRjs :
- https://github.com/unicode-org/cldr-json
- https://github.com/rxaviers/cldrjs


mkuklis (approved):
Great work @JohnC-80! Thank you for addressing it.

zburke (approved):

PR #1585, opened 2021-07-05
STCOM-854 Expose getLocalFormat Datepicker util

NikitaSedyx:
https://issues.folio.org/browse/STCOM-854

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ export { default as ButtonGroup } from './lib/ButtonGroup';
 export { default as Checkbox } from './lib/Checkbox';
 export { default as CurrencySelect } from './lib/CurrencySelect';
 export { default as CountrySelection } from './lib/CountrySelection';
-export { default as Datepicker, Calendar } from './lib/Datepicker';
+export { default as Datepicker, Calendar, getLocalFormat } from './lib/Datepicker';
 export { default as DateRangeWrapper } from './lib/DateRangeWrapper';
 export { default as FormattedDate } from './lib/FormattedDate';
 export { default as FormattedTime } from './lib/FormattedTime';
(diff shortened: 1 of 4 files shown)


AndreiShumski2 (approved):

uladislausamets (approved):

index.js
```diff
@@ -6,7 +6,7 @@ export { default as ButtonGroup } from './lib/ButtonGroup';
 export { default as Checkbox } from './lib/Checkbox';
 export { default as CurrencySelect } from './lib/CurrencySelect';
 export { default as CountrySelection } from './lib/CountrySelection';
-export { default as Datepicker, Calendar } from './lib/Datepicker';
+export { default as Datepicker, Calendar, getLocalFormat } from './lib/Datepicker';
```
axelhunn, line 9:
```suggestion
export { default as Datepicker, Calendar, getLocalFormat as getLocaleDateTimeFormat } from './lib/Datepicker';
```

zburke: At present, this is used internally in only one place even though it is exported from `Datepicker.js`. If you think `getLocaleDateTimeFormat` is a more apt name, I think it would be better to rename the function itself and export that, rather than just renaming the export. 

Having a different name internally and externally feels like something we should avoid if possible.

axelhunn: It's just the name `getLocalFormat` doesn't tell format of what do we get - in general. Although it totally makes sense inside the Datepicker context as it's the component about dates.

JohnC-80 (approved):
Ok guys... if this internal method is working well enough for you all, we can expose it.. 
I've always considered it a fallback over the shortcomings of moment (growing lots of those lately) more than the go-to for everyone needing this sort of functionality.
Once we analyze migration from moment, hopefully something better will come along... We'll be sure to announce it in stripes-updates if a deprecation does happen. (As well as any materialized moment migration plans)

index.js
JohnC-80: As you mention, @axelhunn it's really made for Datepicker - with Datepicker's particular context in mind. While we don't need much variety for formats within the component library, the app level would be different. I would recommend that any one need it just to fork it and tailor the output to your applications needs - but if importing/using this saves boilerplate - and there isn't any doubt we want to spend our time in other places besides figuring out date-related issues - then okay, we can export it.

PR #1587, opened 2021-07-08
Add hideSource prop to MetaSection

EthanFreestone:
Refs ERM-1230, STCOM-859

MetaSection is currently designed so that if user data is provided, it will display that, and if not then it will display "automated process" as a source. For certain applications, such as the mainly German use of ERM apps, user data is not built into the domain model at this level, and so is not attached to lastUpdated/dateCreated.

A lack of user data here implying "AutomatedSource" is not correct, and in ERM at least the desired outcome is to simply be able to hide these lines.

An alternative use case would be the ability to see the user metadata dependent only on permissions.

### DIFF
--- a/lib/MetaSection/MetaSection.js
+++ b/lib/MetaSection/MetaSection.js
@@ -25,6 +25,7 @@ const propTypes = {
   ]),
   createdDate: PropTypes.string,
   headingLevel: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
+  hideSource: PropTypes.bool,
   id: PropTypes.string,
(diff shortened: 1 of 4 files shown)


zburke (approved):
This looks good to me. Before you merge, please either move ERM-1230 into STCOM, or clone/create a new STCOM story for this work and then add the Jira slug (STCOM-NNN) to the commit-subject so we can keep track within Jira of which stripes-components release has this new feature. Thanks!

EthanFreestone:
Thanks @zburke , [STCOM-859]( https://issues.folio.org/browse/STCOM-859) created and currently assigned to me. Should I push it to CR or just close once merged?

PR #1589, opened 2021-07-11
STCOM-860 always use Arabic numerals given backendDateStandard

zburke:
The implied promise of `outputFormatter`, given `backendDateStandard` is
that it will format dates using Arabic numerals (0-9) as this is what
the backend expects. Previously, it handled only for formatting but not
the conversion to Arabic ("Latn") numerals if the current locale used
non-Arabic numerals, which is the case for manu Arabic locales that use
Arabic-Hindi numerals.

Now, `<Datepicker>` delivers on both its promises:

1. it formats the date as expected for the backend
2. it provieds dates in Arabic numerals as expected for the backend

Refs [STCOM-860](https://issues.folio.org/browse/STCOM-860), [FOLIO-3208](https://issues.folio.org/browse/FOLIO-3208)

### DIFF
--- a/lib/Datepicker/tests/Datepicker-test.js
+++ b/lib/Datepicker/tests/Datepicker-test.js
@@ -548,7 +548,7 @@ describe('Datepicker', () => {
     });
   });
 
-  describe('defaultOutputFormatter converts to RFC2822 output', () => {
+  describe('defaultOutputFormatter provides RFC-2822 output', () => {
     const frm = moment().locale('en');
 
     let output;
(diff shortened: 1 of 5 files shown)


zburke:
The only easy way I can think of to add tests for this is to export `defaultOutputFormatter` so I can feed it some non-Latn input and validate the Latn output, but that feels lame. Merely setting the `locale` prop to `ar` does not have the desired effect of turning the calendar-popup to Arabic :/ Better ideas?

mkuklis (approved):
Thank you for the extra documentation @zburke! 💯 . I think exporting `defaultOutputFormatter` actually sounds good.

JohnC-80 (approved):
Good work man... a much shorter way around than we were thinking!
Yeah, exporting for testing sake's not a bad idea.

zburke:
`describe.only(...)` cool like, 
```
START:
  Datepicker
    defaultOutputFormatter converts non-Latn input to Latn-output
      ✔ returns a Latn (Arabic, 0-9) string

Finished in 0.009 secs / 0.001 secs @ 13:35:49 GMT-0400 (Eastern Daylight Time)

SUMMARY:
✔ 1 test completed
```
but `describe(...)` all like, 
```
FAILED TESTS:
  Datepicker
    defaultOutputFormatter converts non-Latn input to Latn-output
      ✖ returns a Latn (Arabic, 0-9) string
        Chrome 91.0.4472.114 (Mac OS 10.15.7)
      AssertionError: expected 'Invalid date' to equal '2021-07-11'

      + expected - actual

      -Invalid date
      +2021-07-11
```
__sigh__

zburke:
I'm giving up on SC coverage as the lines it reports as uncovered clearly are.

lib/Datepicker/tests/Datepicker-test.js
```diff
@@ -496,4 +501,110 @@ describe('Datepicker', () => {
       expect(format).to.equal('DD.MM.YYYY');
     });
   });
+
+  // this test runs successfully in a `describe.only(...)` block
+  // but fails when run as part of the full suite.
+  // grrrrrrr.
+  // the same test is below with a Latn locale (fr) instead of
+  // an Arab one (ar) since that seems to be the deciding factor.
+  describe.skip('defaultOutputFormatter converts non-Latn input to Latn-output', () => {
```
ViktorSoroka07, line 510: It does not work for with `describe.only`.
```
      ✖ returns a Latn (Arabic, 0-9) string
        Chrome 92.0.4515.159 (Mac OS 10.15.7)
      AssertionError: expected 'Invalid date' to equal '2021-07-11'

      + expected - actual

      -Invalid date
      +2021-07-11

```

PR #1593, opened 2021-07-16
STCOM-848 Set max-height on dropdown menu

JohnC-80:
# Problem
Dropdowns with lots of items/options can leave items unreachable since they cannot scroll.

## Approach
Set a maxheight on DropdownMenu so that it won't be pushed beyond the height of the viewport.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * `<Datepicker>` must correctly handle RFC-2822 dates. Refs STCOM-861.
 * `<Datepicker>` always provides Arabic numerals (0-9) given `backendDateStandard` to format values. Refs STCOM-860.
 * `<SingleSelect>` add new `loading` and `loadingMessage` props to display while loading options. Refs STCOM-858.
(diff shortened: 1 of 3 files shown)


zburke (approved):
I'm not super stoked about just hard-coding this to a smallish value since it could still break on somebody itty bitty teeny weeny yellow polkadot smartphone, and no doubt somebody on a monster-sized display will complain that it's scrolling when it doesn't need to do. 

But as far as being a quick and effective fix for 95% of use cases, this has my full support.

JohnC-80:
Yeah, this is a max-height of 80% relative to the viewport, so it should work fine for smaller displays... I considered something more dynamic where it would size itself accordingly depending on where it was on screen, but by that point Popper's taken over and shifted it to avoid any overflow.  `overflow: auto` should cover the rest for the scrolling aspect... we'll see how it goes.

JohnC-80:
Case-by-case, they could just use an `overrideStyle` if it ends up being too finicky.

PR #1609, opened 2021-08-30
Clear timeout before showing the tooltip to avoid setting state on un…

aditya-matukumalli:
…mounted component.


zburke (approved):
Thank you! ❤️ I love it when people help clean up console warnings!

JohnC-80 (approved):

PR #1618, opened 2021-09-24
STCOM-877: Update react-overlays

mkuklis:
https://issues.folio.org/browse/STCOM-877

The previous version of `react-overlays` was not working correctly with react 17. Root close would fire right after opening a dialog / modal which would close it.

This was fixed in: 

https://github.com/react-bootstrap/react-overlays/pull/880

More discussion with @aditya-matukumalli here:

https://folio-project.slack.com/archives/C210UCHQ9/p1632347078155500


aditya-matukumalli (approved):

zburke (approved):
LGTM!

PR #1628, opened 2021-10-08
STCOM-880 focus-related tests

JohnC-80:
React changed its behavior for focus/blur events, so simulating these events programmatically has changed...
`focus()` still works, but `blur()` does not... as opposed to `blur` a `focusout` event must be dispatched.... 

see https://github.com/facebook/react/issues/19606#issuecomment-673629090 for the basic details/solution....

will need https://github.com/folio-org/stripes-testing/pull/124

### DIFF
--- a/.github/workflows/build-npm-release.yml
+++ b/.github/workflows/build-npm-release.yml
@@ -20,7 +20,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  build-npm-release:
     if : ${{ startsWith(github.ref, 'refs/tags/v') }}
     env:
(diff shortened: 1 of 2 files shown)


lib/Popover/tests/Popover-test.js
```diff
@@ -3,17 +3,25 @@
...
 import { describe, beforeEach, it } from '@bigtest/mocha';
 import { expect } from 'chai';
 
 import { mount } from '../../../tests/helpers';
 
 import Button from '../../Button';
-import ButtonInteractor from '../../Button/tests/interactor';
 
 import Popover from '../Popover';
-import PopoverInteractor from './interactor';
```
zburke, line 16: Can we remove `./interactor.js` then? Or should we instead replace it with lines 16-24 below? It feels wrong to have both.

PR #1644, opened 2021-11-02
Add Datepicker dateFormat prop for tests.

JohnC-80:
Adding a dateFormat for tests so that they'll interpret the value correctly rather than defaulting the calendar to 'this month' since we test calendar/weekday alignment with a particular day.


zburke:
Fixes [STCOM-900](https://issues.folio.org/browse/STCOM-900)

PR #1645, opened 2021-11-02
STCOM-884 Fix RadioButton markup/styles for proper announcement in JAWS AT

JohnC-80:
JAWS on Chrome improperly announces the order of rendered `<RadioButtons>` as '1 of 1' regardless of which radio button of a set you're focused.  This was fixed with a mark-up change (moving the input outside of the label.)

With this fix in place, the radio buttons are now announced properly: e.g.: '1 of 3' or '2 of 5'.


aditya-matukumalli (approved):

zburke:
It is unfortunate that we merged this objectively-worse markup, moving the `<input>` outside the `<label>`, due to what feels like a bug in JAWS. This may feel like a long shot, @JohnC-80, but did we consider filing an issue with JAWS? [STCOM-910](https://issues.folio.org/browse/STCOM-910)/#1688 directly follows from this change 🙁. 

To be clear, I'm not arguing against this fix, but our own code is clearly getting more complicated to deal with this glitch, and we if we could get JAWS to just fix the glitch in the first place, well, that'd be awfully nice.

PR #1646, opened 2021-11-03
STCOM-895: Update focus for MultiSelection component.

Vladyslav-Velytskyi:
# STCOM-895 eholdings - fix multi-select component

## Issue
[STCOM-895](https://issues.folio.org/browse/STCOM-895)

## Description
This bug occurs in the `Accordion` component. When the content of the `Accordion` is in focus, then the content inside it gets the `z-index` property equal to 2, and all the contents of the other accordions get the `z-index` 1. When the focus is in another accordion and the control button is pressed in `MultiSelection`, the focus is not hung on it because of what an accordion with a `MultiSelection` still has `z-index: 1`. Adding a bigger `z-index` value for a `MultiSelection` does not solve this situation due to the different nesting of the components.

#### Update
`event.stopPropagation()` blocks handlers in surrounding `<Accordion>` component.

#### Update
Vlad has suggested that we handle the focus in `<Accordion>` component instead of `<MultiSelection>`. So I used existing `useClickOutside` hook to check if click originated from inside accordion's content and if so - then we can set internal `focused` state to `true` which will set correct `z-index`

## Screenshots
Before
bfOYDiX9lC

After
QkdC9jCbnQ


BogdanDenis (approved):
Good catch!

mariiaschv (approved):

JohnC-80 (approved):
Looks good. Great job solving this!

zburke (approved):

PR #1649, opened 2021-11-05
lint

zburke:
lint complains about the `aria-invalid` setting here, though it didn't
used to and this part of the component hasn't changed in years. That's
puzzling.
```
/Users/zburke/projects/folio-org/stripes-components/lib/RadioButton/RadioButton.js
  165:11  error  The attribute aria-invalid is not supported by the role radio. This role is implicit on the element input  jsx-a11y/role-supports-aria-props

/Users/zburke/projects/folio-org/stripes-components/lib/Selection/SingleSelect.js
  692:11  error  The attribute aria-invalid is not supported by the role button. This role is implicit on the element button  jsx-a11y/role-supports-aria-props
```

### DIFF
--- a/lib/RadioButton/RadioButton.js
+++ b/lib/RadioButton/RadioButton.js
@@ -178,7 +178,6 @@ class RadioButton extends React.Component {
             value={value}
             readOnly={readOnly}
             required={required}
-            aria-invalid={!!(this.props.error)}
           />
           <span className={css.labelPseudo} />
           {(label || readOnly) &&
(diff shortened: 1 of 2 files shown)


mkuklis (approved):

JohnC-80 (approved):
I saw this on the `<radiobuttons>` but not on selection in #1645. Thanks!

zburke:
@JohnC-80, do you understand why we're seeing this all of a sudden, e.g. do we have a newer version of lint rules/a11y rules, that are giving us more accurate information? I'd like to understand why this prop is here in the first place, and why we're getting new warnings about it, before removing it. You know, [Chesterton's Fence](https://fs.blog/chestertons-fence/) and all.

JohnC-80:
That's right... it seems an inevitable part of the process for lint upgrades to introduce new rules/new facets of rules and uncover new issues.

zburke:
Lint is no longer complaining about this, so I'm closing it.

PR #1652, opened 2021-11-09
STCOM-820: Add an event handler for accordion opening/closing at users' direct request.

BogdanDenis:
## Description
Add new prop - `onClickToggle`.
This property is a callback function that is called when a user has opened accordion by interacting with it (clicking on header, pressing Enter). It's not called when accordion was toggled via "Expand all" button or any other way.

## Issues
[STCOM-820](https://issues.folio.org/browse/STCOM-820)


JohnC-80 (approved):
Looks good! Thanks!

Vladyslav-Velytskyi (approved):

mariiaschv (approved):

PR #1654, opened 2021-11-09
STCOM-893 text alignment of accordion headers.

JohnC-80:
Added text-align property to label area of accordion headers.
 
 resolves [STCOM-893](https://issues.folio.org/browse/STCOM-893)
 
 
image


zburke (approved):
Thank you!

PR #1661, opened 2021-11-15
STCOM-883 - axe reports nested interactive elements...

JohnC-80:
## Problem

Axe testing within UI modules will report nested interactive elements in numerous places within the scope of stripes-components. This particularly affects icons, iconButtons and accordions - components that ui-modules use FREQUENTLY.

Interactive elements within eachother create an awkward situation for assistive technology and general user experience. If an inner element is clicked, does that click bubble outward? Does the inner element even register to a screen reader? The outer? Nested interactives can create lots of issues.

For the sake of education, any element with a `tab-index` attribute is considered an interactive element. Even if it's removed from the tab order via `tabindex="-1"`, it's still considered interactive. So despite not being directly focusable, Icons end up rendering as interactable elements. `tabindex="-1"` removes an element from tab order, but sets it up so that **it can be programmatically focused**. The intent of it being focused programmatically is key - that is the only circumstance where `tabindex ="-1"` should be used.

### Approach
- remove default `tabindex=-1` from icon, 
- pass arialabelledby correctly in IconButton

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Add new `textLeft` className to `<Layout>` component. Refs STCOM-897.
 * Check that content inside `<Accordion>` was clicked and set focus flag. Refs STCOM-895.
 * Add an event handler for accordion opening/closing at users' direct request. Refs STCOM-820.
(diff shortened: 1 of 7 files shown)


MikeTaylor (approved):
This is welcome.

zburke (approved):

aditya-matukumalli (approved):

PR #1665, opened 2021-11-17
STSCOM-895: Make useClickOutside catch click event on capture phase

BogdanDenis:
## Description
Previous fix didn't cover all cases https://github.com/folio-org/stripes-components/pull/1646
A breakdown of what is happening when adding a tag to a record:
1. Click on `<MultiSelection>` search field. This opens up a dropdown with a few existing tags.
2. Either search for a tag that exists but is not visible.
3. Click on the tag in dropdown. At this point `onClick` that is added to the `<MultiSelection>` option is executed and the option disappears (removed from DOM)
4. `useClickOutside` click handler is executed. `ref.current.contains(e.target)` returns false because `e.target` is not in DOM even though that element was inside accordion's content.

We can fix this by handling `useClickOutside` click event in it's capture phase by passing `true` as third argument to `addEventListener`. This way this handler will be executed before all other click handlers.

## Screenshots

https://user-images.githubusercontent.com/19309423/142250381-ee40d3b9-d9ba-48df-97c6-8d2b1940406a.mp4

## Issues
[STCOM-895](https://issues.folio.org/browse/STCOM-895)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Add new `textLeft` className to `<Layout>` component. Refs STCOM-897.
 * Check that content inside `<Accordion>` was clicked and set focus flag. Refs STCOM-895.
 * Add an event handler for accordion opening/closing at users' direct request. Refs STCOM-820.
+* Make `useClickOutside` click handler work on `capture` event phase. Refs STCOM-895.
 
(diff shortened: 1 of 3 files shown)


zburke (approved):
Wow, impressive work! Thank you for the clear description of the problem and how this change addresses it.

Vladyslav-Velytskyi (approved):

mariiaschv (approved):

PR #1666, opened 2021-11-18
STCOM-889: Create a Conflict Detection banner

mariiaschv:
## Description
Created a Conflict Detection banner which will be displayed when a conflict due to optimistic locking occurs.

Issue: [STCOM-889](https://issues.folio.org/browse/STCOM-889)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Add an event handler for accordion opening/closing at users' direct request. Refs STCOM-820.
 * Remove default tabIndex from Icon (cause of nested interactive axe errors), treated aria-labelledby appropriately in IconButton. Fixes STCOM-883
 * Make `useClickOutside` click handler work on `capture` event phase. Refs STCOM-895.
+* Create a Conflict Detection banner. Refs STCOM-889.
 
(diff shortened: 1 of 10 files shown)


zburke (changes requested):
This is a great start, but at minimum it needs tests and a readme. Please add a link to `/README.md` as well.

zburke:
From the ticket:

> Given the user views the Conflict detection banner
> When s/he clicks on the Cancel button OR "X"
> Then return to the previous page

does this mean we should be passing `dismissible` to the internal `<MessageBanner>`?

mariiaschv:
> From the ticket:
> 
> > Given the user views the Conflict detection banner
> > When s/he clicks on the Cancel button OR "X"
> > Then return to the previous page
> 
> does this mean we should be passing `dismissible` to the internal `<MessageBanner>`?

Hey, I believe this means that if user is on Edit page and a conflict due to optimistic locking occurs so that the banner appears, if the user wants to leave the Edit page by clicking Cancel button or "X" he should be able to do that without any impediments, just going back to the previous page.

BogdanDenis (approved):

Vladyslav-Velytskyi (approved):

PR #1669, opened 2021-11-19
Use version for stripes-testing dependency.

JohnC-80:
In a previous PR, was replaced with a github branch... this PR restores.


zburke (approved):
Good catch; thank you!

PR #1671, opened 2021-11-20
STCOM-902 use correct expanded-accordion selector

zburke:
How did this ever work? There is no `content` class in the imported CSS,
only `content-wrap` and `content-region`. What the ...?

Refs [STCOM-902](https://issues.folio.org/browse/STCOM-902)


JohnC-80 (approved):
Hmm... possibly the scoped CSS just had 'content' as the key for the scoped className. 🤷‍♂️

PR #1675, opened 2021-11-30
STSMACOM-611/STCOM-901 receiving fewer results will have MCL scroll to top

JohnC-80:
## Problem
MCL doesn't scroll to the top-most item when result count is decreased. 
Reproducable on folio-testing in ui-users:
Go to ui-uerss https://folio-snapshot.dev.folio.org/users?sort=name
Check the filter for 'Active'
Scroll all the way to the bottom of the results....
Enter 'System' in the search field...
The list will have few results than can fill the frame, but it will not be scrolled completely to the top so that all are visible.

### Approach
Simple change in the update cycle of MCL - less data should also mean scrolling to top (as does sorting the list)


zburke (approved):
Thanks! I had a feeling this was gonna be, "chalk X: $0.10; knowing where to put it: $49,999.90". I'm glad it turned out that way.

PR #1694, opened 2021-12-20
STCOM-905 - remove @bigtest/mocha dependency...

JohnC-80:
@bigtest/mocha has been long deprecated.
Up-to-date mocha (^9.0.0) allows promises to be returned in `it` functions.

Refs [STCOM-905](https://issues.folio.org/browse/STCOM-905)


zburke (approved):
@JohnC-80, it looks like this is 95% "import mocha directly" and 5% refactors of the tests for `MessageBanner` and `Paneset` resizing. There's nothing really mocha-related there, but these just came along for the ride. 

Is that right? Glancing through changes in 62 files 😅  I am just trying to make sure I didn't miss something.

PR #1696, opened 2021-12-27
STCOM-912 restore props spread to avoid console warning

zburke:
Some prop-spreading was removed in an earlier refactor (#1663) that now
causes a console warning because the prop is now applied directly to the
underlying DOM element. Restore the spread.

Refs [STCOM-912](https://issues.folio.org/browse/STCOM-912) and [STCOM-883](https://issues.folio.org/browse/STCOM-883) by extension.

### DIFF
--- a/lib/MultiSelection/OptionsListWrapper.js
+++ b/lib/MultiSelection/OptionsListWrapper.js
@@ -8,11 +8,12 @@ const OptionsListWrapper = React.forwardRef(({
   controlRef,
   isOpen,
   renderToOverlay,
+  menuPropGetter,
   modifiers,
   ...props
 }, ref) => {
   if (useLegacy) {
-    return <div data-open={props.isOpen} {...props} hidden={!isOpen}>{children}</div>;
+    return <div {...menuPropGetter({ ref })} data-open={props.isOpen} {...props} hidden={!isOpen}>{children}</div>;
   }
 
   let portal;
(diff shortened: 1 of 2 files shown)


JohnC-80:
Hmm... instead of re-applying the spread, the passed-in `menuPropGetter` prop should be removed from the instance of OptionsListWrapper that's rendered here:  here: https://github.com/folio-org/stripes-components/blob/STCOM-912/lib/MultiSelection/MultiSelectOptionsList.js#L146

as part of the refactor, the props from that menu needed to be applied to the `<ul>` that wrapped the list options here: https://github.com/folio-org/stripes-components/blob/STCOM-912/lib/MultiSelection/MultiSelectOptionsList.js#L164

So if the spread is in place, the props that come through from the menuPropGetter get applied in two different places - which we don't want...

lib/MultiSelection/OptionsListWrapper.js
```diff
@@ -8,11 +8,12 @@ const OptionsListWrapper = React.forwardRef(({
   controlRef,
   isOpen,
   renderToOverlay,
+  menuPropGetter,
   modifiers,
   ...props
 }, ref) => {
   if (useLegacy) {
-    return <div data-open={props.isOpen} {...props} hidden={!isOpen}>{children}</div>;
+    return <div {...menuPropGetter({ ref })} data-open={props.isOpen} {...props} hidden={!isOpen}>{children}</div>;
```
JohnC-80, line 16: Let's remove the supplied prop instead of re-applying the spread of that prop... it's odd that this part of the jsx would require it, but the latter element that's rendered does not... anyways, see my earlier comment on the PR...

JohnC-80 (changes requested):

JohnC-80 (approved):

PR #1697, opened 2021-12-31
STCOM-914 correct LayoutHeader documentation

zburke:
Specify the correct shape of the `actions` array's objects.

Refs [STCOM-914](https://issues.folio.org/browse/STCOM-914)

### DIFF
--- a/lib/LayoutHeader/readme.md
+++ b/lib/LayoutHeader/readme.md
@@ -16,7 +16,7 @@ Name | type | description | default | required
 --- | --- | --- | --- | ---
 title | string | text for the header of the section | |
 level | number | The header level of the rendered `<Hx>` tag. (1-6). | 3 |
-actions | array of objects | see examples | |
+actions | array of `{ name, icon, handler }` | see examples | |
 onDelete | function | to support delete actions in the simplest use case. | |
 noActions | bool | hides action buttons. Useful for 'read-only' mode. | |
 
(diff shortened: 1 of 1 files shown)


zburke:
@ryandberger, I noticed this when running tests locally for https://github.com/folio-org/stripes-smart-components/pull/1163 and finding the AddressFeildGroup test failed. The code there is ancient so I'm not sure how tests ever passed 😬 .

JohnC-80 (approved):

zburke:
Haha, I read the [prop-types](https://github.com/folio-org/stripes-components/blob/df6ec972ab510a0716739e959e83e439534afe84/lib/LayoutHeader/LayoutHeader.js#L11-L15) but not the function itself and guess what? They don't agree. The [internal actions object](https://github.com/folio-org/stripes-components/blob/df6ec972ab510a0716739e959e83e439534afe84/lib/LayoutHeader/LayoutHeader.js#L43-L45) uses `title` instead of `name`, as does the [button's content](https://github.com/folio-org/stripes-components/blob/df6ec972ab510a0716739e959e83e439534afe84/lib/LayoutHeader/LayoutHeader.js#L61). The only place `name` is _actually_ used is as a [key](https://github.com/folio-org/stripes-components/blob/df6ec972ab510a0716739e959e83e439534afe84/lib/LayoutHeader/LayoutHeader.js#L54). 

Every place this is actually in use passes `title` instead of `name`, so I think the easiest/rightest thing to do is update the doc _and_ the code to both use `title` and eliminate `name` altogether. Agreed, @JohnC-80?

PR #1705, opened 2022-01-11
STCOM-898: Layout > Add textStart, textEnd properties

mariiaschv:
## Purpose
Adds new `text-start`, `text-end` classNames to `<Layout>` component to support both LTR and RTL languages.

## Issues
[STCOM-898](https://issues.folio.org/browse/STCOM-898)

### DIFF
--- a/lib/Layout/stories/classNamePropertiesMapping.js
+++ b/lib/Layout/stories/classNamePropertiesMapping.js
@@ -41,5 +41,7 @@ export default {
   'right': 'float: right;',
   'textLeft': 'text-align: left;',
   'textCentered': 'text-align: center;',
-  'textRight': 'text-align: right;'
+  'textRight': 'text-align: right;',
+  'text-start': 'text-align: start;',
+  'text-end': 'text-align: end;'
 };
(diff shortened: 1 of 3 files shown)


zburke (approved):
LGTM, but I am not an RTL expert. @JohnC-80, what do you think?

Svitlana-Korol (approved):

JohnC-80 (approved):
Approved - Thanks!!

lib/Layout/stories/classNamePropertiesMapping.js
```diff
@@ -41,5 +41,7 @@ export default {
   'right': 'float: right;',
   'textLeft': 'text-align: left;',
   'textCentered': 'text-align: center;',
-  'textRight': 'text-align: right;'
+  'textRight': 'text-align: right;',
+  'text-start': 'text-align: start;',
```
BogdanDenis, line 45: I think it's good to be consistent with names and use `textStart` and `textEnd` here

BogdanDenis (changes requested):

lib/Layout/stories/classNamePropertiesMapping.js
mariiaschv: I agree with you about consistency but the reason I used `text-start` and `text-end` is [this comment](https://issues.folio.org/browse/STCOM-897?focusedCommentId=116278&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-116278) by @zburke where he said that 

> Kinda ugly that we have multiple naming conventions in place there, but I think we are leaning toward dash-separated-names now

But if you still think the names should be changed to camelCase I will do it, no problem :)

BogdanDenis: @mariiaschv oh I didn't notice that comment... then it's ok to merge as is

BogdanDenis (approved):

PR #1707, opened 2022-01-21
STCOM-850 Standalone Calendar

JohnC-80:
Calendar needed a couple of pushes to help it be better suited for a potential variety of use cases.


zburke (approved):
readme ... stories ... swoon!

PR #1709, opened 2022-01-25
STCOM-923 Add the ability to hide the number of pages and leave the buttons not disabled

vashjs:
**Purpose:**

For tasks 
https://issues.folio.org/browse/UIIN-1890
https://issues.folio.org/browse/UIIN-1897

we need to be able to navigate not by offsets + indexes, but by making a new request to the server, while the next and prev buttons should be active. Also we should  add the ability to hide the the label with number of pages and propagate the direction to parent component.


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -519,7 +521,7 @@ class MCLRenderer extends React.Component {
       } else {
         newState.firstIndex = parseInt(Object.keys(contentData)[0], 10) || 0;
       }
-    } else if (pagingType === 'click' || pagingType === 'prev-next') {
+    } else if (pagingType === pagingTypes.LOAD_MORE || pagingType === pagingTypes.SCROLL) {
```
JohnC-80, line 524:
```suggestion
    } else if (pagingType === pagingTypes.LOAD_MORE || pagingType === pagingTypes.PREV_NEXT) {
```

JohnC-80 (changes requested):

lib/MultiColumnList/MCLRenderer.js
vashjs: Thanks, my bad

UladzislauKutarkin (approved):

JohnC-80 (approved):

PR #1712, opened 2022-01-31
STCOM-926 Settings : Move focus to second pane

vashjs:
https://issues.folio.org/browse/STSMACOM-617

## Problem
In the current implementation, when you click on the link in the "settings" panel, the focus does not switch to the newly opened one. 

current

## Solution
- Add an focus to first focusable element in Pane

fixed

### DIFF
--- a/lib/Pane/Pane.js
+++ b/lib/Pane/Pane.js
@@ -6,6 +6,7 @@ import css from './Pane.css';
 import PaneHeader from '../PaneHeader';
 import { withPaneset } from '../Paneset/PanesetContext';
 import { withResize } from '../Paneset/ResizeContext';
+import { getFirstFocusable } from '../../util/getFocusableElements';
 
 const propTypes = {
   actionMenu: PropTypes.func,
@@ -94,6 +95,10 @@ class Pane extends React.Component {
 
   componentDidMount() {
     const { transition, defaultWidth, paneset, resizer } = this.props;
+
+    const el = getFirstFocusable(this.el.current);
+    el.focus();
+
     const paneObject = {
       transition,
       setStyle: this.setStyle,
--- a/lib/Pane/Pane.js
+++ b/lib/Pane/Pane.js
@@ -95,9 +95,7 @@ class Pane extends React.Component {
 
   componentDidMount() {
     const { transition, defaultWidth, paneset, resizer } = this.props;
-
     const el = getFirstFocusable(this.el.current);
-    el.focus();
 
     const paneObject = {
       transition,
@@ -121,6 +119,8 @@ class Pane extends React.Component {
       this.setState({ contentMinWidth: this.getContentWidth() }); // eslint-disable-line react/no-unused-state
     });
     this._isMounted = true;
+
+    if (el) el.focus();
   }
 
   componentWillUnmount() {


NikitaSedyx:
1. don't think it should be done in `Pane`, IMO it should be hook that can be applied when required
2. as I understand it works only with the second pane when the third should be focused as well - to make user navigation more accessible

vashjs:
> 1. don't think it should be done in `Pane`, IMO it should be hook that can be applied when required
> 2. as I understand it works only with the second pane when the third should be focused as well - to make user navigation more accessible

Hi Nikita, thanks for watching this PR

1) I feel like we need this every time we open a new route. Often, the new page is wrapped in a Pane, which I found to be a good place to add it.
2) It will work for all levels in settings, because all of them wrapped with Pane.

pane

NikitaSedyx:
yep, I see, looked at util method etc

have you tested MCLs? probably it can introduce unexpected behaviour

vashjs:
> yep, I see, looked at util method etc
> 
> have you tested MCLs? probably it can introduce unexpected behaviour

I've checked ui-inventory MCL, and it works fine. I will double check other modules, for more confidence

NikitaSedyx:
what I see in orders module:
1. Actions button is focused right after rendering (2 panes)
2. cross button is focused after render (3 panes)
3. item selection changes focus to 3rd pane close button

I'm not sure that all of these points are expected

JohnC-80 (commented):
While it's a common case to need a transition of focus here, to the first item of content, there's always an exception, so we have to provide the space for the module to decide when focus should move, where it should go, and under what circumstances. If it were always executed when mounting a pane, how does it behave if a direct link to a settings page is followed and multiple panes mount at *almost the same time?

`GetNextFocusable` is exposed from stripes-components for handling this kind of thing as long as you've got a ref to an element at the module level. It might be more straightforward API for `<Pane>` to expose a function that would pass the pane's container ref during its `componentDidMount` lifeCycle . `onMount`.

NikitaSedyx:
@JohnC-80 

> how does it behave if a direct link to a settings page is followed and multiple panes mount at *almost the same time

the last pane should be focused because of render chain

-------
At the same time, the main benefit of Pane extension is avoiding changes in all places. When we provide a callback to manage focusable element we still need to update tone places to avoid unexpected behaviour like I describe before - if such change will go to Lotus it should be done in all places for Lotus.

Additionally few teams already implemented focusing in their modules (kind of copy-paste) - I was implementing hooks before found this request.

IMO it makes sense to discuss pros/cons of updating Pane and utils/hooks that can be used and controlled by teams

JohnC-80:
Of course, render chain... but it that isn't always the expected outcome - and it's the same issue if autofocus is used.

Focus management is very tricky. I'm curious to see what a hooks for handling would look like, Nikita. I've got some ideas in mind. I would be onboard with a global solution that could work with all use-cases - I think it's possible if developers can set up the conditionals for focus on a per-element basis. The case we have in front of us is the 3-pane typical flow - Search -> results -> detail -> edit -> detail -> results... 

I think having a callback like this is generally good for pane, and can be tied to other functionality (besides focus management.)

NikitaSedyx:
@JohnC-80 it's super trivial hook that does the same what @vashjs did in this PR, but with concern that we can use it or not. Effort with both approaches is the same: a) we need to apply hook where we need vs b) manage focusable element in Pane props

```
const usePaneFocus = () => {
  const paneTitleRef = useRef();

  useLayoutEffect(() => {
    if (paneTitleRef.current) {
      paneTitleRef.current.focus();
    }
  }, []);

  return {
    paneTitleRef,
  };
};
```

I've seen someone did the same, but with copy/paste in each place.

> render chain... but it that isn't always the expected outcome - and it's the same issue if autofocus is used

I think autofocus is a bit different here, because useLayoutEffect will be called exactly after render. So if we have Pane1 -Pane2 - Pane3, hooks will be called 3 times from 1 to 3 in order panes are rendered


> it's possible if developers can set up the conditionals for focus on a per-element basis

not sure I've got your point, but in settings it's hard to build any conditions and actually access elements on different levels because of smart-components usage

vashjs:
Closed because of creating new PRs:
stripes-smart-components - https://github.com/folio-org/stripes-smart-components/pull/1179
stripes-compponents - https://github.com/folio-org/stripes-components/pull/1717

PR #1716, opened 2022-02-02
STCOM-929 NavListSection should accept a tag prop

OleksandrHladchenko1:
HI, @zburke! In scope of my task I need to replace DIV for General, Fee/Fine, Patron blocks with H3 tag. So, in NavListSection I added tag prop to Headline component. What do you think about this change? Can we apply it for my task? 
Ref: https://issues.folio.org/browse/UIU-2037

### DIFF
--- a/lib/NavListSection/NavListSection.js
+++ b/lib/NavListSection/NavListSection.js
@@ -10,6 +10,7 @@ const NavListSection = forwardRef(({
   children,
   className,
   label,
+  tag,
   striped,
   ...rest
 }, ref) => (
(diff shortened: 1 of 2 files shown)


JohnC-80 (changes requested):
It's probably best to expose this as a prop to be used at the module level rather than it default to h3's... this way the level can be adjusted per-use. While it's best to have heading elements rather than divs, it can also be problematic if the incorrect heading level is used.

zburke (changes requested):
Please fix the lint error by configuring prop-types, then this looks good. Please note I cloned UIU-2037 to STCOM-929 since this change will be in stripes-components rather than ui-users.

JohnC-80 (approved):

zburke (approved):

PR #1719, opened 2022-02-08
STCOM-929 NavListSection should accept a tag prop

zburke:
In the scope of fixing some a11y issues in [UIU-2037](https://issues.folio.org/browse/UIU-2037), we need to replace a `<div>` for General, Fee/Fine, Patron blocks with an `<h3>`. So, in NavListSection I added tag prop to Headline component. 

Ref: https://issues.folio.org/browse/UIU-2037
Ref: https://issues.folio.org/browse/STCOM-929

Replaces #1716


zburke:
Merged #1716 instead. SonarCloud will not scan a fork, but pushing the forked branch as a local branch and creating a PR from it allows the same scan to validate both branches, allowing the original PR to be merged and this one to be closed.

PR #1721, opened 2022-02-09
STCOM-927 Perform proportional resize on panes with percentage-based widths

JohnC-80:
### Problem
Pane widths using a percentage were not proportionally resized, which would leave them mis-matched if they were in a set with a pixel-based value. Ending panes would end up cut off or falling short (leaving a white space between the pane and the edge of the window)
See https://issues.folio.org/browse/STCOM-927 for the previous case...

### Approach
If the layout cache uses a percentage-based width, parse the value and perform calculations to increase or decrease the number based on the relative size of the container.

### DIFF
--- a/lib/Paneset/Paneset.js
+++ b/lib/Paneset/Paneset.js
@@ -19,6 +19,9 @@ function getNewPanewidthObject(key, CSSvalue, proportionalChange) {
   let newWidth = CSSvalue;
   switch (unit) {
     case 'percent':
+      newWidth = parseFloat(CSSvalue, 10) * proportionalChange;
+      newWidth = `${newWidth}%`;
+      break;
     case 'vw': break;
     case 'px':
       newWidth = Math.ceil(parseInt(CSSvalue, 10) * proportionalChange);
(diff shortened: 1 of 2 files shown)


lib/Paneset/Paneset.js
```diff
@@ -19,6 +19,9 @@ function getNewPanewidthObject(key, CSSvalue, proportionalChange) {
   let newWidth = CSSvalue;
   switch (unit) {
     case 'percent':
+      newWidth = parseFloat(CSSvalue, 10) * proportionalChange;
+      newWidth = `${newWidth}%`;
+      break;
```
zburke, line 24: The other cases get rounded up to a whole integer by `Math.ceil`, and `percent` _used_ to when its case fell through along with `vw` and `px`. How come it doesn't apply now?

JohnC-80: rem and em all amount to exact `px` values in the end .. and browsers (at the time, anyways) would treat these decimals differently - some round up, some don't... with percentage-based values even the rounded numbers could amount to decimals at the pixel level... - this is probably why I just avoided calculating with them previously...

PR #1722, opened 2022-02-11
Converted MetaSection to new interactors

ryandberger:
- Also updated MetaSection.test
- This will also be used by `StripesSmartComponents` `ViewMetaData-test`


zburke (approved):

ryandberger:
Will hold off on merging for now since a number of `stripes-smart-components` tests directly reference `MetaSectionInteractor`, so will need to update those calls and merge lock-step with this PR.

JohnC-80:
This interactor should probably be in stripes-testing. If this one is replaced it may break existing tests that use the old interactor.

JohnC-80 (changes requested):
See my last comment... Another repo added to a multi-repo PR, I know..

ryandberger:
> This interactor should probably be in stripes-testing. If this one is replaced it may break existing tests that use the old interactor.

Makes sense. Will do that!

ryandberger:
> > This interactor should probably be in stripes-testing. If this one is replaced it may break existing tests that use the old interactor.
> 
> Makes sense. Will do that!

See: https://github.com/folio-org/stripes-testing/pull/232

zburke:
@ryandberger , what's the status with this PR? It looks like there was a conversation about leaving the existing interactor alone and updating tests to import a new one from stripes-testing. We did create the new interactor in folio-org/stripes-testing/pull/232, but maybe we didn't finish reverting the interactor here. 

Should this be closed or update?

JohnC-80 (approved):

ryandberger:
> use update interactor from stripes-testing

@JohnC-80, made the update to use the metasection interactor from stripes-testing, so I will plan to merge now that this has been cleaned up.

PR #1724, opened 2022-02-16
STCOM-931 prevent onMount from passing to DOM element

JohnC-80:
This is showing up in settings:
image

The `onMount` prop just needs to be filtered out before passing props to internally rendered HTML elements.


MikeTaylor (approved):
Brilliant! Thanks, @JohnC-80.

PR #1725, opened 2022-02-16
STCOM-932 Check for stable id's before caching pane sizes

JohnC-80:
### Problem

Paneset layout caching is used by panesets to keep track of pane sizes between transitions/changes in the sizes of the panes. This eases the transition between views by retaining the widths that the user had previously set.
the smart component 'PersistedPaneset' pushes the paneset cache up to the browser's local storage, so the user-applied resizing can be applied the next time they log in to FOLIO, or return to an app.

Panes cache their widths based on their id. If an id isn't provided as a prop, one will be generated. A generated id will not be repeated the next time that the pane is rendered, so it will be found as unique and added to the list of caches as-if it were a different paneset.

This can cause the local storage cache to grow every time a Pane is initially rendered.

### Approach
Check to see if current panes have any generated id's before caching widths.

### DIFF
--- a/lib/Pane/Pane.js
+++ b/lib/Pane/Pane.js
@@ -94,13 +94,16 @@ class Pane extends React.Component {
   }
 
   componentDidMount() {
-    const { transition, defaultWidth, paneset, resizer, paneTitleRef, onMount } = this.props;
+    const { transition, defaultWidth, paneset, resizer, paneTitleRef, onMount, id } = this.props;
     const paneObject = {
       transition,
       setStyle: this.setStyle,
       isThisMounted: this.isThisMounted,
       defaultWidth,
       id: this.id,
+      // whether or not the pane had to generate its id.
+      // This is used to detarmine whether the size of the panes should be cached.
+      generatedId: !id,
       getRef: this.getRef,
     };
     if (paneset) {
--- a/lib/Pane/Pane.js
+++ b/lib/Pane/Pane.js
@@ -102,8 +102,8 @@ class Pane extends React.Component {
       defaultWidth,
       id: this.id,
       // whether or not the pane had to generate its id.
-      // This is used to detarmine whether the size of the panes should be cached.
-      generatedId: !id,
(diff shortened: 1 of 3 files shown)


lib/Pane/Pane.js
```diff
@@ -94,13 +94,16 @@ class Pane extends React.Component {
...
-    const { transition, defaultWidth, paneset, resizer, paneTitleRef, onMount } = this.props;
+    const { transition, defaultWidth, paneset, resizer, paneTitleRef, onMount, id } = this.props;
     const paneObject = {
       transition,
       setStyle: this.setStyle,
       isThisMounted: this.isThisMounted,
       defaultWidth,
       id: this.id,
+      // whether or not the pane had to generate its id.
+      // This is used to detarmine whether the size of the panes should be cached.
```
zburke, line 105: pedantic: determine

lib/Pane/Pane.js
```diff
@@ -94,13 +94,16 @@ class Pane extends React.Component {
...
+      generatedId: !id,
```
zburke, line 106: Boolean values that start with `is` are really nice, e.g. `isGeneratedId = !id` or `isStaticId = !!id`. Then you can write conditionals like `if (foo.isAttribute)` instead of `if (foo.attribute === false)` and not worry that you're losing any meaning, e.g. "do I have to compare against false and null and undefined?"

zburke (approved):
Thanks for the code comments explaining what this code does and why. They made this really easy to grok. I left a few style comments which you can take or leave.

lib/Pane/Pane.js
JohnC-80: hmm... how about `hasGeneratedId`?
 -- `is*` for some reason, always sounds like a function you'd call to determine something's status.

JohnC-80:
Thanks for the suggestions. I need to write more comments... I find myself wondering what I was thinking too often nowatimes.

PR #1727, opened 2022-02-22
STCOM-935 Add aria label for `<MultiSelectFilterField>` component

OleksandrHladchenko1:
In scope of UIU-1686 it is necessary to add `aria-label` for `<MultiSelectFilterField>` component.
* multiselect-5-input (in tags multiselect)
* owner-service-point-input (in settings > fee/fine > owners > new)

Refs: https://issues.folio.org/browse/UIU-1686

### DIFF
--- a/lib/MultiSelection/MultiSelectFilterField.js
+++ b/lib/MultiSelection/MultiSelectFilterField.js
@@ -65,6 +65,7 @@ class MultiSelectFilterField extends React.Component {
       <input
         {...getInputProps({
           'aria-labelledby': rest['aria-labelledby'] || ariaLabelledBy,
+          'aria-label': `input-${ariaLabelledBy}`,
           'type': 'text',
           'ref': inputRef,
           'onKeyDown': this.handleInputKeyDown,


zburke (commented):
Typically, a DOM element has either an `aria-labelledby` or `aria-label` attribute rather than both, though the former takes precedent. Do you understand why the `aria-labelledby` isn't providing a useful reference here? i.e. are props missing or incorrectly configured? Or is this an Axe glitch? 

If you're confident providing both attributes is correct, that's fine, but please include an explanation about why since it is unusual.

OleksandrHladchenko1:
> Typically, a DOM element has either an `aria-labelledby` or `aria-label` attribute rather than both, though the former takes precedent. Do you understand why the `aria-labelledby` isn't providing a useful reference here? i.e. are props missing or incorrectly configured? Or is this an Axe glitch?
> 
> If you're confident providing both attributes is correct, that's fine, but please include an explanation about why since it is unusual.

Hi, @zburke!  `aria-labelledby` prop equals **undefined** in the second scenario, as the result `aria-labelledby` attribute is not shown in DOM element. And Axe requires to have label for form element, that's why I use `aria-label` attribute with `aria-labelledby`

zburke (approved):

zburke (commented):
`aria-labelledby` is a reference to an attribute, not intended to be a human-readable string. Is using it here in `aria-label` appropriate, or should there be a new prop as the primary source that falls back on `input-${ariaLabelledBy}`?

PR #1729, opened 2022-02-23
STCOM-936 NoValue component doesn't have role

NikitaSedyx:
purpose: https://issues.folio.org/browse/STCOM-936
axe reports many issues when NoValue component is displayed

### DIFF
--- a/lib/NoValue/NoValue.js
+++ b/lib/NoValue/NoValue.js
@@ -13,13 +13,17 @@ function NoValue({ ariaLabel = null }) {
   return (
     <FormattedMessage id="stripes-components.noValue.noValueSet">
       {([message]) => (
-        <span
-          role="textbox"
-          aria-label={ariaLabel || message}
-          data-test-no-value
-        >
-          {'-'}
-        </span>
+        <>
+          <span className="sr-only">
+            {ariaLabel || message}
+          </span>
+          <span
+            aria-hidden
+            data-test-no-value
+          >
+            {'-'}
+          </span>
+        </>
       )}
     </FormattedMessage>
   );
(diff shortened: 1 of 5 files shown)


AndreiShumski2 (approved):

usavkov-epam (approved):

lib/NoValue/NoValue.js
```diff
@@ -14,6 +14,7 @@ function NoValue({ ariaLabel = null }) {
     <FormattedMessage id="stripes-components.noValue.noValueSet">
       {([message]) => (
         <span
+          role="textbox"
           aria-label={ariaLabel || message}
```
JohnC-80, line 18: This error probably occurs due to the aria-label on an otherwise non-interactable element. 
I think we could remove this `ariaLabel` attribute, and instead, render a span with an `sr-only` class (a globally available CSS class.) This would be better than applying a textbox role to something that isn't a textbox.

JohnC-80 (changes requested):

lib/NoValue/NoValue.js
NikitaSedyx: yep, it's because of ariaLabel, not sure I got what `sr-only` class is, but I'll try

as for now ariaLabel looks reasonable to me to tell what `-` means in key values

JohnC-80: Thanks. You don't have to import any CSS or anything... just mark up as `<span className="sr-only">{message || ariaLabel}</span>` . You can think of roles as a promise to assistive technology that the element behaves in a certain way, so telling it that a span is a `textbox` isn't the best way, although it does resolve the axe error. Aria-labels require focus in order to be announced, which requires tabindex and more axe and probably more axe errors since it's only a span and not something that's interactive/with an interactive role.

NikitaSedyx: @JohnC-80 got it, thank you!

yep, this makes more sense, applied changes with sr class and aria-hidden, please review one more time

JohnC-80 (approved):
Awesome. Thank You @NikitaSedyx .

PR #1731, opened 2022-02-23
lint. it does a body good.

zburke:
@JohnC-80 , I'm not confident about changing `& .root` to `&.root` in `TextLink`, but I can't get stylelint to be quiet otherwise. Moving that block above the main `.root` block causes even more complaints 😕 .

### DIFF
--- a/lib/TextLink/TextLink.css
+++ b/lib/TextLink/TextLink.css
@@ -79,8 +79,7 @@
 }
 
 div[class*="Selected"] {
-
-  & .root {
+  &.root {
     color: var(--link-color-current);
 
     /* Visited */
--- a/lib/TextLink/TextLink.css
+++ b/lib/TextLink/TextLink.css
@@ -79,7 +79,15 @@
 }
 
 div[class*="Selected"] {
-  &.root {
+  /**
+    * :not is the crown prince of pseudo selector specificity.
+    * That's causing stylelint grief here since it's more
+    * specific than the things above it, but if we invert the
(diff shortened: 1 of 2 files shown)


lib/TextLink/TextLink.css
```diff
@@ -79,8 +79,7 @@
 }
 
 div[class*="Selected"] {
-
-  & .root {
+  &.root {
```
JohnC-80, line 82: Yeah... that's definitely not what we want here....
`div[class*="Selected"].root` = both classes on the same element vs
`div[class*="Selected"] .root` = the `.root` class within the `.*Selected*` element. I'll fuss around and see if I can appease it.

JohnC-80: Stylelint is being a bear about this. So `:not` is practically the king of all pseudo selectors - it has high specificity, so SL's going to nag about those selectors as long as anything comes after this. @zburke, I recommend `/* stylelint-disable */` here.

JohnC-80 (changes requested):

lib/TextLink/TextLink.css
JohnC-80: Only way I think we might appease is if we just forgot about the nice nesting postcss allows for us and wrote out the raw selectors old-school style.

zburke:
`linter-thingy-disable` should be `linter-thingy-stfu`. It would be good for developers' mental health.

JohnC-80 (approved):

PR #1732, opened 2022-02-23
lint. it's what's for dinner.

zburke:
@BogdanDenis , do these additional dependencies to the hook functions in `AdvancedSearch` look correct to you?


JohnC-80 (approved):
I think some code could be reduced in the component, but this looks good in tests and seems to behave as expected in storybook.

PR #1733, opened 2022-02-23
STCOM-935 Add aria label for `<MultiSelectFilterField>` component

OleksandrHladchenko1:
In scope of UIU-1686 it is necessary to add `aria-label` for `<MultiSelectFilterField>` component.

* multiselect-5-input (in tags multiselect)
* owner-service-point-input (in settings > fee/fine > owners > new)

Refs: https://issues.folio.org/browse/UIU-1686

Previous PR: https://github.com/folio-org/stripes-components/pull/1727


JohnC-80 (changes requested):
Please see https://github.com/folio-org/stripes-components/blob/master/lib/MultiSelection/stories/CustomAriaLabelledBy.js for an example of how this needs to be used. This needs to be aria-labelledby to the id of the accordion header.

PR #1738, opened 2022-02-25
STCOM-942 Selection - possible axe gotcha

JohnC-80:
not providing a label to Selection could result in the error:
```
 Selection
    × does not have axe issues - Selection
      Chrome 98.0.4758.102 (Windows 10)
    AxeError:  Axe violation(s): 
     Issue #1: button-name - Buttons must have discernible text
        description: Ensures buttons have discernible text
      helpUrl: https://dequeuniversity.com/rules/axe/4.4/button-name?application=axeAPI
    
          html: <button type="button" aria-expanded="false" aria-labelledby="selected-testingId-item" aria-haspopup="listbox" class="selectionControl---1j1xE formControl---3BY__" id="testingId" aria-owns="sl-testingId"><div class="singleValue---1vQYM" id="selected-testingId-item"></div></button>
        target: #testingId
    
          Failure Summary:
            Fix any of the following:
          -  Element does not have inner text that is visible to screen readers
          -  aria-label attribute does not exist or is empty
          -  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
          -  Element has no title attribute
          -  Element's default semantics were not overridden with role="none" or role="presentation"   
    
        at C:/Users/JCoburn/AppData/Local/Temp/_karma_webpack_764450/commons.js:301292:23
 ```

### CODE REVIEW

File: lib/Selection/SingleSelect.js [hunk 1]
```diff
@@ -707,6 +707,7 @@ class SingleSelect extends React.Component {
             autoFocus={this.props.autoFocus}
             disabled={this.props.disabled}
           >
+            <span className="sr-only">select control</span>
```
zburke, line 710: How about `<FormattedMessage id="stripes-components.selection.selectControl" />`

zburke (changes requested):
Don't for get i18n when working on a11y :)

zburke (approved):
LGTM. Yay a11y!

ryandberger (approved):

PR #1660, opened 2021-11-15
STCOM-906 - Axe testing for stripes-components

JohnC-80:
Big PR posting for tracking - will break out into smaller PR's as it covers a few areas

Adds axe tests to stripes components that were not changed individually.

Accessibility changes to individual components still in this PR include:
- ~Autosuggest~ #1735 
- ~Selection (SingleSelect)~ #1738 
- ~Timepicker (Time dropdown)~ #1739 
- ~MCL~ #1736 
-  ~IconButton~ #1737

PR #1739, opened 2022-02-28
STCOM-945 Timepicker axe issue

JohnC-80:
runAxeTest output:
 ```
Timepicker
    when clicking the timepicker input
      × contains no axe errors - Timepicker: open menu
        Chrome 98.0.4758.102 (Windows 10)
      AxeError:  Axe violation(s): 
       Issue #1: aria-hidden-focus - ARIA hidden element must not contain focusable elements
          description: Ensures aria-hidden elements do not contain focusable elements
        helpUrl: https://dequeuniversity.com/rules/axe/4.4/aria-hidden-focus?application=axeAPI
      
            html: <input class="sr-only" aria-hidden="true">
          target: input[aria-hidden="true"]
      
            Failure Summary:
              Fix all of the following:
            -  Focusable content should be disabled or be removed from the DOM

          at C:/Users/JCoburn/AppData/Local/Temp/_karma_webpack_243455/commons.js:298482:23
```

### CODE REVIEW

File: lib/Timepicker/TimeDropdown.js [hunk 1]
```diff
@@ -438,7 +438,7 @@ class TimeDropdown extends React.Component {
             >
               Set time
             </Button>
-            <input className="sr-only" aria-hidden onFocus={this.props.onHide} />
+            <input className="sr-only" aria-label="back to field" onFocus={this.props.onHide} />
```
zburke, line 441: Wait, what? Shouldn't `"back to field"` be sth like,
```
props.intl.formatMessage({ id: "stripes-components.timepicker.backToField" })
```
In fact, there's a whole boat load of hard-coded English here:
* line 358 `"hours"`
* line 378 `"minutes"`
* line 394 `"am or pm"`
* line 401 `{this.state.ampm}` (it's not am/pm in every language, is it?)
* line 439 `Set time`

Do all those `<IconButton>` elements need `aria-label` attributes too? Seems like they should have something, doesn't it?

zburke (changes requested):
Yay for labels and a11y! But they need to be localized, too. I can file a separate localization ticket, [STCOM-946](https://issues.folio.org/browse/STCOM-946) to handle the list, but we should at least not aggravate the problem with more English here.

zburke (changes requested):
This still doesn't work for locales that use 12-hour time but not `am` and `pm`, e.g. Arabic (`ar`) where `formatDateToParts()` for an afternoon timestamp returns `[..., {type: "dayPeriod", value: "م"}]`. I don't think we can lean on am/pm, either for logic or display.

JohnC-80:
I'll be happy to address in internationalization on time logic in a separate PR. The original goal of this one was so that we can get #1660 across the board.

zburke (approved):

PR #1740, opened 2022-02-28
STCOM-944 provide NumberFormat component

zburke:
Refs [STCOM-944](https://issues.folio.org/browse/STCOM-944)


zburke:
Note 
```
      numberField.has({ type: 'nuasdfasdfmber' });
```
but that test passes. Huh? 

Note the second test in each describe fails, despite no changes to the selector/interactor. Huh?

PR #1737, opened 2022-02-25
STCOM-941 Datepicker axe tests, fix attributes for aria-expanded, aria-haspopup for IconButton

JohnC-80:
output of runAxeTest:
```
    render
      × has no axe errors. - Datepicker
        Chrome 98.0.4758.102 (Windows 10)
      AxeError:  Axe violation(s): 
       Issue #1: aria-allowed-attr - Elements must only use allowed ARIA attributes
          description: Ensures ARIA attributes are allowed for an element's role
        helpUrl: https://dequeuniversity.com/rules/axe/4.4/aria-allowed-attr?application=axeAPI        
      
          Sample issue (1 of 1 detections)
            html: <span class="iconButtonInner---3RA-J interactionStyles---G9ZCF" data-test-calendar-button="true" aria-haspopup="true" aria-expanded="false">
          target: .iconButtonInner---3RA-J
      
            Failure Summary:
              Fix any of the following:
            -  ARIA attribute is not allowed: aria-expanded="false"
      
          at C:/Users/JCoburn/AppData/Local/Temp/_karma_webpack_964982/commons.js:301544:23
```

PR #1742, opened 2022-03-03
STCOM-948: Add 'aria-expanded' attr to IconButton

mariia-aloshyna:
## Purpose
There is an accessibility issue in users app Open Loans List: "Elements must only use allowed ARIA attributes". This happens because the button's 'aria-expanded' attribute is extended to child elements.

## Approach
- Catch the attribute and assign it manually to the button

## Ref
https://issues.folio.org/browse/STCOM-948


zburke (changes requested):
Thank you for a clear explanation of the problem and how this change addresses it! Pulling `aria-expanded` out of the values that cascade down makes sense to me. @JohnC-80 , do you agree? 

Please add a prop-types entry for this to clean up the lint warnings.

lib/IconButton/IconButton.js
```diff
@@ -86,6 +88,7 @@ const IconButton = React.forwardRef(({
 IconButton.propTypes = {
   'aria-label': PropTypes.string,
   'aria-labelledby': PropTypes.string,
+  'aria-expanded': PropTypes.string,
```
zburke, line 91: Pedantic: as noted by lint prop-types need to be sorted alphabetically so this should be added at the top of the list.

zburke (changes requested):
So close! Please clean up the prop-types ordering and then I think we're done. Agreed, @JohnC-80?

JohnC-80 (approved):
#1737 was a similar issue - it adds `aria-haspopup` as well.
If we get many more of these, we might do something to consolidate the `aria-` props somehow without adding new proptypes, but for now, this is the way....

mariia-aloshyna:
Thanks for review @zburke and @JohnC-80! Prop-types are cleaned up.

PR #1745, opened 2022-03-10
STCOM-835: Break long words in headings based on zooming 200%

mariia-aloshyna:
## Purpose
Long words in headings don't break. They overlap with the text of the column to the right, making them unreadable.

## Approach
- Apply `overflow-wrap: break-word;` to the `KeyValue`

## Refs
https://issues.folio.org/browse/STCOM-835

## Screenshot
image


zburke (commented):
This looks good to me but I'd like to get approval from a few POs before we move ahead with it. I asked them for comments on the Jira ticket. Let's give them a day or two, then we can merge this and pick it onto a patch branch for Lotus as well.

JohnC-80:
We might need to move on a better grid system that can wrap the columns under these circumstances. Something we can probably replace with container queries once they materialize. This is an okay change for now will be needed for wrapping even if such a system is in place.

JohnC-80 (approved):
Approving this - I think it'll be necessary regardless what PO's decide on... Accessiblity principles don't like a) what we currently have or b) some other type of overflow treatment (elipsis) so breaking words is best.

zburke (approved):
I'm glad to see this, but it does remind me that we are awfully [lazy about columns](https://github.com/folio-org/ui-users/blob/36e521599509e7fa1bfdaef08a8e9126c6765356/src/components/UserDetailSections/UserInfo/UserInfo.js#L73) with lines like 
```
<Col xs={3}>
```
instead of 
```
<Col xs={12} sm={6} md={4}>
```
It almost makes me wonder if those layouts should be even more standardized so that tenants could set config values that determine their column counts. English-speaking tenants might choose a four-column layout while Germans might choose a two- or three-column layout.

JohnC-80:
I could see something like a minimum-column-width setting where the containers/columns can determine the width that would work... I can also see an 'off' switch for certain Grid components that just want to stick to the viewport-based media queries.

PR #1746, opened 2022-03-11
STCOM-951 Export to CSV not handling diacritics

NikitaSedyx:
Purpose: https://issues.folio.org/browse/STCOM-951

Approach: use unicode symbol to properly handle special characters like é


AndreiShumski2 (approved):

zburke (approved):
Perfect fix! Even better, add a comment explaining why you need the BOM here and what effect it has.

JohnC-80 (approved):
Those darn Endians. Thanks @NikitaSedyx !

PR #1749, opened 2022-03-19
STCOM-835: Break long words in MCL headings based on zooming 200%

mariia-aloshyna:
## Purpose

Fix MCL headings

## Refs
https://issues.folio.org/browse/STCOM-835

## Before
image

## After
image


JohnC-80 (approved):
All things considered, this is a change that's easy to roll back if there are any issues. Thank you @mariia-aloshyna !

I do wonder, out of curiousity, did you zoom in after the initial render of the UI or if your browser was already set at that zoom level when the listing rendered. Thanks again!

mariia-aloshyna:
Hi @JohnC-80 I zoomed in after the initial render of the UI. But anyway, at 100% zooming and long headings, it gets mixed up.

zburke (approved):

PR #1751, opened 2022-03-23
STCOM-955: Advanced Search: Updating Advanced Search modal query does not return expected results

BogdanDenis:
## Description
Fix splitting query string that is passed into `<AdvancedSearch>`.

## Approach
There's a pretty complex RegEx that is used to splita string into rows: `((?:or|and|not)\s+)?([\w]+==.+?(?=(or|not|and)\s[\w]+==[^\s]+)|[\w]+==.+)`
Here's what is does:
- ((?:or|and|not)\s+)? - this will match boolean operator at the beginning of a row. this needs to be at the beginning because a boolean that is related to a row is placed before the `searchOption==query` bit
- ([\w]+==.+?(?=(or|not|and)\s[\w]+==[^\s]+) - this will match a structure like `keyword==Something something`
it will match any character until it finds a string: `or`, `and` or `not` with a next search option.
- [\w]+==.+ - this will also match a structure like `keyword==Something something`.
this part of regex is OR'd with previous because you can't use lookahead conditionally

## Screenshots
Uploading MARC Authority - FOLIO (4).mp4…

## Issues
[STCOM-955](https://issues.folio.org/browse/STCOM-955)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Correctly label focus-trap control in `<Timepicker>`. Refs STCOM-945.
 * Avoid `setState` calls in unmounted components. Refs STCOM-952.
 * Break long words in headings based on zooming 200%. Refs STCOM-835.
+* Improve splitting search query into rows in <AdvancedSearch>. Fixes STCOM-955.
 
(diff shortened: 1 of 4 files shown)


zburke (approved):
Thanks for the helpful, thorough comments!

mariiaschv (approved):

Svitlana-Korol (approved):

PR #1758, opened 2022-04-04
Remove react-hot-loader dependency

JohnC-80:
https://issues.folio.org/browse/STCOM-954

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,6 @@
     "query-string": "^6.1.0",
     "react-flexbox-grid": "1.1.3",
     "react-highlight-words": "^0.16.0",
-    "react-hot-loader": "^4.0.0",
     "react-overlays": "^4.1.1",
     "react-quill": "^1.3.3",
     "react-svg-loader": "^3.0.3",


zburke (approved):
I think the `webpack` change should be handled separately. If that trouble ever gets cleaned up from webpack's side, we'd be able to revert that PR, but we can't do that if it's coupled to this RHL change. (I think? Or do you think we'll un-pin webpack as "new work" rather than "undoing old work"?)

If the former, let's separate these. If the latter, go ahead and merge like the wind.

BogdanDenis:
Hey @JohnC-80, we have failing builds in ui-eholdings after we removed react-hot-loader
I see that you also merged the same changes to stripes-components yesterday and there's still one import in a util https://github.com/folio-org/stripes-components/blob/master/util/childrenOf.js
Not sure why this unresolved import wasn't picked up by eslint, however.

I couldn't find any documentation on `react-fast-refresh` or `react-refresh` so not sure if there's a `areComponentsEqual` function to replace current implementation

Also, `areComponentsEqual` is only used in `childrenOf` util to check prop types in `<Tooltip>` component, which seems a bit of an overkill to me
CC: @zburke

BogdanDenis:
@JohnC-80 @zburke update: we have also failed builds in ui-quick-marc, which blocks us from merging bugfix tickets :(
Is it possible to revert this change for now and continue after Lotus Bugfest deadline?

PR #1760, opened 2022-04-05
Downgrade webpack to 5.68.0

JohnC-80:
Downgrade due to 5.69, 5.70 causing heavy logs and failing to find karma-webpack.


zburke:
@JohnC-80, do you know why webpack is even a dep here? I know it's a hard-dep in stripes-cli and eslint-config-stripes, but I thought those were the only two places where we actually needed to specify it.

PR #1762, opened 2022-04-11
STCOM-966: eHoldings app: Apply pagination to Titles list

mariiaschv:
## Purpose
eHoldings app: Apply pagination to Titles list in Usage & analysis accordion on Package Detail Record.

These changes needed to add focus on the correct row in the list when using Prev/Next buttons inside `PrevNextPaginationRow` component.

## Issues
[STCOM-966](https://issues.folio.org/browse/STCOM-966)

Related PRs: [PR for UIEH-1259](https://github.com/folio-org/ui-eholdings/pull/1534)


zburke (commented):
Sorry, with no comments explaining what's happening here, these changes are really opaque to me. I get it that you're adding pagination rather than just previous/next, but there's a ton of stuff that's changing and no indication of how the new stuff works or an indication of how existing stuff will continue to work, e.g. the calculation of `dataStartIndex` and `dataEndIndex` appears to be based on individual props whereas before it was derived from `contentData`. So, what happens to existing implementations that don't provide these props?

I will take a look at this again when there is a clear description of what is changing and why. 

CC: @JohnC-80

mariiaschv:
> Sorry, with no comments explaining what's happening here, these changes are really opaque to me. I get it that you're adding pagination rather than just previous/next, but there's a ton of stuff that's changing and no indication of how the new stuff works or an indication of how existing stuff will continue to work, e.g. the calculation of `dataStartIndex` and `dataEndIndex` appears to be based on individual props whereas before it was derived from `contentData`. So, what happens to existing implementations that don't provide these props?
> 
> I will take a look at this again when there is a clear description of what is changing and why.
> 
> CC: @JohnC-80

@zburke thanks, you are totally right, I should've explained what happens here. I've edited description of the PR and added a couple of comments to the code. I also checked other places where `MCLRenderer` is used with different pagingTypes, seems to work fine for all cases.

BogdanDenis (approved):

Svitlana-Korol (approved):

ODanyor (approved):

JohnC-80 (approved):

PR #1766, opened 2022-04-18
STCOM-968: Add marginBottom0 prop to Timepicker

ncovercash:
# [Jira](https://issues.folio.org/browse/STCOM-968)

This adds a `marginBottom0` property to Timepicker, removing the bottom margin from the field.  This prop works the same as it does in any other similar component, such as `TextField`.


JohnC-80 (approved):

ncovercash:
Updated it to use the variable created from the props destructuring, rather than `this.props` -- fixed the undefined warning.

JohnC-80:
Changes merged in #1766

PR #1767, opened 2022-04-18
STCOM-969: Add usePortal prop to Timepicker

ncovercash:
# [Jira](https://issues.folio.org/browse/STCOM-969)

This adds a `usePortal` property to Timepicker, removing the bottom margin from the field.  This prop works the same as it does in any other similar component, such as `Datepicker`.  A CSS rule had to be added to prevent clicks/mouse events from being missed in the `TimeDropdown` and being instead received by components underneath.


JohnC-80:
Changes merged in a separate PR - closing this one.

PR #1768, opened 2022-04-19
STCOM-972: Fix prop types of <IconButton>

BogdanDenis:
## Description
Props `aria-expanded` and `aria-haspopup` have incorrect prop types validation - some usages pass props as boolean, and some as a string. Since boolean values will be cast to string when added html anyway I think we can safely include this type in prop types check

## Issues
[STCOM-972](https://issues.folio.org/browse/STCOM-972)


zburke (changes requested):
Good catch, and thanks for helping clean this up! Let's keep the `ariaExpanded` and `ariaHasPopup` prop-types in sync with these and then I'm happy to approve this. 

That said, it's tempting to go a step further and _only_ accept `PropTypes.bool`. As you said, although the rendered values will always be strings they will only be the strings `"true"` and `"false"`. We could use `PropTypes.oneOf(["true", "false"])`, but `PropTypes.bool` feels more honest and I think best reflects existing uses. I searched through a platform-complete build and I don't think this would cause any trouble. @JohnC-80, any thoughts on the matter?

BogdanDenis:
@zburke I can find only two places where `aria-expanded` is passed as a string
image
We can leave only `string` for `aria-expanded` if we make changes those components

And `aria-haspopup` can accept boolean and string values https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup

So we should support both `string` and `bool` types for `aria-haspopup`
CC: @JohnC-80

zburke:
Thanks for the thorough research, @BogdanDenis! I never imagined a boolean-named attribute (isThing, hasThing) would take a non-boolean value since that's the entire point of the naming convention 😠  (Given they already use aria-smashed-together instead of aria-not-smashed, I really shouldn't be surprised.) Regardless, clearly we need to allow `string` and `bool` for `aria-haspopup`. 

`SelectList` and `NavDropdownMenu` pass `aria-expanded` directly to DOM elements, not to `<IconButton>` instances, so they're not relevant here.

JohnC-80:
Yeah, we could lock the string down to `"true" or "false"` strings to be more specific. I don't know that it completely blocks things, but it would be more helpful if someone is handling things incorrectly.

BogdanDenis:
Thanks for your input @zburke and @JohnC-80 , I've updated the prop types for `aria-expanded` to use bool because that's what's passed to the component in the system now

zburke (approved):

Svitlana-Korol (approved):

mariiaschv (approved):

JohnC-80 (approved):

ODanyor (approved):

PR #1778, opened 2022-04-27
STCOM-983 Selectively resize percentage-width panesets

JohnC-80:
Currently, the percentage-based panes are always resized to full screen width. This can be strange in cases where there are only 2 panes (like settings) that are not full width. 

The logic in this PR modifies this so that 
- panesets where ALL of the panes are percentage -based are only resized by adjusting the last pane (percentages, for the most part can be left alone, but this takes care of some oddness when the window is resized)
- only panesets with 3 or more panes are held to this percentage resizing.


zburke (approved):
Thanks! The check described in the comments makes sense to me.

PR #1779, opened 2022-04-27
STCOM-982 Switch to percentage-based pane resizing.

JohnC-80:
Currently, panes are resized proportionally via a multiplier value applied to all of the pane widths. The numbers don't work out well when percentage-widths are multiplied.
The new logic just passes percentage values through.


JohnC-80:
Closing for a moment. Will re-open if I determine this can't be simplified.

PR #1780, opened 2022-04-28
Stcom 974

zburke:
#1776 döppleganger so we can see how tests run on a non-forked branch.


zburke:
Replaces #1776, which for some reason never picked up on the successful SonarCloud run here. Weird; usually the original PR is mergeable.

PR #1789, opened 2022-05-03
STCOM-988 Resize nested paneset container

JohnC-80:
Problem:
CSS changes introduced in STCOM-953 (.paneset .paneset {overflow: visible;}) resolved the issue of cropping, but caused additional strange behavior as percentage-based child panes wouldn't size to the parent paneset element, but to its overflow-hidden parent (100% = 100% of window, not 100% of the element)

Approach
This change resizes nested panesets elements so that their rectangle extends to the end boudary of the window.
This removes the previous CSS-based solution, but also resolves the cropping issue the change fixed.

### DIFF
--- a/lib/Paneset/Paneset.css
+++ b/lib/Paneset/Paneset.css
@@ -8,10 +8,6 @@
   will-change: transform;
   overflow: hidden;
 
-  & .paneset {
-    overflow: visible;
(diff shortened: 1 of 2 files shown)


JohnC-80:
Closing for #1791 instead

PR #1790, opened 2022-05-04
Button: 'Link' style should behave as expected

Svitlana-Korol:
## Description
'Link' style should behave as expected

## Issues
[STCOM-938](https://issues.folio.org/browse/STCOM-938)


JohnC-80 (changes requested):
Thanks! Please switch the colors to use variables (similar to TextLink component: https://github.com/folio-org/stripes-components/blob/master/lib/TextLink/TextLink.css ) rather than hard-coded values.

Svitlana-Korol:
> Thanks! Please switch the colors to use variables (similar to TextLink component: https://github.com/folio-org/stripes-components/blob/master/lib/TextLink/TextLink.css ) rather than hard-coded values.

@JohnC-80 ,  I fixed it, thanks!

JohnC-80 (approved):

mariiaschv (approved):

PR #1792, opened 2022-05-04
STCOM-989 Resize the resized panes after window resize 

JohnC-80:
Resize

Before: (snapshot)
pane_problems

After
Panes_fixed


zburke (approved):
The comments here are a huge improvement over how completely opaque things were previously, though I gotta be honest it's still largely opaque. But I appreciate that I at least have some insight about what is supposed to be happening. 

Do you think there's any value in separating some of these private/static functions into their own files so we can easily write tests for them? Coverage is an issue here.

JohnC-80:
I'd love to get better coverage for code that runs on window resize - difficult to test that particular event. It would mean a major refactor of things that I think is better suited for another PR after manual testing of these changes pass. Moving a function out that, say, takes the panes list and produces some widths object.. or takes a container size and resizes some widths to fit that size - those things could probably be moved out.

PR #1795, opened 2022-05-11
STCOM-947 consistent time format of Timepicker

JohnC-80:
# Problem
Tenants of certain locale (en-SE, for example) may not be presented with a time-picker that conforms to their expectations (12-hour selector vs 24-hour selector)

The issue is moment falling back to `en` locale if the requested locale is not available.

# Approach
Use browser Intl API to determine the 'dayPeriods' of the given locale's time format and discern the correct 12 or 24 hour format.

This PR also moves the associated functions to their own file of 'dateTimeUtils'


zburke (changes requested):
Lint is pretty upset; looks like maybe some functions haven't been exported correctly, aren't being imported correctly, etc.? That needs to get cleaned up. But the general plans of "move some common things out of private implementations into util" and "leverage `Intl.DateTimeFormat` when `moment` drops the ball" is sound 👌. 

It is a little hard to follow if you don't already know what's happening though. Reading through the tests helped me understand the code better. Maybe some of the test results could be copied into comments to provide examples? e.g. "given a local of `de`, return a date-format string like `DD.MM.YYYY`".

zburke (approved):

zburke (approved):

PR #1796, opened 2022-05-11
Fix description of rowMetadata in MCL

ncovercash:
Per the source code, [`rowMetadata` is an array](https://github.com/folio-org/stripes-components/blob/master/lib/MultiColumnList/MCLRenderer.js#L222) that, when the [MCL is initialized](https://github.com/folio-org/stripes-components/blob/master/lib/MultiColumnList/MCLRenderer.js#L369), controls [columns which are ignored](https://github.com/folio-org/stripes-components/blob/master/lib/MultiColumnList/MCLRenderer.js#L130).  This PR updates the README to be consistent with the actual usage.


ncovercash:
@JohnC-80 would you please review this?  Thanks!

zburke (approved):
Good catch. Thank you!

PR #1797, opened 2022-05-12
Fix description of rowMetadata in MCL

zburke:
Non-forked dupe of #1796 so we can run tests.


zburke:
Merged the original instead

PR #1799, opened 2022-05-18
STCOM-994 Extend proptypes for avoiding errors

AndreiShumski2:
Purpose:
- add `PropTypes.func` for accordion children for `Open` render-prop usage
- add `PropTypes.node` for `LayoutHeader` actions title to use `<FormattedMessage>` as prop ([AddressView](https://github.com/folio-org/stripes-smart-components/blob/master/lib/AddressFieldGroup/AddressView/AddressView.js#L82))
- introduce new MCL `pagingType` `none` for custom pagination
- add `PropTypes.bool` as one of value type for `Select` option

### DIFF
--- a/lib/LayoutHeader/LayoutHeader.js
+++ b/lib/LayoutHeader/LayoutHeader.js
@@ -11,7 +11,10 @@ const propTypes = {
     PropTypes.shape({
       handler: PropTypes.func,
       icon: PropTypes.string,
-      title: PropTypes.string,
+      title: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.node,
+      ]),
     }),
   ),
   level: PropTypes.number,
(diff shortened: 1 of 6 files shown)


lib/LayoutHeader/LayoutHeader.js
```diff
@@ -11,7 +11,10 @@ const propTypes = {
     PropTypes.shape({
       handler: PropTypes.func,
       icon: PropTypes.string,
-      title: PropTypes.string,
+      title: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.node,
+      ]),
```
zburke, line 17: LGTM, but there is a potential gotcha with expanding acceptable types beyond `string` because values used as DOM element attributes (e.g. `aria-label`, `placeholder`) must be string-only because they are cast to string so a `<FormattedMessage>` will come through as `[object Object]`. 

Again, this is fine, but just be aware that this change cannot be universally applied.

zburke (approved):
😍 A cleaner console makes everybody's life better. THANK YOU!

JohnC-80 (approved):

PR #1801, opened 2022-05-19
Expose additional components via index

mkuklis:
These are used in various modules via `@folio/stripes/components/lib/...` currently. This PR adds them to index so they can be accessed from `@folio/stripes/components`


JohnC-80:
This will resolve https://issues.folio.org/browse/STCOM-978 if you want to take it  🙏

JohnC-80 (approved):

PR #1804, opened 2022-05-22
Update list of components in main README

ncovercash:
I saw there were quite a few components missing from the main listing in the README: this PR adds those.  Please let me know if the categories are incorrect, I tried to follow the current classifications as best as I could.

For components outside of this library, should they be listed here?  There are a number of components in `stripes-smart-components` not listed here, however, there are also a few components in this list pointing directly to `stripes-smart-components` -- please let me know if they should be here or not, and I can add that into this PR.  (There is also `<Pluggable>`, which is in `stripes-core`).

`<AddressFieldGroup>` was moved to `stripes-smart-components` and so I removed it from the list here, however, in accordance to the above I can put it back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,35 +50,52 @@ The following components are available in the [lib](lib) directory.
 Component | doc | categories
 --- | --- | ---
 [`<Accordion>`](lib/Accordion) | [doc](lib/Accordion/readme.md) | structure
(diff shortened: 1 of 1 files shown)


JohnC-80 (approved):

JohnC-80:
Merged in #1806 Thanks @ncovercash!

PR #1805, opened 2022-05-23
STCOM-990: Break long words in (checkbox) labels

Javokhir12:
## Purpose
- Break long words in (checkbox) labels
- https://issues.folio.org/browse/STCOM-990


Javokhir12:
Hi @zburke, @JohnC-80. In the description of the ticket, it says `overflow-wrap: break-word` will fix the issue. But, when I apply it, it didn't work. Words are wrapping only when `overflow-wrap: anywhere;` or `word-break: break-all;` declarations are applied. Is my approach is acceptable?

JohnC-80:
@Javokhir12  Thank you! I think `word-break: break-word` might be a better option - it might result in fewer 'orphaned' letters than `overflow-wrap: anywhere` and `word-break: break-all`. Could you try it and - if it resolves, that'll be what we use... otherwise I think `word-break: break-all` might be best, since overflow judges its breaking based on a single word fitting on a single line...

JohnC-80 (approved):

zburke (approved):

PR #1808, opened 2022-05-30
STCOM-999 Remove paginationBoundaries prop from MCLRenderer and PrevNextPaginationRow components

vashjs:
paginationBoundaries - this property is no longer needed, since by using activeNext and activePrevious we can achieve the same goals and disable the next and prev buttons.

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -218,7 +218,8 @@ class MCLRenderer extends React.Component {
     onRowClick: PropTypes.func,
     onScroll: PropTypes.func,
     pageAmount: PropTypes.number,
-    paginationBoundaries: deprecated(PropTypes.bool), // will be removed in next major release. Use pagingCanGoPrevious and pagingCanGoNext instead.
+    // Use pagingCanGoPrevious and pagingCanGoNext instead.
+    paginationBoundaries: deprecated(PropTypes.bool),
     pagingButtonLabel: PropTypes.node,
     pagingCanGoNext: PropTypes.bool,
     pagingCanGoPrevious: PropTypes.bool,
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -1191,7 +1191,6 @@ class MCLRenderer extends React.Component {
       virtualize,
       pagingType,
       hidePageIndices,
-      paginationBoundaries,
     } = this.props;
 
     const {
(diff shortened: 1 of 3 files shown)


UladzislauKutarkin (approved):

UladzislauKutarkin (commented):
@vashjs update Changelog.md and readme.md about props

JohnC-80:
Since this PR is removing props, that's technically a breaking change... The way we like to handle these things is to deprecate the props via `deprecate` from `prop-types-extra` now and go ahead and remove it from documentation, etc... then when stripes-components goes through a major feature update that would incur a major version, we'll do the actual removal then (along with other deprecated props in the library)... you can see an example of the deprecation here: https://github.com/folio-org/stripes-components/blob/893ea19b2e8a0820b8ce462aaa78cefd7ed736ba/lib/Headline/Headline.js#L59

As always, thank you for the work!

vashjs:
> Since this PR is removing props, that's technically a breaking change... The way we like to handle these things is to deprecate the props via `deprecate` from `prop-types-extra` now and go ahead and remove it from documentation, etc... then when stripes-components goes through a major feature update that would incur a major version, we'll do the actual removal then (along with other deprecated props in the library)... you can see an example of the deprecation here:
> 
> https://github.com/folio-org/stripes-components/blob/893ea19b2e8a0820b8ce462aaa78cefd7ed736ba/lib/Headline/Headline.js#L59
> 
> As always, thank you for the work!

@JohnC-80  Thanks for your update, fixed.

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -217,7 +218,7 @@ class MCLRenderer extends React.Component {
     onRowClick: PropTypes.func,
     onScroll: PropTypes.func,
     pageAmount: PropTypes.number,
-    paginationBoundaries: PropTypes.bool,
+    paginationBoundaries: deprecated(PropTypes.bool), // will be removed in next major release. Use pagingCanGoPrevious and pagingCanGoNext instead.
```
zburke, line 221: Sigh; lint. How about 
```
// use pagingCanGoPrevious and pagingCanGoNext instead.
paginationBoundaries: deprecated(PropTypes.bool), 
```

zburke (changes requested):
Thanks for implementing `deprecate()` as @JohnC-80 suggested; that's perfect! Now we just need to get lint to run cleanly and we can merge this!

lib/MultiColumnList/MCLRenderer.js
vashjs: @zburke  thanks, fixed

zburke (approved):

PR #1820, opened 2022-06-10
STCOM-1009 Add the ability to pass a className to the rows container in the MCL.

usavkov-epam:
## Purpose:
https://issues.folio.org/browse/STCOM-1009


JohnC-80 (approved):
Thank you! This will be useful. I wonder if the class rules that were added on [STSMACOM-670](https://github.com/folio-org/stripes-smart-components/pull/1247/files) should just be added to the existing class.

usavkov-epam:
> Thank you! This will be useful. I wonder if the class rules that were added on [STSMACOM-670](https://github.com/folio-org/stripes-smart-components/pull/1247/files) should just be added to the existing class.

@JohnC-80 So, for `<EditableListForm>` everything looked fine when I tested it locally. But there might probably be some other case where you need to rewrite an existing class. Perhaps in such case it would be better to use function like "getRowContainerClass" that accept default class?

Something like that:
```
  const rowContainerClass = (typeof getRowContainerClass === function)
     ? getRowContainerClass(css.mclRowContainer)
     : css.mclRowContainer
```

And then set `rowContainerClass ` as className for container. What do you think about it?

zburke (approved):

JohnC-80:
@usavkov-epam Yeah, definitely okay for that. 👍

mkuklis (approved):

AndreiShumski2 (approved):

PR #1821, opened 2022-06-10
STCOM-1010: MultiSelection - exception when using special characters in search string

BogdanDenis:
## Description
Fix exception when searching in `<MultiSelection>` with special symbols

## Screenshots

https://user-images.githubusercontent.com/19309423/173042816-3d5633eb-fe9d-4d6e-91c8-c49f5680d997.mp4

## Issues
[STCOM-1010](https://issues.folio.org/browse/STCOM-1010)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 * Remove paginationBoundaries prop from MCLRenderer and PrevNextPaginationRow components. Refs STCOM-999.
 * Paneset logic handles container widths of 0. Refs STCOM-1004.
 * properly provide popper placements. Fixes STCOM-979.
+* MultiSelection - fix exception when using special characters in search string. Fixes STCOM-1010.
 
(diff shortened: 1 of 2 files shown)


Svitlana-Korol (approved):

mariiaschv (approved):

JohnC-80 (approved):
Thanks @BogdanDenis

zburke (approved):
I love little fixes like this! Even more, I would love a comment that explains what this regex is doing so the next person who encounters it doesn't have to look up the PR. IIUC, it's escaping characters in `filterText` that have special meaning to the reg-ex parser, right?

BogdanDenis:
> I love little fixes like this! Even more, I would love a comment that explains what this regex is doing so the next person who encounters it doesn't have to look up the PR. IIUC, it's escaping characters in `filterText` that have special meaning to the reg-ex parser, right?

You're right, I'll add a comment

PR #1822, opened 2022-06-10
Headline: Update description of props

ncovercash:
It seems like the `label` prop was removed a few years ago, however, the documentation for `<Headline>` still refers to the `children` prop as an "alternative": https://github.com/folio-org/stripes-components/commit/990df3318062db963fb15aac4e516c785236ba81

Additionally, this PR adds descriptions for props that were previously not described: `block`, `flex`, and `className`.


JohnC-80:
Merged in #1826

PR #1823, opened 2022-06-10
Update Icon prop documentation

ncovercash:
The <Icon> documentation stated that `aria-label` was supported, however, it was not.   I updated the implementation similar to <IconButton>: https://github.com/folio-org/stripes-components/blob/master/lib/IconButton/IconButton.js#L60

Additionally, the `status` prop was not previously documented.


ncovercash:
Also corrected the tabIndex declaration and added a table row for that in the docs.

zburke:
@ncovercash , I don't understand why `pickby(startsWith('aria'))` does not support `aria-label`. Can you explain? 

Flipping `tabIndex` from string to number may be Technically Correct but it'll cause a barrage of new prop-type warnings that we'll need to correct across stripes-* and ui-* repos:
```
platform-complete/node_modules/@folio(snapshot)*
$ grep -r 'tabIndex="' .  --exclude-dir node_modules | wc
     105     432   12459
```
so I think we should advertise our intent to make the change on Slack#stripes-updates at the start of a development cycle. Admittedly, the props tweak is small, but it's still a bunch of churn, and this will give teams enough time to fix it at their leisure.

ncovercash:
@zburke oops, meant ariaLabel was supposed to be supported, not aria-label. The existing code was fine for the kebab version. 

And completely understand, don't want to break anything existing.  Should I leave it as number in the README or should that describe it as string, too?

zburke (approved):

zburke:
@ncovercash, can you update this branch so we can get it merged? Thanks!

ncovercash:
@zburke done!

ncovercash:
@zburke ah, looks like `aria-label` isn't valid on the icon component.  Should we remove this prop, then?

zburke:
@JohnC-80 , penny for your thoughts. As @ncovercash notes, 

> looks like aria-label isn't valid on the icon component. Should we remove this prop, then? It wasn't initially implemented, but was specified in the readme

IOW, this PR aimed to bring the code into compliance with its documentation, but it appears that the documentation was wrong in the first place.

zburke:
I totally understand, @ncovercash. It turns out this a widespread issue that may affect many components. @JohnC-80 filed [STCOM-1165](https://issues.folio.org/browse/STCOM-1165) so we can address this head-on. It's unfortunate that the fix wasn't as quick and easy as it looked initially, but it's good that you alerted us to the issue. So, even though we don't have a fix, there's definitely some good that came out of it. Thank you for bringing it to our attention!

PR #1824, opened 2022-06-10
STCOM-1012: Add initial typescript type declarations

ncovercash:
# [Jira STCOM-1012](https://issues.folio.org/projects/STCOM/issues/STCOM-1012?filter=allissues)

This PR adds initial type declarations to this library.  These type definitions are effectively meaningless as every export is declared as `any` (which can match functions, classes, constants, etc, with no type information).  However, they provide a starting point for the remainder of STCOM-1012 as types can now be added incrementally.

Please track the Jira ticket to see examples on how these type declarations can be written.


zburke:
I'm a typescript noob; please bear with me. 

Does this change mean we need to keep `index.js` in sync with `types/index.d.ts`, i.e. anything we add to the former we need to add to the latter? Should we add comments to that effect in `index.js`?

JohnC-80:
Thanks @ncovercash for the initial pass.
Handful of questions around this... 
I am typescript-familiar, but I don't have the forward details worked out for this.

I do like the developer experience that TS provides along with its friendliness to developers who are accustomed to working in strongly typed languages... but there are some questions about how things are going to work for this and the maintenance and special cases it could entail.

All that said, this would be a perfect thing to talk about at a stripes-architecture meeting... @ncovercash - we meet every Thursday at 9AM CST - it would be great to talk about it at that meeting with a handful of stripes devs/module devs from elsewhere in the FOLIO ecosystem.

PR #1825, opened 2022-06-14
STCOM-1011: Browse contributors > Second pane should not show a horizontal scrollbar

BogdanDenis:
## Description
Fix horizontal scrollbar in MCL

## Screenshots
image

## Issues
[STCOM-1011](https://issues.folio.org/browse/STCOM-1011)

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -50,10 +50,12 @@ import LoadMorePaginationRow from './LoadMorePaginationRow';
 import PrevNextPaginationRow from './PrevNextPaginationRow';
 
 // used in all margin calculations... static based on CSS value of 15px (gutter-static). (15px * 2 = 30px)
-const ROWMARGIN = 32;
+const ROWMARGIN = 30;
 const CHANGE_NEWDATA = 'new data';
 const CHANGE_LESSDATA = 'less';
 const CHANGE_DIFFERENCE = 'difference';
+
+const SCROLLBAR_WIDTH = 17;
 
 /* some item fields will have arrays... those don't always come back in the same
 *  order, despite the item being the same. This function checks for equality with that in mind.
@@ -977,7 +979,9 @@ class MCLRenderer extends React.Component {
     *  or the rowWidth (sum of the columnWidths)
     *  whichever is greater.
     */
-    const defaultStyle = { minWidth: `${Math.max((widthVar - (hasMargin ? ROWMARGIN : 0)), rowWidthVar)}px` };
+    const defaultStyle = {
+      minWidth: `${Math.max((widthVar - (hasMargin ? ROWMARGIN + SCROLLBAR_WIDTH : 0)), rowWidthVar)}px`,
+    };
 
     const onClick = onRowClick ? (e) => { this.handleRowClick(e, this.getRowData(rowIndex)); } : undefined;
     const onKeyDown = (e) => this.handleRowKeyPress(e, this.getRowData(rowIndex));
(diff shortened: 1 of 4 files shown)


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -1244,15 +1248,12 @@ class MCLRenderer extends React.Component {
...
           positionCache={this.positionCache}
-          prevWidth={prevWidth}
           rowHeightCache={this.rowHeightCache}
           rowIndex={rowIndex}
           sendMessage={this.sendMessage}
           setFocusIndex={this.setFocusIndex}
           staticBody={staticBody}
           virtualize={virtualize}
-          width={width}
-          rowWidth={rowWidth}
```
zburke, line 1255: Lint says `rowWidth` can be omitted now.

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -55,6 +55,8 @@ const CHANGE_NEWDATA = 'new data';
 const CHANGE_LESSDATA = 'less';
 const CHANGE_DIFFERENCE = 'difference';
 
+const SCROLLBAR_WIDTH = 17;
```
zburke, line 58: 17 units of what -- `px`, `em`, ...? And while `SCROLLBAR_WIDTH` is an excellent variable name, I still feel like we gotta explain where on earth `17` comes from, if only so we know where to look when the next version of Chrome changes it to 18 and we start seeing scrollbars again.

zburke (changes requested):
Lint now whines about `rowWidth` being set but unused so we need to clean that up. 

Otherwise, things look good, I think. IIUC, the MCL-width calculation sometimes missed the scrollbar-width, which somebody somehow figured out is 17 pixels. Thus, the main fix here is to consolidate the width calculation into MCLRenderer's `defaultStyle` instead of faffing about with `prevWidth`, `width`, and `rowWidth` props all over the place. If that's the case, can we please say that in the PR description (and make sure that gets copied into the commit comment)? 

If that's not the case, please explain what's actually happening here :)

lib/MultiColumnList/MCLRenderer.js
BogdanDenis: Scrollbar width depends on the browser and OS too, I couldn't find any official documentation on this, only this website from 2018 https://codepen.io/sambible/post/browser-scrollbar-widths
And online tool to measure scrollbar width: https://adminhacks.com/scrollbar-size/
I'll add them to comments

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -55,6 +55,11 @@ const CHANGE_NEWDATA = 'new data';
 const CHANGE_LESSDATA = 'less';
 const CHANGE_DIFFERENCE = 'difference';
 
+// width of scrollbar in pixels. the width depends on user's OS and browser
+// 2018 article on different scroll widths https://codepen.io/sambible/post/browser-scrollbar-widths
+// online tool that will tell you your scrollbar width https://adminhacks.com/scrollbar-size/
+const SCROLLBAR_WIDTH = 17;
```
JohnC-80, line 61: I don't think we need the extra padding of the scroll bar width to the content... in cases where horizontal scroll is necessary, the columns/headers are out of alignment when scrolled all the way to the right side.

I personally prefer the look when the content is flush against the scrollbar, rather than the spacer. 
I found that if I just increment the ROW_MARGIN by 2px (to 32px) to account for the difference of the scroll bar, than the horizontal scroll went a way because the necessary padding is already there, it just isn't quite enough for the current width of the scroll bar.

BogdanDenis: @JohnC-80 Here's how it looks when horizontal scroll is present
image

Increasing the `ROWMARGIN` by 2px also fixed the issue for me, but I just couldn't understand why that was. I guessed it was because some column widths are not integers, and perhaps something didn't add up because of remainders

Svitlana-Korol (approved):

lib/MultiColumnList/MCLRenderer.js
JohnC-80: Yeah... it's fiddly - the browser sees that the element is .1px overflowing and adds a ~17px - wide scrollbar with at least a 17px scrollable range to compensate >.<

PR #1835, opened 2022-06-21
STCOM-975 Record detail panes are empty when printed

usavkov-epam:
## Purpose
https://issues.folio.org/browse/STCOM-975

This bug is the result of an attempt to fix a bug related to an extra blank page ([STCOM-872](https://issues.folio.org/browse/STCOM-872)). Trying to set `height: auto` for the `body` resulted in it collapsing due to absolutely positioned children having no explicit height.

## Approach
Initial issue with extra blank page can be resolved by using `@media print` media query, where margins for `html` and `body` are cleared and page's margins regulated by `@page` at-rule. In this case, the reason for the content collapsing is eliminated and the problem with the empty page is solved.
<details>
  <summary>How it looks locally</summary>
  
 image
image

</details>

---

**NB:** But there is a question about how the pane should be printed. For now it's just one page with visible data - other content is clipped. All because of the visibility settings
<details>
  <summary>How it looks locally</summary>

  image

</details>

I think it makes sense to print the entire contents of the pane. So I have added `@media print` for `Pane` and `Paneset` stylesheets to update `overflow` property for print. It also requires to add `@media print` to [`ModuleContainer.css`](https://github.com/folio-org/stripes-core/blob/master/src/components/ModuleContainer/ModuleContainer.css) in stripes-core.

<details>
  <summary>How it looks locally</summary>
  
chrome_inS0iFgHAW

</details>

If this behavior does not meet the requirements - let me know and I will remove the `@media print` queries from `Pane` and `Paneset`. Otherwise, I can create a ticket to add `@media print` to the `ModuleContainer.css` in stripes-core like:
```
 8
 9  @media print {
10    overflow-y: visible;
11  }
12
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 * Add the ability to pass a className to the rows container in the `<MultiColumnList>`. Refs STCOM-1009.
 * MultiSelection - fix exception when using special characters in search string. Fixes STCOM-1010.
 * Browse contributors > Second pane should not show a horizontal scrollbar. Fixes STCOM-1011.
+* Record detail panes are empty when printed. Refs STCOM-975.
 
 ## [10.1.5](https://github.com/folio-org/stripes-components/tree/v10.1.5) (2022-06-06)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v10.1.4...v10.1.5)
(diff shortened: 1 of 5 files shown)


JohnC-80:
Thank you! @usavkov-epam - this looks great! 👍 👍 Printing the entire contents of the pane makes sense to me.

AndreiShumski2 (approved):

PR #1841, opened 2022-06-27
UIIN-2079 - Browse contributors. Placeholder match does not look right in the contributor column

Terala-Priyanka:
## Description
This PR adds additional style to the cell rendered in the search result list rendered using MCLRenderer, in order the placeholder text is wrapped across cells in the row.

## Screenshots
nckt7qBp5E

## Issues
[STCOM-1018](https://issues.folio.org/browse/STCOM-1018)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 10.2.1 IN PROGRESS
 * Fix 12-hour formatting in `dateTimeUtils` `getLocalizedTimeFormatInfo`. Fixes STCOM-1017
 * * Add `inputRef` prop to `<Timepicker>`. Refs STCOM-1016
+* Browse contributors. Placeholder match does not look right in the contributor column. Refs UIIN-2079
 
 ## [10.2.0](https://github.com/folio-org/stripes-components/tree/v10.2.0) (2022-06-14)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v10.1.0...v10.2.0)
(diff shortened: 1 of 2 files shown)


JohnC-80:
This is an app-level feature/behavior. It should be applied at the app level through the 'resultRowFormatter' that's being passed to the `SearchAndSort` component rather than a special case added to the component.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Fix 12-hour formatting in `dateTimeUtils` `getLocalizedTimeFormatInfo`. Fixes STCOM-1017
 * Add `inputRef` prop to `<Timepicker>`. Refs STCOM-1016
 * `<MultiDownshift>` - highlight first item when searching for options. Fixes STCOM-1015
+* Browse contributors. Placeholder match does not look right in the contributor column. Refs UIIN-2079
```
BogdanDenis, line 7: Please use ticket number from stripes-components Jira project

lib/MultiColumnList/MCLRenderer.js
```diff
@@ -1404,7 +1404,7 @@ class MCLRenderer extends React.Component {
             role="gridcell"
             key={`${col}-${rowIndex}-${this.keyId}`}
             className={`${css.mclCell} ${showOverflow} ${cellStyleClass}`}
-            style={cellStyle}
+            style={rowData.isAnchor && !!value ? { ...cellStyle, overflow: 'visible' } : cellStyle}
```
BogdanDenis, line 1407: Please move this conditional property to around line 1387, there's already some logic that builds `cellStyle` object and I think it'll be good to group it together

BogdanDenis (changes requested):

PR #1851, opened 2022-06-29
STCOM-647 Exposed refs in Selection and Select components

BogdanDenis:
## Description
Expose Ref to control in `<Selection>` and `<Select>`
`<Select>` already provides `inputRef` prop that accepts both `object` and `func` types - so no changes made there
Added a new `innerRef` prop to `<Selection>` that will reference filter input element. 

Fixed a `<DateRangeWrapper>` test - disabled dates are 1st to 9th, but the `-1` in `for` loop checked that disabled dates are 1st to 8th

## Issues
[STCOM-647](https://issues.folio.org/browse/STCOM-647)


JohnC-80 (changes requested):
There may be a use-case for getting a ref to the filter input element, but the more prominent case here is exposing a ref to the rendered control (the button, set up for rendering around line 707 of SingleSelect.js, reffed by the internal `this.control` variable ) so that proper focus management can be performed at the App level.

BogdanDenis:
> There may be a use-case for getting a ref to the filter input element, but the more prominent case here is exposing a ref to the rendered control (the button, set up for rendering around line 707 of SingleSelect.js, reffed by the internal `this.control` variable ) so that proper focus management can be performed at the App level.

Thanks @JohnC-80, I updated the ref to point to the control button
Also, do `<DateRangeWrapper>` tests pass locally for you? I had a failed test locally with up-to-date master and dependencies, so I "fixed" it, but now it fails in actions with opposite expected result

Svitlana-Korol (approved):

JohnC-80:
@BogdanDenis The DateRangeWrapper tests do pass locally for me (and in CI.) This is a strange discrepancy between the standard DateRangeWrapper tests and redux-form DateRangeWrapper tests that I have yet to dig to the bottom of. We might need to skip it for now.

lib/Selection/Selection.js
```diff
@@ -7,6 +7,7 @@ import SingleSelect from './SingleSelect';
 const propTypes = {
   emptyMessage: PropTypes.node,
   formatter: PropTypes.func,
+  innerRef: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
```
JohnC-80, line 10: We should name this `inputRef` to follow the convention of other form controls like TextField, Select, etc. Sorry for a late mention of this. Thank you!

JohnC-80 (changes requested):

JohnC-80 (approved):

ODanyor (approved):

PR #1855, opened 2022-06-30
STCOM-991 Add transpilation

mkuklis:


webpack.transpile.config.js
```diff
@@ -0,0 +1,52 @@
...
+  entry: path.resolve('./index.js'),
+  output: {
+    library: {
+      type: 'umd',
+      name: '@folio/stripes-components',
+    },
+    path: path.resolve('./dist'),
+    filename: 'index.js',
+  },
+  externals: {
```
mkuklis, line 13: I think we can probably move these common 3rd party externals to the default transpilation config in stripes-webpack so we don't repeat them in every module.

PR #1861, opened 2022-07-11
STCOM-1025 include static docs in storybook.

JohnC-80:
This PR is a first step in revamping FOLIO front-end documentation through storybook. It unifies the format of all component stories to the latest and here-forward style for storybook and includes the static documentation for accessibility, styling, etc. More updates and additions to come.

Using the .mdx format for documentation will allow us to render react components right inline with the markdown. The .format will still render normally through github, although without the active rendering of components.

Conversion of stories was performed via a code-mod, hence a whole lot of change - all stories aside from the Button story were converted (Button was already formatted, which is why it was rendering at the END of the list, rather than in the usual alphabetical order)

We can add new 'root' headings along with 'Guides' and 'Components' as needed.

image


ncovercash (approved):
Love this!

mkuklis (approved):
❤️

PR #1866, opened 2022-07-13
STCOM-1027 Tooltip flakiness.

JohnC-80:
Seeing some random failures with the tooltip tests... I suspect the setTimeout - isn't that always the case?

Approach - the escape key press to close *can happen before the component's open state is set due to the setTimeout. It's a little awkward to say 'when tooltip is visible, trigger the escape key on this *other element'.

This PR keeps the keyboard listener added while the trigger is mounted and if the mentioned timing happens, the timeout variable is cleared, regardless of whether the component is in the open state or not...

Bonus: eliminating a whole lifecycle declaration.


mkuklis (approved):
Nice cleanup @JohnC-80!

zburke (approved):

PR #1868, opened 2022-07-15
STCOM-977 Aria-labeling MultiSelection

JohnC-80:
This PR adds the `ariaLabel` support to Multiselect. (`aria-label` also works). This will provide an alternative accessible labeling mechanism for when visible labels don't suite requirements.

### DIFF
--- a/lib/MultiSelection/stories/HiddenLabel.js
+++ b/lib/MultiSelection/stories/HiddenLabel.js
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import MultiSelection from '../MultiSelection';
+import Headline from '../../Headline';
 
 const optionList = [
   { value: 'test0', label: 'Option 0' },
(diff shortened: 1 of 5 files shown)


lib/MultiSelection/stories/HiddenLabel.js
```diff
@@ -0,0 +1,31 @@
...
+
+const HiddenLabel = () => {
+  const [values, setValues] = useState([]);
+
+  return (
+    <div>
+      <Headline size="large">
+        Multiselect with no visible label
+      </Headline>
+      <p>The "ariaLabel" prop is used to apply a label that isn't visible to the user, but accessible to screenreaders. This is preferable for "required" fields over "ariaLabelledby" since the labeling elements won't need to display an asterisk.</p>
```
zburke, line 19: Gotta either obey lint or add `lib/**/stories/**` to the `.eslintrc` overrides.

zburke (changes requested):
**Comment:** Sorry, it isn't clear to me how these changes resolve the problems described in the ticket ("axe complains about missing roles..." but there are no "role" changes here). Likewise, one of the stories describes using `ariaLabel` for "required" fields that "won't need to display an asterisk" which I don't understand since the label for a required field _should_ contain an asterisk. Can we either update the Jira, or update the PR description, or update the storybook stories to more clearly indicate when `ariaLabel` should/should not be used, and how? 

**Request changes:** Lint please. I know it's just documentation, but lint will barf all over every future PR if we don't fix it here.

JohnC-80:
@zburke Apologies - some explanation is definitely required here. The necessary role change happened back in #1764 axe complained about the `div` back then - which is why the PR/JIRA cited in in STCOM-977 wasn't blocked on this issue - as the JIRA is titled "Ensure proper accessible labeling logic for Multiselect" - that's generally what this PR does.

At the time it was written, the role change was applicable, as it had not been done yet - I just never looked back.

In helping the devs resolve STSMACOM-448, after much go-round about labeling using accordion headers,  my initial answer was 'just use aria-label' - and then I felt boneheaded in the discovery that it didn't work properly on `MultiSelection`- so that was a no-go. I wasn't completely clear about the solution (or the *actual technical problem when I wrote the JIRA.) So it's far off in the description from the solution which ultimately ended up being 'make `aria-label` work with `Multiselection`

zburke (approved):
This is still a bit opaque to me if I'm being totally honest, but I'm willing to accept that it amounts to "`aria-label` now works the way it's supposed to".

PR #1869, opened 2022-07-15
Rename docs folder

JohnC-80:
New updates to our documentation include the .mdx files from the folder formerly known as "docs" in our storybook build. The docs folder is the one that storybook writes its files to in the `storybook` branch and, as part of that process, nukes the contents of the docs folder, causing the newly included stories to be missing.


zburke (changes requested):
Personally, I don't think you need to rewrite the CHANGELOG history but I don't have strong feelings about that. 

Looks like maybe you did `s/\(docs/guides/` instead of `s/\(docs/\(guides/`? It's missing across the board.

JohnC-80:
@zburke It ain't never as easy as the autopilot I'm apparently running on would have it. Thanks for the catch.

zburke (approved):

PR #1874, opened 2022-07-27
STCOM-1032 - Extend Modal and ModalFooter Components

Terala-Priyanka:
## Purpose
Extend Modal and ModalFooter Components
Issue: [STCOM-1032](https://issues.folio.org/browse/STCOM-1032)

### DIFF
--- a/lib/Modal/Modal.js
+++ b/lib/Modal/Modal.js
@@ -38,6 +38,8 @@ const propTypes = {
    * callback does not actually close the modal.
    */
   onClose: PropTypes.func,
+  /** Modal can be dismissed by clicking the background overlay */
+  onEscapeKeyDown: PropTypes.bool,
   /** Callback triggered when modal opens */
   onOpen: PropTypes.func,
   /** Size defaults to 60% the scope element's width.
@@ -66,6 +68,7 @@ const defaultProps = {
   closeOnBackgroundClick: false,
   dismissible: false,
   enforceFocus: true,
+  onEscapeKeyDown: false,
   onOpen: () => { },
   open: false,
   restoreFocus: true,
@@ -85,6 +88,7 @@ const Modal = forwardRef((props, ref) => {
     footer,
     id,
     onClose,
+    onEscapeKeyDown,
     onOpen,
     open,
     restoreFocus,
@@ -146,6 +150,7 @@ const Modal = forwardRef((props, ref) => {
                 restoreFocus={restoreFocus}
                 renderBackdrop={renderBackdrop}
                 aria-label={rest['aria-label'] || props.ariaLabel}
+                onEscapeKeyDown={onEscapeKeyDown ? onClose : () => { }}
               >
                 <WrappingElement
                   className={getModalClass()}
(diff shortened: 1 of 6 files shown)


JohnC-80 (changes requested):
@Terala-Priyanka ,The arrangement of modal buttons is by convention as stated here: https://ux.folio.org/docs/guidelines/language-rules/confirmation-message-language/
What is the use case for having a reversed arrangement?

For the escape key functionality in Modals, escape is a pretty safe convention, and I'm not sure why we'd need an extra prop, save for a more direct mapping to the `react-overlays` modal. 
Seems like it would be better for the `dismissible` prop to enable/allow this functionality and implement the key handler internally rather then exposing the prop on the outside, feeding it to the react-overlays Modal instance (OverlayModal). I'm not sure of a use-case where 'Dismissible' close button and close-on-escape functionality to be exclusive from eachother.

Terala-Priyanka:
@JohnC-80  I understand your point.

This requirement/ticket has been created in the interest of ticket https://issues.folio.org/browse/UIU-2653

In brief,
1. The ticket has few requirements that deal with user notes pop up that is implemented with [NotePopupModal ](https://github.com/folio-org/stripes-smart-components/blob/master/lib/Notes/NotePopupModal/NotePopupModal.js) component in [userDetail.js](https://github.com/folio-org/ui-users/blob/master/src/views/UserDetail/UserDetail.js)
2. NotePopupModal uses Modal and Modal footer components from stripes-components.
3. Modal Footer from NotePopupModal looks like below - 
image
The ticket [UIU-2653](https://issues.folio.org/browse/UIU-2653) needs the positions of the buttons to be exchanged, for which I need the alignment to be reversed. Hence this approach.
I agree a prop for reversed alignment may not sound great. 
If this approach sounds good...do you agree to replace the "reversed alignment" from with styles prop?
4. The ticket [UIU-2653](https://issues.folio.org/browse/UIU-2653) also needs to close the modal on "escape key" hit i.e., close-on-escape functionality. The mock provided in the ticket do not have the dismissible icon button. Hence I had to expose this extra prop.
5. Just as a side note, 
image

JohnC-80:
@Terala-Priyanka thank you for the explanation. With the logic in the NotesModal and the optional delete button, it would be better to just switch those buttons in the JSX within that modal... if the `hideDelete` variable is true, the close button would switch positions anyhow. It's best for that variability to not affect the layout in this way. I understand some awkwardness to the buttons rendering in reverse order, but It's best for the rendering order/styles to remain internal to stripes-components so that if we opt to change the global convention, the amount we'd have to change is smaller. I'd accept a PR in `stripes-smart-components` for this.

lib/Modal/Modal.js
```diff
@@ -146,6 +150,7 @@ const Modal = forwardRef((props, ref) => {
                 restoreFocus={restoreFocus}
                 renderBackdrop={renderBackdrop}
                 aria-label={rest['aria-label'] || props.ariaLabel}
+                onEscapeKeyDown={onEscapeKeyDown ? onClose : () => { }}
```
JohnC-80, line 153: After some accessibility discussion, I think `onClose` should always be called for this... whether it's hooked to the close handler or wired into a footer button - So I don't think we need to add another prop for this functionality.

JohnC-80 (changes requested):

lib/Modal/Modal.js
Terala-Priyanka: @JohnC-80 
Thank you for your review. I will be closing this PR and include all the above discussed scope in https://issues.folio.org/browse/STSMACOM-688

PR #1877, opened 2022-07-29
STCOM-1033 - SearchField - expose name prop for index field

JohnC-80:
SearchField is currently unable to apply a `name` attribute to its rendered `<Select>` for Index - The `name` attribute is necessary for generic input handlers (as well as form libraries) to work properly.

SASQ has a generic handler for its `SearchHandlers().query` which will check the `name` attribute of the event target to make updates accordingly.


zburke (approved):
I have to be honest, the description here is pretty opaque to me. IIUC, SearchField ends up rendering multiple elements to the DOM and the `name` prop is used for the actual text input field. Thus, `indexField` is used to hold the `name` attribute that will be applied to the `Select` field that holds the list of indexes when multiple indexes are present.

PR #1832, opened 2022-06-19
STCOM-1017: Fix 12-hour formatting in getLocalizedTimeFormatInfo

ncovercash:
# [Jira](https://issues.folio.org/projects/STCOM/issues/STCOM-1017)

The function `getLocalizedTimeFormatInfo` in `dateTimeUtils` always returns a format string with "HH", indicating a format with hours from 00-23, regardless of if it adds AM or PM.

This fixes this, changing `HH` to `hh` (for 1-12 hours) whenever `A` is present in the format string.

PR #1881, opened 2022-08-14
Update getLocalizedTimeFormatInfo docs

ncovercash:
For consistency with #1832/#1837.  This reflects that the format code `hh` should be used for 12-hour time, rather than `HH`, whenever `A` is present.


JohnC-80 (approved):

JohnC-80:
Merged in #1892

PR #1895, opened 2022-09-20
STCOM-1041 A11y labeling, focus management for modals.

JohnC-80:
[JIRA](https://issues.folio.org/browse/STCOM-1041)

Users requested that focus be automatically placed on the 'confirm' button of `ConfirmationModal`s. This moves focus beyond read-only content that probably contains important information. The solution is to include the message in the labeling of the modal so that it will automatically be announced when focus enters the modal. This will have it behave in a similar way to Javascript alerts.

Other explored approaches: `aria-describedby` - this attribute has varying degrees of functionality between browsers. It would be nice if this just worked in JAWS, but the current `aria-labelledby`-based implementation will have to do for now.

`aria-live` region - this rather-forceful approach always created a double announcement somewhere. Since it's element-based, there's not a straightforward place to actually put it since both the heading of the modal and the content would need announcement.

There is a standard 'dialog' announcement that happens on NVDA, but not in JAWS, so we unfortunately get a double announcement there. A double is flawed, but better than 'none'

This PR includes an example of modal focus-management in documentation as well as functionality to properly set the accessible label attributes of the modal. If both an `aria-label` and `aria-labelledby` prop are applied, `aria-labelledby` is applied in favor since `aria-label` will not be announced (lower priority that `aria-labelledby`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Export `staticFirstWeekDay` and `staticLangCountryCodes` from `Datepicker`. Refs STCOM-1038.
 * Button: Button link style has a min-height, which can offset it from text. Fixes STCOM-1039.
 * The vertical scroll bar displays at the second pane when it doesn't need. Fixes STCOM-1044.
+* Focus management and accessible labeling of confirmation modals. Confirmation modals announce in a way similart to Javascript alerts. Fixes STCOM-1041.
 
 ## [10.2.0](https://github.com/folio-org/stripes-components/tree/v10.2.0) (2022-06-14)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v10.1.0...v10.2.0)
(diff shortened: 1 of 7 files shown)


zburke (commented):
Wow, I learned so much from the description here! I'm so impressed with your knowledge of the details, and, ugh, so anguished by the conflicts in how different browsers handle different attributes in the a11y space. It's sad really. 

The only question I have here is whether removing the `ariaLabel` prop from `<ConfirmationModal>` constitutes a breaking change. I think it does. Granted, using the modal's content as the label is clearly better, but if we're gonna sneak this change into a minor release we need to advertise it, or we just need to wait until the next major release.

JohnC-80:
@zburke. Let me revisit a bit and see if we can mitigate in a less-semver-breaky way... this prop is very low-usage(if any), but it's there.

JohnC-80:
Updated. I can write up a follow-up ticket to remove all the deprecation/compatibility code. For the most, it just keeps test from breaking if there are any... also, the inclusion of `ariaLabel` in the children will mean the label isn't completely discarded if it's trumped by `ariaLabelledBy` or its kebab-cased sister.

zburke (approved):
LGTM. I appreciate the non-breaking way you handled this, including adding comments and deprecated-props warnings that will surface these changes to folks doing things in an outdated way. 

You even snuck past the lint troll! Nice work.

PR #1896, opened 2022-09-27
STCOM-1046: Make MCL row unclickable when a placeholder row is displayed

odoniyor:
## Purpose
Make MCL row unclickable when a placeholder row is displayed.

## Approach
Remove the `onClick` callback if `rowData` doesn't have an `id`.

## Issues
[STCOM-1046](https://issues.folio.org/browse/STCOM-1046)

### DIFF
--- a/lib/MultiColumnList/defaultRowFormatter.js
+++ b/lib/MultiColumnList/defaultRowFormatter.js
@@ -48,6 +48,10 @@ function defaultRowFormatter(row) {
 
   delete rowProps.labelStrings; // We don't want to spread this onto the DOM element.
 
+  if (!rowData.id) {
+    delete rowProps.onClick;
+  }
+
   return (
     <Element
       key={`row-${rowIndex}`}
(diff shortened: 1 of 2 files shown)


JohnC-80 (changes requested):
This is very dependent on the shape of the data received by the component and the case seems very specific to 'Browse' functionality.
Row-level clicks should be avoided, if possible, and the link rendered within the cell should be used for navigation rather than the entire row. The link can just be conditionally rendered (which I think is already happening, based on the screenshot in the JIRA)
We've kept the row-click functionality for sake of legacy support, but more accessible patterns have been implemented in [ui-users](https://github.com/folio-org/ui-users/pull/2003/files) and [ui-requests](https://github.com/folio-org/ui-requests/pull/849)

lib/MultiColumnList/defaultRowFormatter.js
```diff
@@ -48,6 +48,10 @@ function defaultRowFormatter(row) {
 
   delete rowProps.labelStrings; // We don't want to spread this onto the DOM element.
 
+  if (!rowData.id) {
+    delete rowProps.onClick;
+  }
+
```
Dmytro-Melnyshyn, line 54: It is possible to make a row unclickable by not passing the `onRowClick` prop to the `MultiColumnList`. There is no requirement that something should happen when a row is clicked.

odoniyor: Yeah, the solution isn't good here and I don't think it's possible to bypass the `onRowClick` callback for a specific column row.

PR #1904, opened 2022-11-01
STCOM-1054: The "Are you sure?" modal window doesn't close when pressing on the "ESC" keyboard button

Dmytro-Melnyshyn:
## Purpose
The `ConfirmationModal` has to be closed after clicking the `ESC` key.

## Approach
Use the already existing `onHide` cb by passing the `onClose` prop to `Modal` component. `onHide` also fires when `ESC` is pressed ([doc](https://react-bootstrap.github.io/react-overlays/api/Modal#onHide)).

There is another way - to use `onEscapeKeyDown` cb, but then we will need to prevent `onHide` from being called so that `onClose` is not called twice ([doc](https://github.com/react-bootstrap/react-overlays/blob/master/src/Modal.tsx#L291)).

## Issues
[STCOM-1054](https://issues.folio.org/browse/STCOM-1054)

## Screencast

https://user-images.githubusercontent.com/77053927/199209494-f6b6e466-fe12-483a-b9b2-4f03a7f0df19.mp4


zburke (approved):
Thanks for explaining the reasoning behind this in your description! This seems right to me (and simpler!), but I would really like to get @JohnC-80's opinion before we merge this.

JohnC-80 (approved):

PR #1919, opened 2022-12-05
STCOM-1059 Login screen missing scrollbar.

JohnC-80:
This was happening due to a global style on the body element with `overflow: hidden`

These styles were in use 5 years ago, prior to using flexbox for the outer DOM hierarchy (root, nav, module-container) when there was some risk of introducing a new hierarchical 'layer' causing overflow to occur on the body element and a vertical scrollbar to appear.

Approach: remove the offending global style. The `root` element is rendered as a flex column with its height at `100%` - removing this flex *will make a browser scrollbar appear, but since it's there, the `root` does not become taller than the viewport, which renders the removal of body overflow as unnecessary.


mkuklis (approved):

zburke (commented):
At a glance this looks fine, your explanation makes sense, and tests continue to pass. Good! But ... have you confirmed that applications continue to render as expected? Even if it's just spot-testing, it still seems like it could be a good idea. Forgive me; I'm a bit nervous changing anything that so deliberately attaches to the global scope!

zburke (approved):
I did a bit of spot-testing with ui-users and things look OK.

PR #1924, opened 2022-12-13
STCOM-1093 MCL- ColumnWidths should update when the WidthHint API is used.

JohnC-80:
Columns with width hints were previously marked as undefined when the values updated, so the MCL's columns would not re-render unless data was updated (causing measurement caches to be reset)

Approach - when hinted columnWidths are updated, this PR will reset those widths in state and trigger a re-measure of the DOM elements.

Refs [STCOM-1093](https://issues.folio.org/browse/STCOM-1093)


zburke:
Ping @EthanFreestone. We need to get this work merged in the next few days if we want it in the 2023-R1/Orchid release.

PR #1926, opened 2022-12-15
STCOM-1067 remove deprecated properties and components

zburke:
Many things have long been labeled as deprecated but could not be removed until a breaking release. The next release will be a breaking change, so we're finally clearing out the cruft:

* `UncontrolledDropdown` will be removed
* `AutoSuggest` will no longer accept `tether`
* `Card` will no longer accept `hasMargin`
* `ConfirmationModal` will no longer accept `ariaLabel`
* `Headline` will no longer accept `bold`
* `Modal` will no longer accept `ariaLabel`
* `MultiColumnList` will no longer accept `paginationBoundaries`
* `Popover` will no longer accept `alignment` or `position`
* `Selection` will no longer accept `tether`
* `TextField` will no longer provide `getInput()` or `focusInput()` (Affects cases where a `ref` may be used to access these methods)
* `TimePicker` will no longer accept `passThroughValue`

Refs [STCOM-1067](https://issues.folio.org/browse/STCOM-1067)


zburke:
@JohnC-80, admittedly I just bashed through this and there are probably some doc and test updates to add. The biggest surgery I think was in Timepicker where removing a prop allowed for quite a bit of code to be excised. If/when this looks good, do you think we should announce it on Slack#stripes-updates to encourage people to try running their apps against this branch before we commit it?

JohnC-80:
@zburke thanks! Yes - I think we should definitely announce this branch on stripes-update.

lib/MultiColumnList/PrevNextPaginationRow.js
```diff
@@ -1,19 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { deprecated } from 'prop-types-extra';
```
zburke, line 3: I don't see this being used, @JohnC-80. Can you explain?

JohnC-80: Hasty commit - I probably pulled in the version from master when fixing the test failure here. Fixed now.

JohnC-80 (approved):

PR #1933, opened 2022-12-21
STCOM-1069: Add 'focusIndex' property to be able to manually control the focus of an element in the list.

Dmytro-Melnyshyn:
## Description
The story for which this is being done is [UIPFPAT-47](https://issues.folio.org/browse/UIPFPAT-47).

The focus doesn't go to the first item in the list after clicking the `Previous`/`Next` button. This is because the list can have a different number of items for each page, for example, there are 50 items on page 1, and 100 on the second. 

This was solved by adding the `focusIndex` prop to be able to manually pass the index of the item to be focused.

## Issues
[STCOM-1069](https://issues.folio.org/browse/STCOM-1069)

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -188,6 +188,7 @@ class MCLRenderer extends React.Component {
     containerRef: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
     contentData: PropTypes.arrayOf(PropTypes.object),
     dataEndReached: PropTypes.bool,
+    focusIndex: PropTypes.string,
     formatter: PropTypes.object,
     getCellClass: PropTypes.func,
     getHeaderCellClass: PropTypes.func,
@@ -260,6 +261,7 @@ class MCLRenderer extends React.Component {
     columnWidths: {},
     contentData: [],
     dataEndReached: false,
+    focusIndex: '',
     formatter: {},
     hasMargin: false,
     hotKeys: { keyMap: {}, handlers: {} },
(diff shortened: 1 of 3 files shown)


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -547,9 +549,9 @@ class MCLRenderer extends React.Component {
       }
     } else if (pagingType === pagingTypes.LOAD_MORE || pagingType === pagingTypes.PREV_NEXT) {
       // this.focusTargetIndex can be 0, so we need to check for undefined
-      if (this.focusTargetIndex !== undefined) {
+      if (focusIndex || this.focusTargetIndex !== undefined) {
```
BogdanDenis, line 552: I think we can actually use `dataStartIndex` from state. Adding a new prop means that focus index is controlled from both the outside of component via props and inside via `setFocusIndex` function, and this like this can get really messy
But using component's state should help us avoid it

BogdanDenis (changes requested):

JohnC-80 (changes requested):
I think that the best way to do this is going to be to expose a ref to the `rowContainer` and use application logic along with our [focusManagement API](https://github.com/folio-org/stripes-components/blob/master/util/getFocusableElements.js) to control/place focus wherever you want to put it, and - most importantly - control when you want it to go there instead of relying on prop/state changes or state within this component to handle it.
Pagination is a very specific scenario with internal focus-management - and the instant focus goes to that focusIndex, the focus index is reset... you can see how it's handled internally here -  
https://github.com/folio-org/stripes-components/blob/master/lib/MultiColumnList/MCLRenderer.js#L550-L558

Dmytro-Melnyshyn:
Thanks for your review, it was really helpful. I've found a way to not change anything here. Just use `containerRef` and focus on the element I need.

PR #1957, opened 2023-01-18
STCOM-1080 Update react-redux to v8

ryandberger:
- Refs https://issues.folio.org/browse/STCOM-1080.
- Upgrade will be happening across all of Stripes framework and modules that reference react-redux.
- v7 to v8 is a very harmless upgrade and is only considered breaking since they removed a few rarely used functions (which I confirmed we don't use). See https://github.com/reduxjs/react-redux/releases/tag/v8.0.0 for more details.


JohnC-80 (approved):
`devDepenedency` here for the test harness - notably used for redux-form/final-form test scenarios. Tests pass, so seems alright to me.

zburke (approved):

mkuklis (approved):

PR #1960, opened 2023-01-20
STCOM-1085 Proptype validation issue

JohnC-80:
PaneResizeContainer: parentElement prop to instanceOf(Element)


zburke (approved):
❤️ :shipit:

PR #1965, opened 2023-01-24
fix tyops in lib/Icon/readme

zburke:


zburke:
@MikeTaylor types faster than me; see #1964.

PR #1967, opened 2023-01-26
STCOM-1097: Add line break at newline characters in KeyValue component

OleksandrHladchenko1:
## Links
[STCOM-1097](https://issues.folio.org/browse/STCOM-1097)
[UIDATIMP-1342](https://issues.folio.org/browse/UIDATIMP-1342)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Replace tyop in `<Accordion>`'s test-id. Refs STCOM-1079.
 * Upgrade `react-redux` to `v8`. Refs STCOM-1080.
 * Fix bug with MCL columnwidths not updating with width hint API. Refs STCOM-1090.
+* Add line break at newline characters. Refs STCOM-1097.
 
 ## [10.3.0](https://github.com/folio-org/stripes-components/tree/v10.3.0) (2022-10-13)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v10.2.0...v10.3.0)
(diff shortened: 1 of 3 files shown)


zburke:
@OleksandrHladchenko1, do we have an example of the display problem this PR resolves? Sorry, it isn't clearly articulated in any of the linked tickets.

JohnC-80:
@OleksandrHladchenko1 perhaps just a before/after screenshot with the change in place? I think I understand they want to render multiline values - like addresses - correctly. Thank you!

OleksandrHladchenko1:
> 

Hi @zburke I think the problem is displayed in this PR https://github.com/folio-org/ui-data-import/pull/1314.

OleksandrHladchenko1:
> @OleksandrHladchenko1 perhaps just a before/after screenshot with the change in place? I think I understand they want to render multiline values - like addresses - correctly. Thank you!

Hi @JohnC-80 We get the address from backend as `"Acquisitions Staff Accountant \nTexas A&M University Libraries \n5000 TAMU \nCollege Station, TX   77843-5000"` and we need to display this address as
image

When I added `white-space: pre-wrap` to the `KeyValue` component, it started to see newline breaks and render the address as expected

>

JohnC-80 (approved):

zburke (changes requested):
Gotcha; thanks for the clarification. The code change is fine, but we need to update `readme.md` with a note that whitespace in the `value` prop will be honored since this is 100% contrary to how it ordinarily works. Breaking the rules is fine, as long as you do it knowingly and loudly.

zburke (approved):
LGTM! Thank you. Sorry; I know that was pedantic of me

PR #1968, opened 2023-01-27
STCOM-1099 Refactor DateRangeWrapper

BogdanDenis:
## Description
Create a export reports modal for MARC Authorities app

### `<DateRangeWrapper>`
`stripes-components` has what looks like a perfect component for this functionality - `<DateRangeWrapper>` that contains logic and validation for start and end date inputs
The issue is that it's been originally written to work with `react-redux`. Now that we're moving to use `react-final-form` the component can be refactored to use it instead
Searching `platform-complete`, I didn't find any uses of the component, so we don't need backwards compatibility

Changes include:
- rewrite to be a functional component
- instead of using `onChange` to get field values - use `useFormState` hook

## Issues
[STCOM-1099](https://issues.folio.org/browse/STCOM-1099)


JohnC-80 (changes requested):
Despite the apparent lack of use this component has, this is still a breaking change. I suggest we place these in a separate component - "DateRangeFieldset" is a possible name. We can plan for deprecation/removal of the existing DateRangeWrapper in a future breaking Stripes release.

Please try and use the `usePortal` prop with Datepicker to resolve the `overflow` problem. If it cannot be resolved this way and the `Modal` adjustment is necessary, that should be moved to a separate PR. For the worst case, it's best to keep them separate - we wouldn't want reverting one change to also revert the other, as we do like to squash commits and merge these once they're accepted.

BogdanDenis:
Hey @zburke would it be possible to release this new component that John has suggested as part of Orchid release?

BogdanDenis:
Will create a copy of `DateRangeFieldset` locally in `ui-marc-authorities` for Orchid release and after that will decide on how to add it to `stripes-components`

PR #1976, opened 2023-01-31
STCOM-1101 Add `rootClass` and `fitContent` props to `<TextArea>`

BogdanDenis:
## Description
Add `rootClass` prop to be able to define custom styles for `<TextArea>` wrapper element
Add `fitContent` prop to be able to set initial width of `<TextArea>` to fit content

## Screenshots

https://user-images.githubusercontent.com/19309423/215756427-29c7e504-0572-4010-b9b6-fb439c73b15e.mp4

## Issues
[STCOM-1101](https://issues.folio.org/browse/STCOM-1101)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-components
 
+## [11.1.0](IN PROGRESS)
+
+* Add `rootClass` and `fitContent` props to `<TextArea>`. Refs STCOM-1101.
+
 ## [11.0.0](https://github.com/folio-org/stripes-components/tree/v11.0.0) (2023-01-30)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v10.3.0...v11.0.0)
 
--- a/lib/TextArea/TextArea.js
+++ b/lib/TextArea/TextArea.js
@@ -18,7 +18,13 @@ class TextArea extends Component {
     disabled: PropTypes.bool,
     endControl: PropTypes.element,
     error: PropTypes.node,
+    /**
+     * Will resize the textarea to resize to show all it's content
+     */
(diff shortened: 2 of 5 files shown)


zburke (changes requested):
I understand we don't already have a README.md for this component, but we do show several different options in the storybook in `stories/BasicUsage.js` so let's at least add an example there instead of digging the no-documentation hole any deeper. It would be great to write a quick README too, but I understand that's asking a lot given the small modifications you're making to this big component. 

As to the new props: it isn't immediately obvious how `fitContent` will differ from `fullWidth` so please make sure to explain this. Maybe it'll be obvious in the storybook UI, but just reading the description they sound similar. 

New props should be covered by new tests as well, i.e. making sure that the desired attributes are present in the DOM when the component receives these new props.

zburke:
@BogdanDenis, FYI I filed https://issues.folio.org/browse/STCOM-1102 for adding a `README.md` since it's unfair to spring that requirement on this ticket.

BogdanDenis:
Hey @zburke thanks for your comments - I've added some details on difference between `fitContent` and `fullWidth` in prop types declaration, as well as storybook examples and tests
Also, there's a change in `stripes-testing` that we need for tests https://github.com/folio-org/stripes-testing/pull/820

zburke (changes requested):
Thanks for the documentation and tests! Given this relies on a new interactor in stripes-testing, that means we need to bump the minimum version of `@folio/stripes-testing` in the dev-deps to the version that includes that interactor. It looks like that'll be 4.5.0.

zburke (approved):

Dmytro-Melnyshyn (approved):

JohnC-80 (approved):

PR #1979, opened 2023-02-01
STCOM-1103 Timepicker too aggressive in formatting user input.

JohnC-80:
https://issues.folio.org/browse/STCOM-1103

Approach - removed the range of formats for checking validity of value. The component will only format once a full string matching the `H:MM` or `H:MM A` format is entered (corresponding with the time format of the ui locale.)

### DIFF
--- a/lib/Timepicker/Timepicker.js
+++ b/lib/Timepicker/Timepicker.js
@@ -26,6 +26,8 @@ const containsUTCOffset = (value) => {
   return offsetRegex.test(value) || offsetRE2.test(value) || offsetZRE.test(value);
 };
 
+// Not all parameters may be used in this function, but they are provided as a convenience for
+// overrides to the props using this as default.
 const defaultParser = (value, timezone, timeFormat, intl) => { // eslint-disable-line no-unused-vars
   if (!value) return '';
   let time;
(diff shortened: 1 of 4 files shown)


ncovercash:
@JohnC-80 I've cloned this branch locally, but I'm still getting the error when the field is controlled.

For controlling, I'm just storing a value and throwing it back:
```ts
  const [v, setV] = useState();
  return <Timepicker value={v} onChange={(e, newV) => setV(newV)} />;
```

ncovercash (approved):
Looks great, thanks!

lib/Timepicker/Timepicker.js
```diff
@@ -26,7 +26,7 @@ const containsUTCOffset = (value) => {
   return offsetRegex.test(value) || offsetRE2.test(value) || offsetZRE.test(value);
 };
 
-const defaultParser = (value, timezone, timeFormat, intl) => {
+const defaultParser = (value, timezone, timeFormat, intl) => { // eslint-disable-line no-unused-vars
```
zburke, line 29: Please explain why this function accepts `intl` if we're not using it. I know it seems lame to have a comment to explain a comment, but it would actually be helpful here :)

lib/Timepicker/Timepicker.js
```diff
@@ -42,7 +42,7 @@ const defaultParser = (value, timezone, timeFormat, intl) => {
   return time.isValid() ? removeDST(time.toISOString(), timeFormat) : value;
 };
 
-const defaultOutputFormatter = ({ value, formats, timezone, intl }) => {
+const defaultOutputFormatter = ({ value, formats, timezone, intl }) => { // eslint-disable-line no-unused-vars
```
zburke, line 45: Please explain why this function accepts `timezone` and `intl` if we're not using them. I know it seems lame to have a comment to explain a comment, but it would actually be helpful here :)

zburke (approved):
LGTM

PR #1980, opened 2023-02-01
STCOM-1104 Timedropdown spinners not respecting ranges.

JohnC-80:
https://issues.folio.org/browse/STCOM-1104

Currently, the time values can go into negative or well beyond the 59 minute mark...

The Timepicker spinners should function within the ranges of their particular hour format, with a range of 0 - 59 for minutes and 0-23 for 24 hour format , 1-12 in 12 hour format.

This PR focuses on that logic to ensure a value within the ranges is set.
The main issue was a mistake in trying (and failing) to mutate a value.
Tests are added for both 12 and 24 hour modes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 * Add `rootClass` and `fitContent` props to `<TextArea>`. Refs STCOM-1101.
 * Fix bug with Timepicker formatting user input too quickly/aggressively. Refs STCOM-1103.
-
+* Fix big with Timepicker timedropdown spinners not respecting their appropriate ranges. Refs STCOM-1104.
+ 
 ## [11.0.0](https://github.com/folio-org/stripes-components/tree/v11.0.0) (2023-01-30)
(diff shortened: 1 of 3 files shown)


ncovercash:
Should we permit users to wrap the values around (for example, incrementing 59 minutes back around to 0)?

zburke (approved):
LGTM, though comments on `keepInRange` and `rangedIncrement` describing what they do and their args should would be nice. Their behavior is nicely covered by the new tests (thank you!!!) but it would be nice to have documentation in-situ rather than needing to read through the function or its tests to figure out what they do.

zburke:
@JohnC-80 : also please note the initial Orchid release is already published. If we need this fix in Orchid, please be sure to set the "Release" field accordingly and set the status to "Awaiting Release".

ncovercash:
> @JohnC-80 : also please note the initial Orchid release is already published. If we need this fix in Orchid, please be sure to set the "Release" field accordingly and set the status to "Awaiting Release".

for my 2¢, I'd really appreciate STCOM-1103 and this being fixed for Orchid

ncovercash (approved):

PR #1984, opened 2023-02-07
STCOM-1106 MultiSelection - onRemove should support remove button clicks and list de-selection.

JohnC-80:
Currently, it only supports backspace presses.

Approach: provide `onRemove` to internal `MultiDownshift` component so that it can be called when items are removed both via the value chip 'X' button and clicking the selected item within the list.


EthanFreestone:
Questions still remain from the ERM side I'm afraid. They won't stop the merge of this PR, but do perhaps change whether this fix will actually unblock us or not:
https://folio-project.slack.com/archives/C210UCHQ9/p1676038389044729?thread_ts=1675791423.960979&cid=C210UCHQ9

PR #1985, opened 2023-02-09
STCOM-1074: MCL - Axe error... 'required children' 

JohnC-80:
TLDR, MCL's that use links within the cells rather than row-level anchors perform fine with assistive technology. Axe-core has a couple of rules that seem to work in opposition to eachother given recent updates >=4.5.0. 

The axe rules themselves affecting this case are too generic and potentially wide sweeping to just turn off. `aria-required-children` checks for proper parent/child structure of elements with aria roles. It's a very important test to have for libraries that export complex widgets.

Maintainers of `axe-core` understand that their library [needs work with these detections](https://github.com/dequelabs/axe-core/issues/3888). There has been some work around the approach for `aria-required-children` but [suggesting an attribute such as `aria-busy`](https://github.com/dequelabs/axe-core/issues/3208) seems ill-advised if nothing is truly busy, despite that it stifles the axe errors - unfortunately, that's the path that *some developers may take since it's the shortest path to the goal of appeasing axe software.

Approach: conditionally render a tabindex on the scrollable container if it contains no focusable elements. The vast majority of our grids in FOLIO contain focusable elements, so this will  omit the `tabindex` attribute from those grids and calm those errors down. If there are any spots that persist, a check for the 'virtualize' prop may be necessary, but the fewer conditionals we can get away with, the better.

Elaborating on the problem, the WAI-ARIA spec includes DOM-descendents in their [definition of 'owned elements.'](https://www.w3.org/TR/wai-aria/#dfn-owned-element) By this definition, the structure in which MCL arranges its aria-roles of MCL (`grid` , `row`, `gridcell`) are acceptable and allow for screen readers to function normally. Role-less, non-interactive elements are ignored in these sort of hierarchy checks... but if so much as a `tabindex` is added to an element, it now causes a gap in the hierarchy (for axe's check) resulting in an error. Since there is another potential violation for [keyboard interaction with scrollable areas](https://dequeuniversity.com/rules/axe/4.3/scrollable-region-focusable?application=AxeChrome), [common](https://bvaughn.github.io/react-virtualized/#/components/Table) [virtualized components](https://ag-grid.com/examples/infinite-scrolling/simple/packages/vanilla/index.html) add a `tabindex`-ed element into their structures - and axe throws the error for them as well... If the `tabindex` on the scroll container is removed, then the error will go away *so long as there are focusable elements within the container - and that's the logic that this PR follows.

### DIFF
--- a/lib/MultiColumnList/PrevNextPaginationRow.js
+++ b/lib/MultiColumnList/PrevNextPaginationRow.js
@@ -90,10 +90,11 @@ const PrevNextPaginationRow = ({
             top: `${position}px`,
             marginTop: '1rem'
           }}
+          role="row"
         >
           <FormattedMessage id="stripes-components.mcl.itemsRequestedMessage">
             { ([message]) => (
-              <div className={css.mclPrevNextStickyContainer}>
+              <div className={css.mclPrevNextStickyContainer} role="gridcell">
                 <div className={css.mclPaginationCenterContainer}>
                   <div className={css.mclPrevNextButtonContainer}>
                     <PagingButton
--- a/lib/MultiColumnList/CenteredContainer.js
+++ b/lib/MultiColumnList/CenteredContainer.js
@@ -4,7 +4,7 @@ import { isFinite } from 'lodash';
 import Layout from '../Layout';
 import css from './MCLRenderer.css';
 
-const CenteredContainer = ({ innerRef, visible, width, children, style: styleProp }) => {
+const CenteredContainer = ({ innerRef, visible, width, children, role, style: styleProp }) => {
   // subtracting the margins to prevent horizontal scroll
   const endOfListWidth = isFinite(width) ? `${Math.max(width - 20, 200)}px` : '100%';
 
@@ -19,6 +19,7 @@ const CenteredContainer = ({ innerRef, visible, width, children, style: stylePro
           padding: `${visible ? null : 0}`,
           ...styleProp }
         }
+      role={role}
     >
       <Layout className="textCentered">
         {children}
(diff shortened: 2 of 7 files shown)


zburke (approved):
Wow, I'm astonished at the depth of your understanding of the problem, the tradeoffs among the different solutions, the thoughtful and clever you resolved this, and most of all the clear way you explained it all. My hat is off to you, sir. Nice work.

zburke:
@JohnC-80: this change breaks tests in ui-inventory because the previous/next container now has the `row` role, as do data containers. Thus queries like this
```
      expect(document.querySelectorAll('#pane-results-content .mclRowContainer > [role=row]').length).toEqual(3);
```
for an MCL with 3 rows of data now return 4 rows. Do you think the previous/next block belongs within the `.mclScrollable` and `mclRowContainer` containers or should it be outside, perhaps along with the `.mclEndOfListContainer`? I realize some of these classes may be about CSS, not about semantics, but I think we should have an easy way to separate the footer-cells of an MCL from the content cells, just as the header-cells are. If we need to adjust some tests as a result, that's OK, but right now I don't think we have a good way to distinguish the data-row cells from the footer-row cells.

lib/MultiColumnList/PrevNextPaginationRow.js
```diff
@@ -90,10 +90,11 @@ const PrevNextPaginationRow = ({
             top: `${position}px`,
             marginTop: '1rem'
           }}
+          role="row"
```
zburke, line 93: @JohnC-80: this change breaks tests in ui-inventory because the previous/next container now has the `row` role, as do data containers. Thus queries like this
```
expect(document.querySelectorAll('#pane-results-content .mclRowContainer > [role=row]').length).toEqual(3);
```
for an MCL with 3 rows of data now return 4 rows. Do you think the previous/next block belongs within the `.mclScrollable` and `mclRowContainer` containers or should it be outside, perhaps along with the `.mclEndOfListContainer`? I realize some of these classes may be about CSS, not about semantics, but I think we should have an easy way to separate the footer-cells of an MCL from the content cells, just as the header-cells are. If we need to adjust some tests as a result, that's OK, but right now I don't think we have a good way to distinguish the data-row cells from the footer-row cells.

PR #1986, opened 2023-02-13
STCOM-1114 remove roles from pagination elements

JohnC-80:
https://issues.folio.org/browse/STCOM-1114


zburke (approved):
LGTM! It's too bad we need to revert these changes from #1985, but I think it's the right call for now.

PR #1987, opened 2023-02-15
STCOM-1118 Timepicker dropdown not spinning values when value is empty.

JohnC-80:
Pre-refactor `<Timepicker>` would default to using the current time in its `TimeDrpodown` if the selected time value was empty. The refactor should behave in the same way, enabling spinner buttons to work.

Approach, modify the initial state of TimeDropdown for the case where the `selectedTime` prop is empty. The initial state will be equal to `moment()` with the applicable `format()` call depending on whether it's in the hour or minute field.


zburke (approved):
True confession: the changes in `handlerBlur` are a bit opaque to me, but the new initialization looks correct. Thank you!

PR #1991, opened 2023-02-21
STCOM-1120 Timepicker - Refactor convert 24hr function

JohnC-80:
Problem: When hours format is 12-hour, confirming a time (12+ PM) in the `<TimeDropdown>`, "Invalid Date" appears in the input. 

Approach - the "invalid date" message appeared as a result of faulty logic in a 24-hr conversion project for moment/date library parsing. The logic was revised to perform correctly, and tests written to uphold the behavior.


zburke (approved):
tests so pretty 😍 thanks for the clear explanation of the problem and for documentation in the form of tests that demonstrate how it works.

PR #1993, opened 2023-02-22
STCOM-1115 - Move pagination element outside of MCL's containing element.

JohnC-80:
Problem: MCL's built-in pagination does not have a proper "role" attribute for being placed directly within the scroll container with the other data rows. This causes errors in accessibility scanning.
Adding the attributes to the elements can cause tests to fail since they often check for the number of rows via the `[role=row]` selector.

To also conform this component to UX expectations, this will mean that pagination for most result listings will be pushed to the bottom of the page. Users will no longer have to scroll to reach the pagination 🎉 

This change also incurs an update to MCL's interactor in stripes-testing: https://github.com/folio-org/stripes-testing/pull/887


JohnC-80:
Closed in favor of a slightly lighter solution, #2125

PR #1995, opened 2023-03-03
STCOM-1105 - MCL sortable header focus styles.

JohnC-80:
No more blind pressing of the 'tab' key when focus reaches this part....

Refs [STCOM-1105](https://issues.folio.org/browse/STCOM-1105)

header-focus


zburke (approved):
LGTM. Is this in fact a bugfix since `inner` isn't actually a CSS attribute? (Or is it a postcss thing I just don't know about?)

JohnC-80:
Indeed... most definitely a defect - both on account of the CSS problem and the fact that nothing happens in the UI when something should most definitely happen.

PR #1996, opened 2023-03-03
STCOM-680 - fix MCL sorting via keyboard keys (enter, space click focused header)

JohnC-80:
Hand-in-hand with focus styles, appropriate keyboard behavior also applied. The focused column header can be clicked via 'Enter' key (keydown) or 'Spacebar' (keyup) (common button behavior)


zburke (approved):
`Enter` and ` ` are also way more clear than `13` and `32`, though of course a [Real Programmer](https://xkcd.com/378/) probably has those values memorized.

JohnC-80:
I admit, I was surprised to see that the 'key' value for spacebar is ` ` I expected something like `Space` or more programmerly&trade; like `spaceBar`

PR #1997, opened 2023-03-03
STCOM-1128 Handle timezone conversion with defaultOutputFormatter

JohnC-80:
Problem: time should convert to utc for backend values in respect to the provided timezone.

This PR includes a call to moment.tz in case the timezone is something besides `utc` and still uses `moment.utc()` for cases where it doesn't receive a timezone (rare) instead of always using utc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Fix Convert24hr function of timepicker to fix 'invalid date' message when timedropdown is used. Refs STCOM-1120.
 * Fix MCL Columnheaders' focus styling. Refs STCOM-1105.
 * Fix keyboard interaction with MCL Columnheaders - Enter and Spacebar can now be used to 'click' them. Refs STCOM-680.
+* Implement timeZone support in `<Timepicker/>` default output formatter. Refs STCOM-1128.
 
 ## [11.0.0](https://github.com/folio-org/stripes-components/tree/v11.0.0) (2023-01-30)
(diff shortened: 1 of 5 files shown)


zburke (commented):
I got tagged on the Jira related to this PR. I confess I'm still confused about all the things that are changing. Was the bug passing `timeZone` _to_ `outputFormatter` but destructuring `timezone` _in_ `outputFormatter`? Or was it failing to set moment's timezone for non-UTC values? Or ... both? 

Given we had a test that accepted a `timeZone` prop, how did this bug creep in? Is it because the test only covers when the timezone is explicitly provided as a prop instead of coming through via `intl` and context? If so, can we add a test that exercises that case, i.e. setting the timezone in the harness rather than directly via a prop to the component? Would it be valuable to write a storybook story that shows the timezone coming in via context since that is the most common use case? 

Finally, can we update the description for the test that exercises `useInput` to distinguish it from the test above that, currently, has an identical description?

JohnC-80:
@zburke - both - the timeZone prop wasn't getting properly passed to the formatter - nor was the value of timezone being used in the formatter function. The timeZone prop overrides the value from intl context.

I don't think that converge is dependably failing/passing here... will fix with some updates.

JohnC-80:
@zburke yep, misuse of converge was the culprit for previously-passing tests. Going to have to check other tests that use it in another story...

zburke (approved):

PR #1998, opened 2023-03-06
STCOM-1110: Add correct validation of data

daniil-patckin:
## Description
The Field component cannot identify that user entered an invalid date in Datepicker. 

## Screenshots

https://user-images.githubusercontent.com/124583895/223055151-6a1c0780-b933-4d80-b44b-176ee4b95825.mov

## Issues
[STCOM-1110](https://issues.folio.org/browse/STCOM-1110)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
 * Fix Convert24hr function of timepicker to fix 'invalid date' message when timedropdown is used. Refs STCOM-1120.
 * Fix MCL Columnheaders' focus styling. Refs STCOM-1105.
 * Fix keyboard interaction with MCL Columnheaders - Enter and Spacebar can now be used to 'click' them. Refs STCOM-680.
-* No validation in "Set date range for MARC authority headings updates (CSV) report" modal after valid values were inputted once. Refs STCOM-1110.
+* The Datepicker works correctly with an invalid date. Refs STCOM-1110.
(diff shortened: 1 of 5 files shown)


JohnC-80:
This is good for the particular case for now... The component will probably need to include its own internal validation indicator to let the user know that the value they've entered is an invalid one.

JohnC-80 (approved):

BogdanDenis (approved):

BogdanDenis (changes requested):
Is this a PR for STCOM-1130 or STCOM-1110?
On scrum board 1130 is in progress, but here 1110 is mentioned.
Also please update CHANGELOG

zburke:
@daniil-patckin, saying that STCOM-1130/UIMARCAUTH-245 is the blocker to this one implies that work must be done in UIMARCAUTH before this issue can be resolved. That is backwards; the UIMARCAUTH problem is _caused_ by this bug. Indeed, an STCOM issue _cannot_ be blocked by any ui-$application issue because STCOM does not depend on any UI application. When both tickets were in STCOM, this confusion is what led me to move 1130 to "In code review". 

Thank you for adding a CHANGELOG entry. In the future, please use commit subjects and CHANGELOG entries that describe what is changing, i.e. that describe the fix, instead of describing the bug. These values should describe the state of the code after the change is committed, e.g. "Datepicker correctly shows validation warnings".

PR #2001, opened 2023-03-07
STCOM-1127 Move react-flexbox-grid into stripes-components

JohnC-80:
no more waiting on this dependency to update its peers or node versions...

- moved components over.
- adopted the CSS dependency `flexboxgrid2`
- tests.

image


zburke (approved):
LGTM

PR #2007, opened 2023-03-17
STCOM-1092 Replace file-loader dependency with asset-modules

JohnC-80:
As per `webpack` docs: https://webpack.js.org/guides/asset-modules/, we can jettison our decrepit version of `file-loader`


zburke (commented):
The storybook webpack config looks good, and https://github.com/folio-org/stripes-webpack/pull/73 contains similar changes buried inside it. Should we merge that work first, or doesn't it matter? 

The DatepickerDemo changes seem out of place in this PR.

zburke (approved):

PR #2013, opened 2023-03-23
STCOM-1144 use svgr instead of react-svg-loader, jestify hopefully

JohnC-80:
Jest tests fail if `stripes-components` Icon isn't mocked - this was originally due to usage of webpack 'require.context' api to dynamically load all of the contents of the `icons` directory without having to manually maintain a list.

This PR 
 - removes the dependency `react-svg-loader`
 - replaces it with `@svgr/webpack`
 - sets up a 'default' for Icon to render in JSDOM...
 - falls back to the default if the provided 'icon' prop cannot be found.
 - removes the local webpack loader override.
 - The specific behavior from the override is now accomplished through resourceQueries. Built-in icon SVG imports will use the `?icon` query in their import path... all other SVG's will be inlined via stripes-webpack's configuration.
 
 This PR depends on https://github.com/folio-org/stripes-webpack/pull/92 in order to work.

### DIFF
--- a/.storybook/stcom-webpack.config.js
+++ b/.storybook/stcom-webpack.config.js
@@ -105,7 +105,5 @@ module.exports = async (config) => {
   ];
 
   config.module.rules.splice(svgRuleIndex,1, ...svgrRules);
-
-  console.log(JSON.stringify(config, null, 2 ));
   return config;
 }
--- a/.storybook/stcom-webpack.config.js
+++ b/.storybook/stcom-webpack.config.js
@@ -75,19 +75,18 @@ module.exports = async (config) => {
     config.module.rules = config.module.rules.concat([
     {
       test: /\.(jpe?g|png|gif)$/i,
-      loader: "file-loader"
+      type: 'asset/resource',
+      generator: {
+        filename: './img/[name].[contenthash].[ext]'
+      }
     },
-    
+
     {
       test: /\.(eot|ttf|woff|woff2?)$/,
-      use: [
-        {
-          loader: 'file-loader',
-          options: {
-            name: 'fonts/[name].[hash].[ext]'
-          }
-        }
-      ],
+      type: 'asset/resource',
+      generator: {
+        filename: './fonts/[name].[hash].[ext]'
+      }
     }
   ]);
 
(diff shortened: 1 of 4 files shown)


lib/Icon/Icon.js
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { pickBy } from 'lodash';
+import _pickBy from 'lodash/pickBy';
+import _get from 'lodash/get';
```
zburke, line 5: I know it doesn't matter technically, but the most common way we import from lodash in stripes with named imports from the top-level package.

.storybook/stcom-webpack.config.js
```diff
@@ -91,17 +91,27 @@ module.exports = async (config) => {
   ]);
 
   const svgRuleIndex = config.module.rules.findIndex(r => { const t = new RegExp(r.test); return t.test('m.svg'); });
-  config.module.rules[svgRuleIndex] = {
+  // config.module.rules[svgRuleIndex] = {
+  //     test: /\.svg$/,
+  //     type: 'asset',
+  //     resourceQuery: /url/, // *.svg?url
+  //   };
```
zburke, line 98: Can we omit this? Alternatively, if it's a helpful explanation of how to do something, let's add a comment that describes this sample config.

package.json
```diff
@@ -57,7 +57,7 @@
     "@storybook/builder-webpack5": "^6.3.12",
     "@storybook/manager-webpack5": "^6.3.12",
     "@storybook/react": "^6.3.6",
-    "@svgr/webpack": "^5.5.0",
+    "@svgr/webpack": "7.0.0",
```
zburke, line 60: Do we want this exact version or would a ^ or ~ range be acceptable? If we need this exact range, please explain why in the PR description and final commit-comment so we have a record of the decision when spelunking through the history.

zburke (commented):
This looks good but I think needs a little more clean up before it's ready to commit.

JohnC-80:
Thanks @zburke should be ready to rock now...

mkuklis (approved):
Nicely done @JohnC-80!

zburke (approved):

PR #2019, opened 2023-04-04
Revert "STCOM-1144 use svgr instead of react-svg-loader, jestify hope…

JohnC-80:
…fully (#2013)"

This reverts commit 82218cc871b0f585ce3946717b5d2ac1f38d45b6.

### DIFF
--- a/.storybook/stcom-webpack.config.js
+++ b/.storybook/stcom-webpack.config.js
@@ -91,19 +91,17 @@ module.exports = async (config) => {
   ]);
 
   const svgRuleIndex = config.module.rules.findIndex(r => { const t = new RegExp(r.test); return t.test('m.svg'); });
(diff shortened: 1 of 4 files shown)


mkuklis (approved):

zburke:
Backstory: effectively, handling of SVGs acts like a peer-dep (i.e. it needs to be sync'ed across multiple repos) but we don't have a good way to express that because stripes-cli/stripes-webpack are only dev-deps and there is no way to express a dev-peer. 

IOW, we need to keep handling of SVGs sync'ed between stripes-webpack and stripes-components, which means only making changes in major versions. #2013 and folio-org/stripes-webpack/pull/92 did not change the major versions, which meant you could easily get yourself into a situation where svg handling was out of sync.

PR #2023, opened 2023-04-12
STCOM-1148 absolutely positioned content can block pane header.

JohnC-80:
[STCOM-1148](https://issues.folio.org/browse/STCOM-1148)

*Problem* -  absolutely positioned content such as smart-components' NoResultsMessage will bust right out of Pane's content div, despite it having `overflow: hidden` applied in its CSS. In order to cull the overflow of non-static elements, the content containing div should be 'positioned' as well.

Approach - 
Set `position: relative` on pane's content div's class.

### Screenshot 🎉 
hooray, action dropdowns now work when results are empty.
image


zburke (approved):
Wow, what a clear explanation and surgical fix for a totally mysterious and would-have-been-impossible-for-me-to-debug situation. Nice work!

PR #2025, opened 2023-04-14
STCOM-1150: Remove unchecked calls to Moment.locale

ncovercash:
# [Jira STCOM-1150](https://issues.folio.org/browse/STCOM-1150)

This fixes the bug reported in [STCOM-1150](https://issues.folio.org/browse/STCOM-1150) and causing [UICAL-267](https://issues.folio.org/browse/UICAL-267) by removing the erroneous call to `moment.locale`.

To summarize [this Slack thread](https://folio-project.slack.com/archives/C210UCHQ9/p1680101882014399), the call to `moment.locale` for locales unknown to Moment (such as `en-SE` for Swedish English) caused Moment to fallback to default locale settings, however, it would also cache these default settings as the correct locale for `en-SE`.  When `Timepicker` later checked to see if Moment knew the locale, Moment then reported that it did, returning incorrectly assumed locale information, resulting in 24-hour `Timepicker`s switching to 12-hour at reconstruction.

@JohnC-80 / @zburke one of y'all will likely want to clone this branch over to `folio-org` to check CI (I don't have permissions here), however, running tests revealed no new failures (the only date-related failure was due to the change of the space in locale times during a recent Chrome update (replaced with `\u202f`)).  Fixed that here too, for sanity's sake.

### DIFF
--- a/lib/Datepicker/tests/Datepicker-test.js
+++ b/lib/Datepicker/tests/Datepicker-test.js
@@ -594,6 +594,8 @@ describe('Datepicker', () => {
       await datepicker.openCalendar();
     });
 
+    it('does not poison moment locale cache', () => expect(moment.locales()).to.not.include('en-se'));
+
     it('renders Monday as the first day of the week', () => Weekday('Mon', { index: 0 }).exists());
 
     it('renders weekday and month days in correct alignment', async () => {
--- a/lib/Timepicker/tests/Timepicker-test.js
+++ b/lib/Timepicker/tests/Timepicker-test.js
@@ -109,7 +109,7 @@ describe('Timepicker', () => {
     it('emits an event with the time formatted as displayed', () => converge(() => (timeOutput === '05:00 PM')));
   });
 
-  describe('correctly handles unknown locales', () => {
+  describe('correctly handles unknown locales (en-SE)', () => {
     beforeEach(async () => {
       await mountWithContext(
         <Timepicker
--- a/lib/Timepicker/tests/Timepicker-test.js
+++ b/lib/Timepicker/tests/Timepicker-test.js
@@ -1,6 +1,8 @@
+import { expect } from 'chai';
 import React from 'react';
 import { describe, beforeEach, it } from 'mocha';
-import moment from 'moment-timezone';
+import moment from 'moment';
+import momentTz from 'moment-timezone';
 import {
   runAxeTest,
   HTML,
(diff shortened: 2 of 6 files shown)


JohnC-80 (approved):

zburke:
[Reference for those `\u202f` changes](https://github.com/nodejs/node/issues/46123). So much fun. 🙄

ncovercash:
> [Reference for those `\u202f` changes](https://github.com/nodejs/node/issues/46123). So much fun. 🙄

Every time I deal with weird time zone/localization bugs, I am reminded of [this amazing video](https://www.youtube.com/watch?v=-5wpm-gesOY)

zburke:
Ha! Great video, and he sure gets the moral correct (make somebody else do this if possible)! It's no small wonder that date/time pickers seem to always be the missing components in every single OSS component library!

I remember learning about unix in college and being totally bowled over by the command `cal 9 1752`. Part of it was the specific programming challenges, but more it was the eye-opening experience that something I'd thought was simple was actually really REALLY complicated, interesting, contentious, that a "factual" thing like the date could even _be_ contentious, etc.

lib/Datepicker/tests/Datepicker-test.js
```diff
@@ -594,6 +594,8 @@ describe('Datepicker', () => {
       await datepicker.openCalendar();
     });
 
+    it('does not poison moment locale cache', () => expect(moment.locales()).to.not.include('en-se'));
+
```
zburke, line 598: I believe moment is case-sensitive WRT locales so we need to check for `en-SE` here.

util/dateTimeUtils.js
```diff
@@ -48,7 +48,7 @@ export const getLocaleDateFormat = ({ intl }) => {
     format = getMomentLocalizedFormat(intl);
   }
 
-  return format;
+  return format.replaceAll('\u202f', ' ');
```
zburke, line 51: I agree with this change, but please add comment, e.g. a link to https://github.com/nodejs/node/issues/46123. Without any context, it's pretty opaque

lib/Timepicker/tests/Timepicker-test.js
```diff
@@ -107,6 +109,21 @@ describe('Timepicker', () => {
...
+        <Timepicker
+          id="timepicker-locale-test"
+          locale="en-SE"
+        />
+      );
+
+      await timepicker.clickInput();
+    });
+
+    it('does not poison moment locale cache', () => expect(moment.locales()).to.not.include('en-se'));
```
zburke, line 124: en-SE

zburke (changes requested):
I believe moment is case-senstive WRT locales so those tests need to check for the exact same locale provided as the prop.

lib/Datepicker/tests/Datepicker-test.js
ncovercash: Checked; moment will add both the string with its original casing and another in all lowercase

lib/Timepicker/tests/Timepicker-test.js
ncovercash: see comments on other one

JohnC-80:
Closing this due to age/time moving forward and future releases helping this issue in their moment locale-picking logic.. and also DayJS here we come!

PR #2034, opened 2023-04-18
STCOM-1129 Accordion incorrectly rendering `aria-label`

JohnC-80:
[STCOM-1129](https://issues.folio.org/browse/STCOM-1129)

Accordion accepts a 'label' prop that is a string/node and attempts to apply it to 'aria-label' - which results in `aria-labe='[object Object]" and is an accessibility fail.

The best appraoch was just to remove the `aria-label` attribute altogether since `aria-labelledby` was already in use and overrode `aria-label` anyhow.

Additionally, passed the `labelId` to the header - best to pass it along than to hard-code it in both places.


zburke (approved):
LGTM!

PR #2035, opened 2023-04-20
STCOM-1154: Provide onChange and ariaLabel props to TextArea component

mariia-aloshyna:
## Purpose
To provide **onChange**, **value**, and **ariaLabel** prop to **TextArea** component as it's done for TextField component e.g. to make it possible to provide a callback on textaria change and set ariaLabel (although it's possible to set it as 'aria-label' but added it for consistency)

## Refs
https://issues.folio.org/browse/STCOM-1154


przemyslawturek (approved):

przemyslawturek (changes requested):
I don't think that props valid, dirty and others should be removed.

artem-blazhko (approved):

JohnC-80 (approved):

PR #2036, opened 2023-04-20
STCOM-1150v2

JohnC-80:
This PR checks that the moment's active locale matches the lowercase version of the locale passed through props/intl before using its static locale. If not, it falls back to getting a localized time format from Browser Intl API.

`moment` doesn't have an `en-SE` locale.. it falls back to `en` if it's attempted using `moment.locale('en-SE')` As far as time-picker is concerned, the discrepancy between them is that `en-SE` is a 24 hour format, but `en` is 12 hour... accurate formatting info is provided from browser INTL API, however, so that's what we can fall back to.

### DIFF
--- a/lib/Timepicker/TimeDropdown.js
+++ b/lib/Timepicker/TimeDropdown.js
@@ -23,6 +23,7 @@ const propTypes = {
   intl: PropTypes.shape({
     formatMessage: PropTypes.func.isRequired
   }).isRequired,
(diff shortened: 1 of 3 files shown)


JohnC-80:
Closing this due to age/time moving forward and future releases helping this issue in their moment locale-picking logic.. and also DayJS here we come!

PR #2037, opened 2023-04-21
STCOM-1155: Fix the key property for <SelectOption/> to fix duplications when the label is not unique.

Dmytro-Melnyshyn:
## Purpose
Fix the `key` property for `<SelectOption/>` to fix duplications when the label is not unique.

<details>
<summary>Screencast of the issue.</summary>

https://user-images.githubusercontent.com/77053927/233616927-539123e0-7ac5-4315-9f37-507942b04938.mp4

</details>

## Issues
[STCOM-1155](https://issues.folio.org/browse/STCOM-1155)

## Screencast

https://user-images.githubusercontent.com/77053927/233617539-7703942b-f61d-4b42-a5cc-9b9537b9d7d6.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Timepicker considers timezone when parsing value prop. Refs STCOM-1141.
 * PaneContent div should be position: relative to hide overflow from absolutely positioned contents. Refs STCOM-1148.
 * Adjust styles for overlay controls rendered in MCL rows. Refs STCOM-1149, UIRS-100.
+* Fix the `key` property for `<SelectOption/>` to fix duplications when the label is not unique. Refs STCOM-1155.
 
 ## [11.0.0](https://github.com/folio-org/stripes-components/tree/v11.0.0) (2023-01-30)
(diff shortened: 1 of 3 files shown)


BogdanDenis (approved):

JohnC-80 (changes requested):
The type of `value` within items isn't necessarily a string - it can be an object also. Is there a use-case where it is not possible to use the `itemToString` prop itself to achieve uniqueness? It is provided the item it allows you to contrive the string from any of its properties. You would probably want the same uniqueness among accessible annoucements among the items, rather than having it repeat.

Dmytro-Melnyshyn:
> The type of `value` within items isn't necessarily a string - it can be an object also. Is there a use-case where it is not possible to use the `itemToString` prop itself to achieve uniqueness? It is provided the item it allows you to contrive the string from any of its properties. You would probably want the same uniqueness among accessible annoucements among the items, rather than having it repeat.

Yes, I can fix it in `itemToString` and will get uniqueness among accessible announcements among the items. Thanks.

PR #2050, opened 2023-04-26
STCOM-1157 Two X's displayed on search fields

JohnC-80:
https://issues.folio.org/browse/STCOM-1157

These styles went away when we recently upgraded the `normalize.css` dependency. This PR simply re-adds them in our global styles.

Before:
image

After: 
image


JohnC-80:
Autoprefixer does try to do this for most cases so you could just exclude the prefis... I'm not sure if there's something we aren't doing correctly to make it not work for these, or if it's just the obscurity/exclusivity of the pseudo elements.

JohnC-80:
Doesn't seem like there's an obvious way to make it cleaner according to https://preset-env.cssdb.org/playground/ so just going to keep the obscure prefixes here for now...

PR #2057, opened 2023-05-02
STCOM-1158 - center pagination to  scroll parent.

JohnC-80:
https://issues.folio.org/browse/STCOM-1158

Fixes the cut off pagination in result frames.
Next/Prev pagination is rendered with flexbox and makes use of CSS for positioning - but CSS doesn't have a way to size an element based on `scrollParent` - just its direct parent, so we employ some JS to do that for us... 

Also of note: the appearance on this is blessed by KKester and KGambrell

No more ugly cutoff.... spacing similar to acq-components.
### Before
image

### After
image


Terala-Priyanka (approved):

zburke (commented):
@JohnC-80 , I think you gotta choose between removing dom-helpers in #2059, and importing `scrollParent` from it here :)

JohnC-80:
@zburke Good catch. Ugh! 🙄

JohnC-80:
@zburke went ahead and contrived a local utility function so we can still exnay the ependencyday (yay!)

lib/MultiColumnList/CenteredContainer.js
```diff
@@ -1,24 +1,33 @@
...
 import Layout from '../Layout';
 import css from './MCLRenderer.css';
 
-const CenteredContainer = ({ innerRef, visible, width, children, role, style: styleProp }) => {
-  // subtracting the margins to prevent horizontal scroll
-  const endOfListWidth = isFinite(width) ? `${Math.max(width - 20, 200)}px` : '100%';
+const CenteredContainer = ({ innerRef, visible, children, role, style: styleProp }) => {
+  const wrappingElement = useRef(innerRef || null);
+
+  useLayoutEffect(() => {
```
zburke, line 10:
```
// subtracting the margins to prevent horizontal scroll
```
This `useLayoutEffect` function is pretty opaque. Can we restore the comment that was removed (or provide a fresh explanation if it's no longer accurate)?

zburke (approved):
OK by me, but this code is a bit opaque and would benefit from additional documentation.

PR #2072, opened 2023-05-30
STCOM-1232 Update FilterGroups.js

JohnC-80:
Use Regex replace to remove illegal characters.

## screencast
chrome_Y0FmXVP63g

Refs [STCOM-1232](https://issues.folio.org/browse/STCOM-1232)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Ensure CSS visibility of datepicker's year input number spinner. Refs STCOM-1225.
 * Include pagination height in MCL container height calculation. Refs STCOM-1224.
 * Added `indexRef` and `inputRef` props to `<SearchField>`. Refs STCOM-1231.
+* Update FilterGroups.js. Refs UIU-2731.
```
mariia-aloshyna, line 12: Honestly, this change description looks not informative to me. Let's be more specific here maybe?

zburke (commented):
How have you assessed that ALL parties that leverage this component agree with this change? We’ve seen a lot of trouble recently where somebody makes a change to escape a special character in order to resolve their bug (e.g. people typing foo* into a search box, which needs to be converted to a search like `"foo\*"`), only to realize that somebody else has a function that converts user input from `foo` to `foo*` and relies on that executing a search like `"foo*"`. Additional details at [STUTL-33](https://issues.folio.org/browse/STUTL-33).

Maaaaaybe this is safe because it just affects filter checkboxes (where we're doing an exact match, and so special characters _should_  be escaped) rather than something that handles text input (where applications often append a wildcard in order to do left-anchored searches and only special characters input by the user directly into the field should be escaped, but those added by the application must remain as-is). 

Also, I agree with @mariia-aloshyna that the PR description and CHANGELOG entry need more detail: you are escaping CQL special characters here, so say that. "Update filters" is no better a comment than "fixed bug".

CHANGELOG.md
Terala-Priyanka: Updated the change log description. Please review.

Terala-Priyanka:
> How have you assessed that ALL parties that leverage this component agree with this change? We’ve seen a lot of trouble recently where somebody makes a change to escape a special character in order to resolve their bug (e.g. people typing foo* into a search box, which needs to be converted to a search like `"foo\*"`), only to realize that somebody else has a function that converts user input from `foo` to `foo*` and relies on that executing a search like `"foo*"`. Additional details at [STUTL-33](https://issues.folio.org/browse/STUTL-33).
> 
> Maaaaaybe this is safe because it just affects filter checkboxes (where we're doing an exact match, and so special characters _should_ be escaped) rather than something that handles text input (where applications often append a wildcard in order to do left-anchored searches and only special characters input by the user directly into the field should be escaped, but those added by the application must remain as-is).
> 
> Also, I agree with @mariia-aloshyna that the PR description and CHANGELOG entry need more detail: you are escaping CQL special characters here, so say that. "Update filters" is no better a comment than "fixed bug".

Updated the change log description. Please review.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Ensure CSS visibility of datepicker's year input number spinner. Refs STCOM-1225.
 * Include pagination height in MCL container height calculation. Refs STCOM-1224.
 * Added `indexRef` and `inputRef` props to `<SearchField>`. Refs STCOM-1231.
+* Use Regex replace to remove illegal characters in filter groups. Refs UIU-2731.
```
zburke, line 12: Better, but better still:
> * Use Regex replace to escape CQL special characters in filter groups. Refs STCOM-1232.

since they're not illegal and they're not being removed. And this is an STCOM story not a UIU one. (I moved the Jira. If a bug presents in one app/repo (e.g. ui-users) but is caused from a shared library in a different repo, either move the ticket or clone it into the project where the change will be applied. If we left this in UIU, there'd be no way to look at Jira and see, "Oh, this was fixed in STCOM v12.1.0" because it would be affiliated with a UIU fixversion, not an STCOM fixversion.)

zburke:
@Terala-Priyanka, do you know why @JohnC-80 closed the original PR? And can you explain why it is being reopened now? What changed?

JohnC-80:
This PR is currently not the solution for this issue. Since there has been such a divergent spread of which "illegal" characters are actually "meaningful" characters between ui-modules, hard-wiring this into any centralized FE library has never worked out as the best path forward.
Maybe I need to gain a better understanding about the issue, but I believe the goal of this PR needs to happen through a utility function, but with the ability to modify the 'default' set of filtered characters per-module - at least don't hard-wire the solution, given our current state of history for this brand of fix.

Terala-Priyanka:
> @Terala-Priyanka, do you know why @JohnC-80 closed the original PR? And can you explain why it is being reopened now? What changed?

@zburke I think it was closed in the interest of https://issues.folio.org/browse/STUTL-33

Terala-Priyanka:
@JohnC-80 @zburke @mariia-aloshyna 
I would be moving this PR to draft as of now. I will provide an alternate solution within ui-users through another PR.

PR #2082, opened 2023-06-20
STCOM-1171 - expand options of meta section component

Terala-Priyanka:
## Purpose
Expand options of MetaSection component

## Approach
As described in the description of ticket, there came up a pattern where MetaSection should be rendered without an Accordion. The changes addressed this requirement (whose roots lay in https://issues.folio.org/browse/UICIRC-858).
Implementation  accespts three new props useAccordion, noBackground and inlineLayout which pertain to conditional rendering of Accordion, its content and its background.

## Refs
https://issues.folio.org/browse/STCOM-1171

## Note
Due to something that might have broken, test coverage show 0%. All the conditions have been covered through tests.


lib/MetaSection/MetaSection.js
```diff
@@ -39,13 +40,18 @@ const propTypes = {
...
+  useAccordion: PropTypes.bool,
 };
 
 const defaultProps = {
   hideSource: false,
   showUserLink: false,
-  headingLevel: 4
+  headingLevel: 4,
+  inlineLayout: false,
+  noBackGround: false,
```
JohnC-80, line 53: The boolean props(`inlineLayout`, and `noBackGround` here) that default to false don't need a default prop... they'll just be undefined/falsey.

lib/MetaSection/MetaSection.js
```diff
@@ -119,36 +114,112 @@ class MetaSection extends React.Component {
...
+      hideSource,
+      inlineLayout,
+      lastUpdatedBy,
+      lastUpdatedDate,
+    } = this.props;
+
+    return (
+      <div
+        className={css.metaSectionContent}
+        style={inlineLayout ? { display: 'flex' } : { display: 'flex column' }
```
JohnC-80, line 169: It will be better to bake these values out into CSS classes for testability of this and use the `classnames` dependency to functionally apply them - Then you can assert on classnames vs digging into style attribute values and the actual implemenation of those classes/values would be hidden away.

lib/MetaSection/MetaSection.js
```diff
@@ -119,36 +114,112 @@ class MetaSection extends React.Component {
...
+    const {
+      noBackGround,
+      useAccordion,
+    } = this.props;
+
+    return (
+      <div
+        className={css.metaSectionRoot}
+        style={{
+          'background-color': `${noBackGround ? 'unset' : 'var(--color-fill)'}`
```
JohnC-80, line 215: Same as the `inlineLayout` comment above. The more of these inline styles that could be cleaned up by moving them to CSS classes, the better.

JohnC-80 (changes requested):

JohnC-80 (approved):

BogdanDenis (approved):

PR #2083, opened 2023-06-22
React 18 POC

zburke:
Lots of little clean up changes awaiting promises and such, but a few notable things:
* [`tests/helpers.js`](https://github.com/folio-org/stripes-components/compare/master...react-18-poc?expand=1#diff-b8a297bda8e92eda02b4cbca05b2a53f684f286d286e314d53052933157d84b0): refactor to use the React 18 [`createRoot` API](https://react.dev/reference/react-dom/client/createRoot), retaining the unmount/createRoot behavior. Merely calling `root.render()`, even though it has a new component in each test, was not sufficient because BigTest operates "outside" of React to remove test elements from the DOM. Thus, we 'need to tell React to “stop” managing the removed root’s content by calling `root.unmount`. Otherwise, the components inside the removed root won’t know to clean up and free up global resources...'. 
* [`tests/helpers/getHookExecutionResult.js`](https://github.com/folio-org/stripes-components/compare/master...react-18-poc?expand=1#diff-75c28b2bd18cb4350d7713cc499076e771389ce588dbd548602e9faf1fd5ad04): since `mountWithContext()` returns a promise, let it resolve and _then_ return the value assigned to the hook. Otherwise, we may get a value from an intermediate render instead of the final value.


mkuklis (approved):
👏

zburke:
Replaced by #2084, which accomplishes the same with fewer changes

PR #2095, opened 2023-07-12
STCOM-1186 Provide the `getFieldUniqueKey` prop to define a key for the list items in the `<RepeatableField>` component

usavkov-epam:
## Purpose
https://issues.folio.org/browse/STCOM-1186

Since the React library insists that keys must be unique and not change during rendering, the ability to specify a key for each element should prevent possible bugs.

Basically, this improvement is meant to help fix [UISACQCOMP-154](https://issues.folio.org/browse/UISACQCOMP-154): changing the order of items in an array (deleting, adding) causes an incorrect display of them, which is associated with using the index of this field as a `key` prop.

## Learning
Why it's not OK to use an item's index as key - https://react.dev/learn/rendering-lists#why-does-react-need-keys

## Approach
Provide `getFieldUniqueKey` as a prop for the `<RepeatableField>` component. This function accepts `field`, `index`, and `fields` as arguments.

Set the default `getFieldUniqueKey` prop which will return a function that will return an item's index.

##### Using items indexes as a `key` prop:
chrome_VINMkBpCT7

##### Using unique identifiers as a `key` prop:
chrome_VcaZZgdgnX

### DIFF
--- a/lib/RepeatableField/RepeatableField.js
+++ b/lib/RepeatableField/RepeatableField.js
@@ -103,7 +103,7 @@ class RepeatableField extends Component {
             {fields.map((field, index) => (
               <li
                 className={css.repeatableFieldItem}
-                key={getFieldUniqueKey(field, index, fields) || index}
+                key={getFieldUniqueKey(field, index, fields)}
                 data-test-repeatable-field-list-item
               >
                 <div className={css.repeatableFieldItemContent}>
--- a/lib/RepeatableField/RepeatableField.js
+++ b/lib/RepeatableField/RepeatableField.js
@@ -21,6 +21,7 @@ class RepeatableField extends Component {
       PropTypes.array,
       PropTypes.object
     ]).isRequired,
+    getFieldUniqueKey: PropTypes.func,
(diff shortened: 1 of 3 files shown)


lib/RepeatableField/RepeatableField.js
```diff
@@ -100,7 +103,7 @@ class RepeatableField extends Component {
             {fields.map((field, index) => (
               <li
                 className={css.repeatableFieldItem}
-                key={index}
+                key={getFieldUniqueKey(field, index, fields) || index}
```
zburke, line 106: Since you're supplying a default prop value here the `|| index` is just redundant, isn't it?

zburke (approved):
LGTM. Thanks for the explanation of the bug in the PR description and in the readme! It's really nice being able to avoid the thought, "OK, I see what this does, but ... why do we need this?"

lib/RepeatableField/RepeatableField.js
usavkov-epam: Oh right, thanks!

NikitaSedyx (approved):

PR #2114, opened 2023-08-14
STCOM-1185 Number spinners are missing from input[type=number]

JohnC-80:
This symptom is present in Chrome, but not Firefox.
Default behavior in Chrome is to only display spinner buttons on number inputs when the input is hovered or focused.

The approach: add a global style to always set the opacity of the `-webkit-inner-spin-button` to `1`, making it visible.


Terala-Priyanka (approved):
Could you please update changelog file?

PR #2115, opened 2023-08-14
STCOM-1144 update to use SVGR for loading SVG's

JohnC-80:
This upgrade was [added](https://github.com/folio-org/stripes-components/pull/2013), then [removed](https://github.com/folio-org/stripes-components/pull/2019) - as it was best suited for a breaking version, since it did change the API between stripes-webpack and stripes-components... So now, after a breaking version bump, this is hopefully the conclusion to this saga.

From the original 'added' PR:
> Jest tests fail if stripes-components Icon isn't mocked - this was originally due to usage of webpack 'require.context' api to dynamically load all of the contents of the icons directory without having to manually maintain a list.

> This PR

> removes the dependency react-svg-loader
replaces it with @svgr/webpack
sets up a 'default' for Icon to render in JSDOM...
falls back to the default if the provided 'icon' prop cannot be found.
removes the local webpack loader override.
The specific behavior from the override is now accomplished through resourceQueries. Built-in icon SVG imports will use the ?icon query in their import path... all other SVG's will be inlined via stripes-webpack's configuration.
This PR depends on https://github.com/folio-org/stripes-webpack/pull/92 in order to work.


zburke (approved):
LGTM. Platform and workspace builds are clean!

PR #2119, opened 2023-08-29
STCOM-1196: Adjust scroll position to the top for non-sparse array

OleksandrHladchenko1:
[STCOM-1196](https://issues.folio.org/browse/STCOM-1196)

* When array is sparse focus index starts with `this.focusTargetIndex`
* When array is non-sparse, set focusIndex to 0


mariia-aloshyna (approved):

zburke (approved):
LGTM. It would be nice to include the PR description in comments directly in the code in order to explain what is happening here. It'll save the next person who comes to this function from needing to do some git archaeology to figure out what's going on and why.

PR #2124, opened 2023-09-12
STCOM-1199: Add current for selectList and container refs due to removal of dom-helpers dependency.

Dmytro-Melnyshyn:
## Purpose
Add `current` for `selectList` and `container` refs due to the removal of dom-helpers dependency.

## Issue
[STCOM-1199](https://issues.folio.org/browse/STCOM-1199)

## Screencast of the issue

https://github.com/folio-org/stripes-components/assets/77053927/43542479-715d-4812-ba3c-389838332f3e


BogdanDenis (approved):

JohnC-80 (approved):
Thank you @Dmytro-Melnyshyn!

PR #2133, opened 2023-09-28
STCOM-1207 - fix incorrect accordion keyboard handler, add tests

JohnC-80:
Numerous modules experiencing difficulty with keyboard shortcuts for expandAll/Collapse all not working.

## Approach
As part of updates for Reactv18, the state of accordions had to change, pushing status to its own key rather than being the whole state. The connected parts that were the cause of this bug, untested at the time, were the centralized keyboard shortcut handlers (`expandAllSections`, `collapseAllSections`) so those were broken, as they were expecting the previous shape of AccordionStatus state.


zburke (approved):
LGTM. Thank you for the detailed explanation of the problem in the PR description. That high-level overview really helped me understand the source of the bug and how the changes in `keyboardShortcuts.js` address it.

PR #2138, opened 2023-10-05
STCOM-1210 Fix state mutation in `<AccordionSet>`. Add `onRegisterAccordion` and `onUnregisterAccordion` props.

BogdanDenis:
## Description
Fix issue in Inventory app where facet counts don't reset when switching between browse options.

## Approach
The issue is caused by two things:
1. Mutation of `<AccordionSet>` state. Because of this mutation other components aren't updated
2. `<AccordionSet>` doesn't notify consumers when accordions are registered and unregistered. The component propagates this info down into child `<Accordion>` components, but not up to a parent. In Inventory's case this is required because accordion state is used to determine which facets endpoints need to be called.
To fix this new `onRegisterAccordion` and `onUnregisterAccordion` props have been added. They will allow the parent component to be notified when accordions are mounted/unmounted.

## Screenshots

https://github.com/folio-org/stripes-components/assets/19309423/4da75421-b6ce-40b4-878d-65466892e25e

## Issues
[STCOM-1210](https://issues.folio.org/browse/STCOM-1210)

## Related PRs
https://github.com/folio-org/ui-inventory/pull/2293


Dmytro-Melnyshyn (approved):

JohnC-80 (approved):

lib/Accordion/tests/Accordion-test.js
```diff
@@ -287,6 +287,46 @@ describe('Accordion - as part of an AccordionSet', () => {
   });
 });
 
+describe.only('unmounting Accordion - as part of an AccordionSet', () => {
```
Dmytro-Melnyshyn, line 290: please remove `only`

PR #2158, opened 2023-10-25
STCOM-1220 Make `<SearchField>` support input and textarea as an input field

BogdanDenis:
## Description
Make <SearchField> support input and textarea as an input field by adding a new `inputType` prop

## Screenshots

https://github.com/folio-org/stripes-components/assets/19309423/f7a76344-c3dd-45f7-8254-854a8ace9ccf

## Issues
[STCOM-1220](https://issues.folio.org/browse/STCOM-1220)

## Related PRs
https://github.com/folio-org/stripes-smart-components/pull/1407
https://github.com/folio-org/ui-inventory/pull/2322
https://github.com/folio-org/stripes-testing/pull/2040


JohnC-80 (changes requested):
Thank you @BogdanDenis ! I'm seeing a good bit of console noise from `react-dom` when I switch out this prop locally in `ui-users`. It looks like we need to add a handful of props to the 'omit' list in `Textarea.js` at/around https://github.com/folio-org/stripes-components/blob/master/lib/TextArea/TextArea.js#L160... including `marginBottom0` , `rootClass`, `lockWidth`, `onSubmitSearch`, and `newLineOnShiftEnter`, as those just get passed right along to the dom element (which causes the errors to be echoed in the console.)
Other than that, the code here looks good.

lib/SearchField/SearchField.js
```diff
@@ -112,35 +133,61 @@ const SearchField = (props) => {
     inputPlaceholder = selectedIndexConfig.placeholder || '';
   }
 
+  const getInputComponentProps = () => {
+    // TextField and TextArea have slidghtly different APIs so we need to pass props correctly
```
JohnC-80, line 137:
```suggestion
    // TextField and TextArea have slightly different APIs so we need to pass props correctly
```
pedantic - typo...

Dmytro-Melnyshyn (approved):

JohnC-80 (approved):

PR #2161, opened 2023-10-30
STCOM-1000: Firefox: TextLink - underline not showing up on nested spans with 'display: inline-flex'

UladzislauKutarkin:
I have added changes in story book, since it's not a bug. 
Screenshot 2023-10-26 at 17 31 55

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add `hasMatchSelection` to `<AdvancedSearch>` to hide/show search match selection dropdown. Refs STCOM-1211.
 * Add z-index of 1 to callout out to have it always render on top of sibling elements. Fixes STCOM-1217.
+* Firefox: TextLink - underline not showing up on nested spans with 'display: inline-flex'. Refs STCOM-1000.
 
 ## [12.0.0](https://github.com/folio-org/stripes-components/tree/v12.0.0) (2023-10-11)
 [Full Changelog](https://github.com/folio-org/stripes-components/compare/v11.0.0...v12.0.0)
--- a/lib/TextLink/stories/BasicUsage.js
+++ b/lib/TextLink/stories/BasicUsage.js
@@ -40,6 +40,13 @@ const BasicUsage = () => (
         </TextLink>
       )}
     </Tooltip>
+    <br /><br />
+    <div style={{ width: 300 }}>
+      <TextLink href="#some-hash">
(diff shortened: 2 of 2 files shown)


vashjs (approved):

JohnC-80:
@UladzislauKutarkin in cases where this kind of layout is required, we would want to be able to manually apply this style to children. This should probably be provided as either a render-prop from `TextLink` or a global className.

UladzislauKutarkin:
> @UladzislauKutarkin in cases where this kind of layout is required, we would want to be able to manually apply this style to children since it isn't inherited. This should probably be provided as either a render-prop from `TextLink` or a global className.
> 
> True, this is not a browser bug, this behavior from browsers is quite unexpected by developers (as it was when we ran into it.)

@JohnC-80, maybe we need to add a description about that in readme file ? I mean we can provide `children` with there own `classnames`, because it will be useful and more flexible to use Textlink in different cases.

JohnC-80:
> @JohnC-80, maybe we need to add a description about that in readme file ? I mean we can provide `children` with there own `classnames`, because it will be useful and more flexible to use Textlink in different cases. So, I don't think that we need to make a new one render props for `children`

Part of this is saving developers' what would essentially be 'boilerplate' for their case - if they needed this layout, they would -always- have to set up the class themselves for each ui module that needs this. If it doesn't come from a centralized library, it would probably be a different class with at least `text-decoration: underline;`, plus any other properties necessary to match the existing `TextLink` styling. We would definitely need to mention it in documentation, but it's also nicer if we can prescriptively say 'use `<Layout className="textLink" >` ' for example, or "apply the provided `classname` render prop like so...".

UladzislauKutarkin:
> > @JohnC-80, maybe we need to add a description about that in readme file ? I mean we can provide `children` with there own `classnames`, because it will be useful and more flexible to use Textlink in different cases. So, I don't think that we need to make a new one render props for `children`
> 
> Part of this is saving developers' what would essentially be 'boilerplate' for their case - if they needed this layout, they would -always- have to set up the class themselves for each ui module that needs this. If it doesn't come from a centralized library, it would probably be a different class with at least `text-decoration: underline;`, plus any other properties necessary to match the existing `TextLink` styling. We would definitely need to mention it in documentation, but it's also nicer if we can prescriptively say 'use `<Layout className="textLink" >` ' for example, or "apply the provided `classname` render prop like so...".

Can you give me example of usage for that case ? Because it's not clear how it will looks

PR #2170, opened 2023-11-08
STCOM-1228 Print styling of panes (users want to print the detail pane, but full screen display is printed)

JohnC-80:
Printing with a detail pane open should print only the detail pane, without printing the others. The before state is captured in the associated JIRAs.

### Approach

Adjusted the styles of panes to have only the last pane visible and at 100% width when the page is printed.
Used a combination of hiding panes in print mode, and displaying the last pane via a higher specificity pseudo-selector (`nth-last-of-type(1)`)
The `flex` property is `!important` in this case in order to override the inline `flex` styling that is applied to the element.

Two stories, https://issues.folio.org/browse/STCOM-1228 and https://issues.folio.org/browse/STCOM-1229 are covered by this...

After:
Inventory:
image

Users:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Ensure CSS visibility of datepicker's year input number spinner. Refs STCOM-1225.
 * Include pagination height in MCL container height calculation. Refs STCOM-1224.
 * Added `indexRef` and `inputRef` props to `<SearchField>`. Refs STCOM-1231.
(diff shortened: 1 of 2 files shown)


zburke (approved):
LGTM, but this isn't the first time we've faced problems. I'm OK moving ahead with this, but in an ideal world it would be good to have a better understanding of what has changed since STCOM-975 / #1835 such that we are seeing this problem now.

JohnC-80:
Previous PR's fixed different problems from this one - but this issue definitely seems like one that was resolved a long time ago :/

PR #2176, opened 2023-11-30
STCOM-1164 Refactor away from `color()` function

JohnC-80:
https://issues.folio.org/browse/STCOM-1164?jql=text%20~%20%22postcss%22

A while back, it was necessary to fork `postcss-color-function` into folio-org to update its dependencies and keep our own css/dependency stack happy. It's currently both an extra burden and an outdated way of handling color adjustments.

The functionality that it emulates was (renamed and eventually) removed from the CSS color-module 4 spec. There is a native `color()` function -but it is not used in the same way as we used it - for adjustment of colors.

Enter CSS color module 5 - with a very nice color-adjustment syntax. It's available in Chrome, and soon-to-be-actualized in Firefox and Safari. Since it isn't available across all browsers quite yet, for now we can employ another postCSS plugin - `@csstools/postcss-relative-color-syntax` along with a newer color model, `oklch` that makes lightness/accessible contrast much more predictable across hue shifts.

This change is enough to remove the `postcss-color-function` dependency from `stripes-components`. This change will partner with a change in `stripes-webpack` to add the dependency to the new plugin there. Eventually, we will remove the dependency for `postCSS-color-function` from `stripes-webpack` as there are a few cases of `color()` usage within ui-modules.

### Basic migration
If your module makes use of `color()` for modification via a `shade()`, your conversion will look something like as follows:

```
/** color() implementation **/
background-color: color(var(--primary) shade(8%));

/* relative-color syntax  subtract from the "l" - lightess channel */
background-color: oklch(from var(--primary) calc(l - 0.03) c h);
/* if lightening the color, simply add to the "l" channel instead */
background-color: oklch(from var(--primary) calc(l + 0.03) c h);
```

To replace `color()` with  `blend()` for blending multiple colors, use the native `color-mix()` function:

```
/* color() implementation */
background-color: color(var(--bg) blend(var(--bg-hover) 20%));

/* color-mix() implementation 
 * now you'll have to supply a color model for transform and percentages for each color.
 * note the commas since we're blending from a variable that also contains another blend
 */
background-color: color-mix(in oklch, var(--bg) 80%, var(--bg-hover) 20%);
```

Read more about the [relative color syntax](https://developer.chrome.com/blog/css-relative-color-syntax) here and the excellent [oklch color model](https://evilmartians.com/chronicles/oklch-in-css-why-quit-rgb-hsl).


zburke (approved):
Yay! Thanks for all the research. It's great to finally be able to purge our postcss-color-function fork!

PR #2180, opened 2023-12-08
STCOM-1226: TextLink - underline showing up on nested spans with 'display: inline-flex'

UladzislauKutarkin:
[STCOM-1226](https://issues.folio.org/browse/STCOM-1226)
Here we made possibility to use inline-flex with nested spans and save underline by default.
Storybook was updated.

https://github.com/folio-org/stripes-components/assets/72550466/55a3b4ff-6ca0-42dc-90e9-5e792a0aae30


lib/TextLink/TextLink.js
```diff
@@ -18,9 +18,19 @@ const TextLink = forwardRef(({ className, element, href, to, children, ...rest }
...
+  const childrenWithRootClass = React.Children.map(children, child => {
+    if (React.isValidElement(child)) {
+      return React.cloneElement(child, {
+        className: classnames(css.root, child.props.className),
+      });
+    } else {
+      return child;
+    }
+  });
+
```
zburke, line 30: Please add a comment explaining what's going on here, and why it is necessary. Previously, `props.className` was only applied to the wrapper `Element` but this change adds it as well on the Element's `children`. Why? Even sth brief like this would be helpful:
```
// apply root CSS class to child element since it may not be correctly inherited
// see details at STCOM-1226, STCOM-1000
```

zburke (changes requested):
It's hard to tell exactly what is going on here. STCOM-1226 is a Task but reads like a Bug that I think is "nested spans with CSS style `display: inline-flex` should be underlined, but they aren't". Is that correct? If so, this looks good, but a bit of in-line documentation describing why it's necessary to apply `css.root` on both the wrapper `Element` and the `child` would be helpful. 

So, code looks good. We just need a little more documentation.

JohnC-80:
With the cloneElement approach and multiple spans for children, the appearance ends up partitioned, since each child always gets the class - even non-inline-flex children. Also, if the span is rendered internally, it won't receive the class.
Given a set up like the following we should end up with as few partitions in the styling as possible.
```
          <span>Lorem ipsum dolor sit amet,</span>
          <span>This is a multiline link.</span>
          <span>consectetur adipiscing elit.
            Mauris nec elementum quam.
            Aliquam porta sapien dolor, vel gravida risus condimentum et. Nam eget elit purus.
          </span>
          <Icon icon={icon}>
            <span>Label for icon.</span>
          </Icon>
```
Is there a particular issue with an implementation like 
```
<Element data-test-text-link {...props} ref={ref}>
   {typeof children === 'function' ? children(css.root) : children}
</Element>
```
So that the class could only be applied as needed?

lib/TextLink/TextLink.js
UladzislauKutarkin: Okay @zburke , I'll add such comment

UladzislauKutarkin:
> With the cloneElement approach and multiple spans for children, the appearance ends up partitioned, since each child always gets the class - even non-inline-flex children. Also, if the span is rendered internally, it won't receive the class. Given children like the following we should end up with as few partitions in the styling as possible.
> 
> ```
>           <span>Lorem ipsum dolor sit amet,</span>
>           <span>This is a multiline link.</span>
>           <span>consectetur adipiscing elit.
>             Mauris nec elementum quam.
>             Aliquam porta sapien dolor, vel gravida risus condimentum et. Nam eget elit purus.
>           </span>
>           <Icon icon={icon}>
>             <span>Label for icon.</span>
>           </Icon>
> ```
> 
> This is the way it behaves with the classname applied to each immediate child component (note the Icon's child not receiving the class.) link-style-partition [ link-style-partition ](https://private-user-images.githubusercontent.com/20704067/289125247-e557dc82-d109-4216-9276-1ef73f186ca9.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDIyODI5MzUsIm5iZiI6MTcwMjI4MjYzNSwicGF0aCI6Ii8yMDcwNDA2Ny8yODkxMjUyNDctZTU1N2RjODItZDEwOS00MjE2LTkyNzYtMWVmNzNmMTg2Y2E5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFJV05KWUFYNENTVkVINTNBJTJGMjAyMzEyMTElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjMxMjExVDA4MTcxNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTYwNTliZGQ4ODE2ODAzMDRlMDQxZTRlZDFhNWI1N2FhYTNjMjRiZmI1MzhiYjkzZWU2MDIxM2FmMWJmZGFmNTgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.mlNuvs3BFmX5ohE-Dfnd06wPBpi1ByeoM1nxZp6l52Q) [ ](https://private-user-images.githubusercontent.com/20704067/289125247-e557dc82-d109-4216-9276-1ef73f186ca9.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDIyODI5MzUsIm5iZiI6MTcwMjI4MjYzNSwicGF0aCI6Ii8yMDcwNDA2Ny8yODkxMjUyNDctZTU1N2RjODItZDEwOS00MjE2LTkyNzYtMWVmNzNmMTg2Y2E5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFJV05KWUFYNENTVkVINTNBJTJGMjAyMzEyMTElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjMxMjExVDA4MTcxNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTYwNTliZGQ4ODE2ODAzMDRlMDQxZTRlZDFhNWI1N2FhYTNjMjRiZmI1MzhiYjkzZWU2MDIxM2FmMWJmZGFmNTgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.mlNuvs3BFmX5ohE-Dfnd06wPBpi1ByeoM1nxZp6l52Q)
> 
> Is there a particular issue with an implementation like
> 
> ```
> <Element data-test-text-link {...props} ref={ref}>
>    {typeof children === 'function' ? children(css.root) : children}
> </Element>
> ```
> 
> So that the class could only be applied as needed?

With such implementation it won't work as we expect ```
<Element data-test-text-link {...props} ref={ref}>
   {typeof children === 'function' ? children(css.root) : children}
</Element>
```

UladzislauKutarkin:
Hi @JohnC-80, @zburke  I made small updates to the code. I think the approach proposed by @JohnC-80  looks better and cleaner since we can control whether we apply `css.root` or not.

JohnC-80 (approved):

vashjs (approved):

PR #2193, opened 2024-01-04
Update ConfirmationModal prop types/documentation

ncovercash:
This fixes the following:
- button labels allow `node`s (via prop-types and the code itself defaults to `<FormattedMessage>` nodes), so the README must be updated
- `heading` is limited to a `string` since it's used as an aria-label; the prop-types must be updated
- `buttonStyle` and `cancelButtonStyle` were undocumented.

### DIFF
--- a/lib/ConfirmationModal/ConfirmationModal.js
+++ b/lib/ConfirmationModal/ConfirmationModal.js
@@ -16,7 +16,7 @@ const propTypes = {
   cancelButtonStyle: PropTypes.string,
   cancelLabel: PropTypes.node,
   confirmLabel: PropTypes.node,
-  heading: PropTypes.node.isRequired,
+  heading: PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


zburke (approved):

bvsharp (approved):

kjain110 (approved):

alb3rtino:
@ncovercash 

> heading is limited to a string since it's used as an aria-label; the prop-types must be updated

Can you please help me understand the necessity for this change? This change is causing warnings for `<ConfirmationModal>` throughout our codebase. From what i can see, the `heading` prop is passed to the `label` prop of `Modal` which explicitly does accept `node`:
```js
label: PropTypes.node.isRequired,
```

ncovercash:
@alb3rtino it looks like I pulled that information from the README, which appears to be incorrect upon further investigation. I'll make a PR to fix this.

PR #2194, opened 2024-01-04
Update Checkbox label prop documentation

ncovercash:
The `label` prop accepts a node in the prop-types and [the usage concurs](https://github.com/folio-org/stripes-components/blob/master/lib/Checkbox/Checkbox.js#L157); this brings the README in line with this.

### DIFF
--- a/lib/Checkbox/readme.md
+++ b/lib/Checkbox/readme.md
@@ -41,7 +41,7 @@ id | string | HTML id attribute applied to input - will also set the HtmlFor att
 inline | bool | Renders the checkbox inline | `false`
 innerClass | string | Adds a custom class name for the inner element of the component | `undefined`
 inputRef | object or func | Supplies a ref to the rendered `<input>` | `undefined`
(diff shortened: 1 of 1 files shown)


zburke (dismissed):
```
label: PropTypes.node,
```
So in fact this should be plain `node`, not `string, node`.

ncovercash:
> ```
> label: PropTypes.node,
> ```
> 
> So in fact this should be plain `node`, not `string, node`.

`PropTypes.node` includes `string` ([docs](https://github.com/facebook/prop-types/blob/main/README.md?plain=1#L78-L81), [source](https://github.com/facebook/prop-types/blob/main/factoryWithTypeCheckers.js#L474)), just as `ReactNode` contains `string`

Happy to add it if it makes it more readable, though!

JohnC-80 (approved):

bvsharp (approved):

kjain110 (approved):

PR #2195, opened 2024-01-04
Cleanup FilterGroups README

ncovercash:
This updates the FilterGroups README to list properties in tables, consistent with most other component documentation in FOLIO, and increasing readability.  It also restricts the `prop-types` of `label` to a `string`, which is required due to its usage as an ARIA label.  Finally, it also adds a few undocumented props (one of which is required!)

### DIFF
--- a/lib/FilterControlGroup/FilterControlGroup.js
+++ b/lib/FilterControlGroup/FilterControlGroup.js
@@ -7,10 +7,7 @@ const propTypes = {
     PropTypes.arrayOf(PropTypes.node),
     PropTypes.node,
   ]),
-  label: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.node,
-  ]),
+  label: PropTypes.string,
   style: PropTypes.object,
 };
 
--- a/lib/FilterGroups/FilterGroups.js
+++ b/lib/FilterGroups/FilterGroups.js
@@ -74,7 +74,7 @@ FilterGroup.propTypes = {
   disabled: PropTypes.bool,
   filters: PropTypes.object.isRequired,
   groupName: PropTypes.string.isRequired,
-  label: PropTypes.node.isRequired,
+  label: PropTypes.string.isRequired,
   names: PropTypes.arrayOf(PropTypes.object).isRequired,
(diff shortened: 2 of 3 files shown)


lib/FilterControlGroup/FilterControlGroup.js
```diff
@@ -7,10 +7,7 @@ const propTypes = {
     PropTypes.arrayOf(PropTypes.node),
     PropTypes.node,
   ]),
-  label: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.node,
-  ]),
+  label: PropTypes.string,
```
zburke, line 10: This will cause a prop-types warning for sth like 
```
value: <FormattedMessage id="whatever" />
```
If we're certain the value is gonna end up in an `aria-label` or some other DOM attribute, then we do indeed want  a string. Are we certain?

zburke (commented):
Are you certain about converting from `node` to `string`? I know it's just prop-types so the "real" impact here is nothing, but we should still be certain.

lib/FilterControlGroup/FilterControlGroup.js
ncovercash: Pretty certain: https://github.com/folio-org/stripes-components/blob/c6b916f1b1377cbf942d70241f7808f534d9ab4e/lib/FilterControlGroup/FilterControlGroup.js#L19

ncovercash: This is passed up the chain:
- `<FilterGroups>`'s `config`'s `label` is passed to `FilterGroup` (internal component) at https://github.com/folio-org/stripes-components/blob/c6b916f1b1377cbf942d70241f7808f534d9ab4e/lib/FilterGroups/FilterGroups.js#L299
- `<FilterGroup>` passes its label straight to `FilterControlGroup`: https://github.com/folio-org/stripes-components/blob/c6b916f1b1377cbf942d70241f7808f534d9ab4e/lib/FilterGroups/FilterGroups.js#L56

So the external `FilterGroups` prop had to change, too.

bvsharp (approved):

zburke (approved):

PR #2196, opened 2024-01-04
Document Popover children prop as required

ncovercash:
This prop is required, as denoted by the component's prop-types.  The README should denote this requirement.


zburke (commented):
`| Yes |` or `| | Yes |`? I don't think `Yes` makes sense as a default value.

ncovercash:
> `| Yes |` or `| | Yes |`? I don't think `Yes` makes sense as a default value.

oops, the lack of a side vertical line tripped me up.

zburke (approved):

bvsharp (approved):

PR #2004, opened 2023-03-16
STCOM-1137: Add auto placement to <Popper> component

ncovercash:
# [STCOM-1137](https://issues.folio.org/browse/STCOM-1137)

This change allows `auto` to be specified for popper component placements; this is a valid placement based on the [popper documentation](https://popper.js.org/docs/v1/#Popper.placements).

PR #2197, opened 2024-01-04
Fix Popper documentation consistency

ncovercash:
There are several props available to the `Popper` component which were not documented in the README.

Additionally, the placements `auto-start` and `auto-end` were missing from the `AVAILABLE_PLACEMENTS` array, despite being valid (looks like I missed these in https://github.com/folio-org/stripes-components/pull/2004, whoops).

### DIFF
--- a/lib/Popper/Popper.js
+++ b/lib/Popper/Popper.js
@@ -18,6 +18,8 @@ export const AVAILABLE_PLACEMENTS = [
   'right-start',
   'right-end',
   'auto',
+  'auto-start',
+  'auto-end',
 ];
 
 const [DEFAULT_PLACEMENT] = AVAILABLE_PLACEMENTS;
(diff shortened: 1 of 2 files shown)


zburke (approved):
Consider creating a Jira for this PR so we have way to document which release contains the `auto-start` and `auto-end` props.

JohnC-80:
Things might get interesting for this component when we look at migrating to floating-ui and its possibilities for placement... Will potentially mean even more docs changes, so if we want to hold on this, will be okay...

bvsharp (approved):

ncovercash:
@JohnC-80 IMO we should go ahead and merge this — the "new" placements are already accepted per the readme and supported by the code, just forgotten in the prop-types, so we're not adding something new

JohnC-80 (approved):

PR #2204, opened 2024-01-05
Add README for TextArea

ncovercash:
This adds a README for the `<TextArea>` component.  It also removes the `type` prop, as it is unused and does not exist on `<textarea>`, only `<input>`.  My hunch is that it was accidentally copied from `TextField`.


zburke (approved):
❤️

bvsharp (approved):

kjain110 (approved):

PR #2209, opened 2024-01-16
STCOM-1238 Accordions - apply content  z-index in reverse-registration order

JohnC-80:
### Problem 
Toggling accordions with open overlay components can cause z-index fighting of content and subsequent accordion headers.

### Approach
Render the content of accordions with stepping z-index - top accordion gets the highest and the last gets a z-index of 1. This keeps the content rendered in front of headers and content from subsequent accordions.

Accordions do *usually get registered in `BoundingClientRect.top` order with standard registration - but if any similar problem does re-appear, we can apply some more intensive sorting logic on the accordions.

Bonus points - we lose some state tracking/updating with the 'focus' events, which were inadequate for maintaining a dynamic z-index.

Before:
accordion-z-fix-before

After:
accordion-z-fix-after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * TextLink - underline showing up on nested spans with 'display: inline-flex'. Refs STCOM-1226.
 * Use the default search option instead of an unsupported one in Advanced search. Refs STCOM-1242.
 * Added support for clear icon in `<TextArea>`. Refs STCOM-1240.
+* Bug fix for Accordion content z-index rendering. Refs STCOM-1238.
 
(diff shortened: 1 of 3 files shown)


BogdanDenis (approved):

BogdanDenis:
Hey @JohnC-80, it looks like some accordions in some apps now experience this issue after this PR. For example: MARC Authority, Courses.

https://github.com/folio-org/stripes-components/assets/19309423/0f4fa4bb-b913-49c0-886d-26aaa3336f44


Could you take a look what is the issue there please?

PR #2210, opened 2024-01-17
STCOM-1165: Format aria attributes in `<Icon>`

UladzislauKutarkin:
[STCOM-1165](https://issues.folio.org/browse/STCOM-1165)
We have ensured that aria attributes are rendered correctly by formation via the util function.  All other components that were in the list in that story are working correctly since it pass `aria-label ` directly.


zburke (approved):
LGTM

JohnC-80 (approved):

PR #2214, opened 2024-01-30
STCOM-1256: Show action buttons in correct color

Dmitriy-Litvinenko:
## Purpose
Show action buttons in correct color

## Refs
https://issues.folio.org/browse/STCOM-1256

## Approach
Revert removing comma that was removed by mistake in https://github.com/folio-org/stripes-components/pull/2212/files#diff-a346a5e30d164673f59d1283343c9ba1b507f01e94e09b325ad2ac4fffedc6e6L29


zburke (approved):
Sharp eyes; good catch!

JohnC-80 (approved):
Thank you!

PR #2215, opened 2024-01-30
STCOM-766 Accessibility | Repeatable Field Component

vashjs:
When working with a RepeatableField component, it's required to shift focus to the newly added element (if there are focusable elements inside). Ref [STCOM-766](https://issues.folio.org/browse/STCOM-766)

STCOM-766


JohnC-80 (changes requested):
While I do like the elegance of this approach, the case where it's odd is if a row in the middle is removed - focus probably should not move in this case (just keep focus on the button of the row where it was clicked...)
repeatable_middle

vashjs:
> While I do like the elegance of this approach, the case where it's odd is if a row in the middle is removed - focus probably should not move in this case (just keep focus on the button of the row where it was clicked...) repeatable_middle [ repeatable_middle ](https://private-user-images.githubusercontent.com/20704067/301282415-a4b508a8-2954-4860-8f79-cf809ec88d79.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDY3ODI2NzEsIm5iZiI6MTcwNjc4MjM3MSwicGF0aCI6Ii8yMDcwNDA2Ny8zMDEyODI0MTUtYTRiNTA4YTgtMjk1NC00ODYwLThmNzktY2Y4MDllYzg4ZDc5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMDElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjAxVDEwMTI1MVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJlYWM2NTA1ZjdlNzEwOWRlNzExNTI3ZjNlMGJiY2ZlYWEwMTZlMjFkYTdkNzEyNmRhMzk3YTFhYmE0NTNkOTUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.IxvTjLD5A31ejbgJVuCNyzVsbsblyaxfJ1pr5j8dfIs) [ ](https://private-user-images.githubusercontent.com/20704067/301282415-a4b508a8-2954-4860-8f79-cf809ec88d79.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDY3ODI2NzEsIm5iZiI6MTcwNjc4MjM3MSwicGF0aCI6Ii8yMDcwNDA2Ny8zMDEyODI0MTUtYTRiNTA4YTgtMjk1NC00ODYwLThmNzktY2Y4MDllYzg4ZDc5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMDElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjAxVDEwMTI1MVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJlYWM2NTA1ZjdlNzEwOWRlNzExNTI3ZjNlMGJiY2ZlYWEwMTZlMjFkYTdkNzEyNmRhMzk3YTFhYmE0NTNkOTUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.IxvTjLD5A31ejbgJVuCNyzVsbsblyaxfJ1pr5j8dfIs)

Hi @JohnC-80, changed behavior based on your inputs + change class component to functional, because one day all of them will be moved anyway. So now we have 3 handled cases:

1. when user added something - focus will move to last added element 
2. when user removed something in middle, we will keep focus on same index where it was clicked (it's a little bit tricky and works only because we are using indexes as keys 😃, so in our case it's not a bug but feature )
3. when user removed last element in list, we don't do anything, because index with such element doesn't exist

Please check GIF:

STCOM-766

hooks/usePrevius/usePrevious.js
```diff
@@ -0,0 +1,13 @@
+import { useEffect, useRef } from "react";
+
+export const usePrevious = (value) => {
```
NikitaSedyx, line 3: usePrevious

SergeYvas (approved):

NikitaSedyx:
@vashjs you need to pin `polished` to make CI work

package.json
```diff
@@ -113,6 +113,7 @@
     "moment-range": "^4.0.2",
     "moment-timezone": "^0.5.14",
     "normalize.css": "^8.0.1",
+    "polished": "4.2.2",
```
NikitaSedyx, line 116: it's not a dependency, you just need a resolution

JohnC-80 (approved):

PR #2220, opened 2024-02-06
STCOM-1257 - Focused accordion should have highest z-index.

JohnC-80:
https://folio-org.atlassian.net/browse/STCOM-1257

in the story, Datepicker is cited, but this is an issue with Accordions/stacking order that was introduced in https://github.com/folio-org/stripes-components/pull/2209

Approach - Set the focused accordion's z-index to the length of the registering accordionset's list of registered accordions.

image


BogdanDenis:
Hey @JohnC-80, I like the appoach with `:focus-within`, but it still doesn't fix issues with dropdowns inside filter accordions in MARC Authorities app.
This is because when dropdowns inside accordions are opened - accordion's state doesn't change so `z-index` is not re-calculated.
I suggest to add an `onFocus` handler to the `data-test-accordion-wrapper` element, and re-calculate and store `z-index` in state, so that when focus shifts to an element inside the accordion it can re-render with new value

JohnC-80:
Thank you for the feedback @BogdanDenis good point... I've implemented the solution with `focus`/`blur` handlers now and stateful handling of the current `z-index` value.

BogdanDenis (approved):

PR #2223, opened 2024-02-09
STCOM-1262 MCL  - loading state bugfix. 

JohnC-80:
Experiencing this bug, loading/pagination buttons may remain disabled after new data has been received.

This was working correctly prior to #2106 And affects MCL's where pages are appended to the current set of `contentData` - a 'load more' setup.

The approach was to include non-sparse contentData with the same state update as sparse data.

Additionally - added tests for disabling the paging button and unskipped a handful of other MCL tests.


lib/MultiColumnList/tests/MCLHarness.js
```diff
@@ -36,9 +37,18 @@ class MCLHarness extends React.Component {
...
 
-  onNeedMore() {
+  async onNeedMore() {
+    const { onNeedMoreSpy, asyncPaging } = this.props;
+    if (onNeedMoreSpy) onNeedMoreSpy();
+    let newData;
+    if (asyncPaging) {
+      newData = await asyncGenerate(30, null, 500);
+    } else {
+      await Promise.resolve(syncGenerate(30)).then(data => { newData = data; });
```
vashjs, line 47: What about, just for consistency
`newData  = await Promise.resolve(syncGenerate(30));`

vashjs (approved):

EthanFreestone (approved):

PR #2226, opened 2024-02-16
STCOM-1192 Group checkboxes in FilterGroups (a11y)

JohnC-80:
https://folio-org.atlassian.net/browse/STCOM-1192

Related inputs are commonly group within a `<fieldset>` element or a `div` element using `role=group`. Both do similar things, assigning a top-level label. The downside to fieldset is that it has a built-in styling that has to be overridden and `<legend>` elements can also be stubborn for styling, although that is less of a matter. A `div[role="group"]` with an `aria-label` suffices.

`<Filtergroups>` often wouldn't be able to render a decent unique id due to objects (`<FormattedMessage>`) being passed as the label to the groups, despite propTypes dictating it be a `string`.  A small utility function was used to derive the string from a passed `FormatMessage` instance, converting it to a call to `formatMessage` fron `Intl` contenxt.

Proptypes of FilterGroups were expanded to relax common proptype errors in the console and we reduce the number of `id="[object Object]"` occurrences in the DOM.

image

From Snapshot Environment: `[object Object]` id's... egad!
image

Ah... no longer (after change)
image

### DIFF
--- a/lib/FilterGroups/FilterGroups.js
+++ b/lib/FilterGroups/FilterGroups.js
@@ -272,6 +272,7 @@ export function handleClearAllFilters() {
 const FilterGroups = (props) => {
   const { config, filters, onChangeFilter: ocf, onClearFilter } = props;
   const disableNames = props.disableNames || {};
+  const { formatMessage } = useIntl();
 
   const filterGroupNames = (group) => {
     const names = [];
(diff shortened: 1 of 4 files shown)


lib/FilterGroups/FilterGroups.js
```diff
@@ -314,7 +327,7 @@ FilterGroups.propTypes = {
   config: PropTypes.arrayOf(
     PropTypes.shape({
       cql: PropTypes.string.isRequired,
-      label: PropTypes.string.isRequired,
+      label: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
```
ncovercash, line 330: made a stripes-types edit https://github.com/folio-org/stripes-types/pull/55

ncovercash (approved):

PR #2230, opened 2024-02-27
Release 12.0.5  

JohnC-80:
* update loading state with difference in data length for both sparse and non-sparse arrays

* expand columns for MCL stories

* add async loading to MCL test harness

* test disabled state of loading button

* remove only

* log changes

* simplify MCL Harness code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change history for stripes-components
 
+## [12.0.5](https://github.com/folio-org/stripes-components/tree/v12.0.5) (2024-02-27)
+[Full Changelog](https://github.com/folio-org/stripes-components/compare/v12.0.4...v12.0.5)
+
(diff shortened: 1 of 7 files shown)


zburke:
@JohnC-80, TLDR, let's handle the code change in one PR and the release in another. 

This doesn't make sense to me as a release PR given that it contains code changes that need to be attached to a Jira ticket but none is mentioned in the PR description here. Typically, commits on a release branch are composed mostly of cherry-picks off `#master` and occasionally (but rarely) like this one with code changes that are relevant only to the branch. Still, in the latter, I would expect a PR analogous to how we manage things on master, i.e. pertaining only to the feature and not modifying the version. 

Is there a reason we couldn't just `git cherry-pick -x 0e06c1cfc160e025345aee6378dea934373a9ccc` to grab the commit from #2223?

PR #2238, opened 2024-03-05
STCOM-1094 Supply DayJS/moment conversion utilities.

JohnC-80:
This PR Adds functionality to stripes components to help UI modules convert from using the sunset `moment`, `moment-timezone`, and `moment-range` libraries.

These include utilties for 
* importing a pre-extended version of DayJS for use in the ui-module.
* loading DayJS locales asynchronously.
* creation, comparison of date ranges of DayJS objects.
* documentation for utilities and obtaining a localized date-time format (as opposed to obtaining that format from `moment` as was done previously`

Documentation for the utilities is included here: https://github.com/folio-org/stripes-components/blob/a4fb5fdf6802580e8ca4a89d4e05f627642cc237/util/DateUtils_readme.md

This PR can be merged as-is as it's all feature add with changes to existing components/functions.
https://github.com/folio-org/stripes-core/pull/1384 can be merged along with it, as that will load DayJS locales at the top level, of the whole FOLIO app, reducing the necessity for ui-modules to load locales within their own code.

Full Datepicker, Timepicker conversions. Will be included in future changes so that those can be fully tested before being released.

Hopefully we can get a great deal of conversion across the FOLIO platform done in 1-2 versions - but these tools should make it possible to convert during Ramsons development.


util/dateTimeUtils.js
```diff
@@ -1,14 +1,240 @@
...
+    this.start = dayjs(start);
+    this.end = dayjs(end);
+    this.isRange = true;
+  }
+
+  /**
+   * returns an array of contained dayjs day objects.
+   *
+   * @method
+   * @name DayRange#expand
```
zburke, line 51: `expand` feels like an odd name for this function that doesn't, uh, expand the range; it, just returns the range represented by `this` as an array of dayjs objects, right?

util/dateTimeUtils.js
```diff
@@ -1,14 +1,240 @@
...
+  isSame = (candidate) => {
+    return this.start.isSame(candidate.start, 'day') && this.end.isSame(candidate.end, 'day');
+  };
+
+  /**
+   * returns true if candidate is fully within or equal to the range. Can be used with single dayjs objects
+   * or date strings as well.
+   *
+   * @method
+   * @name DayRange#isEqual
```
zburke, line 81: `s/isEqual/contains/`

util/dateTimeUtils.js
```diff
@@ -1,14 +1,240 @@
...
+   * @param { DayRange|Dayjs }
+   * @returns {Bool}
+   */
+
```
zburke, line 85: Please remove this extra whitespace. I'm surprised lint didn't yelp about this. It's normally so cranky.

util/dateTimeUtils.js
```diff
@@ -1,14 +1,240 @@
...
+    } else {
+      return dayjs(candidate).isBetween(this.start, this.end);
+    }
+  };
+
+  /**
+   * returns true if candidate start or end is within the range, or if candidate is equal to the range.
+   *
+   * @method
+   * @name DayRange#isEqual
```
zburke, line 99: `s/isEqual/overlaps/`

util/dateTimeUtils.js
```diff
@@ -1,14 +1,240 @@
...
+/**
+ * loadDayJSLocale
+ * dynamically loads a DayJS locale and sets the global DayJS locale.
+ * @param {string} locale
+ * */
+export function loadDayJSLocale(locale, cb = noop) {
+  const parentLocale = locale.split('-')[0];
+  // Locale loading setup for DayJS
+  // 'en-US' is default and always loaded, so we don't even worry about loading another if the language is English.
+  if (locale !== 'en-US') {
```
zburke, line 198: Consider `const DEFAULT_LOCALE = 'en-US'`. This will have a few benefits:
* It becomes obvious when you're doing something "default-y"
* If we ever refactor this, it'll be more typo-proof

util/dateTimeUtils.js
```diff
@@ -174,3 +411,23 @@ export function removeDST(dateTime, timeFormat) {
...
+ * @returns {String}
+ */
+export function removeDSTDayJS(dateTime, timeFormat) {
+  const julDate = '2022-07-20';
+  const janDate = '2022-01-01';
+
+  const julOffset = dayjs(julDate).utcOffset();
+  const janOffset = dayjs(janDate).utcOffset();
+
+  const offsetDate = janOffset < julOffset ? janDate : julDate;
```
zburke, line 430: I understand this isn't new code since it parrots the logic from `removeDST` above, but: isn't `offsetDate` gonna be a constant value since it only deals with `julDate` and `janDate`, which are constants? Or does this rely on a global `dayjs` instance, whose locale can change, and therefore we can't pre-calculate this?

zburke (approved):
Code changes look good and new documentation 😍 . Please make sure the JSDoc values and function names in code align completely before merging.

util/dateTimeUtils.js
JohnC-80: Yes. The DST range of dates is different per locale. Part of the reason why DST is such a PITA.

zburke: "Yes" meaning "this relies on a global `dayjs` instance whose locale may change, and of course different locales == different DST dates because DST == 🖕🖕🖕"?

PR #2239, opened 2024-03-06
STCOM-1270 - add component for rendering blocks of presentational HTML

JohnC-80:
Adding `<CleanHTML>`  component as a centralized alternative for implementations that set `innerHTML` via refs of via `react`'s `dangerouslySetInnerHTML`.
https://github.com/folio-org/stripes-smart-components/blob/master/lib/Notes/NoteViewPage/components/NoteView/NoteView.js#L287-L290

## Approach
Make use of the `sanitize-html` library to clean potentially troublesome tags from content which requires no interactivity.

Storybook example depicted:
image


JohnC-80:
Closing in favor of initially using the dompurify library in places where user-provided markup is rendered.

PR #2240, opened 2024-03-08
STCOM-1238 Accordion vs overlay z-index issues.

JohnC-80:
### Problem
Accordions with overlay components could have their overlays overlapped by other accordions due to z-index issues.

### Approach
What we've previously done...
Accordions content would overlap overlays/dropdowns from previous accordions... (render with reverse z-index order to accordion index)
fixed, but..
Accordions content would overlap overlays/dropdowns from the following accordions...(have the focused accordion assume highest zindex from the set)
fixed, but
Accordions/overlays would go back to being overlapped if focus left the accordion or went to another pane...

In this PR, when focus moves to the accordion, it will find/assume the highest z-index among other accordions. It will retain this same z-index after it is blurred. If an accordion already has the highest z-index among accordions, it will not update its own. The z-indexes do continue to grow, but only for the life of a view... if navigation happens or accordions leave the DOM, all z-indices will be reset upon return.

accordionZResults

### DIFF
--- a/lib/Accordion/Accordion.js
+++ b/lib/Accordion/Accordion.js
@@ -122,11 +122,13 @@ const Accordion = (props) => {
   }).current;
 
   const handleFocusIn = () => {
+    // only update the accordion z-index if it does not contain focus or if it's not
+    // already the highest z-index among other accordions.
     if (!focused) {
       updateFocused(true);
       const highest = getHighestStackOrder();
       if (zIndex !== highest) {
-        updateZIndex(highest + 1)
+        updateZIndex(highest)
       }
     }
   }
--- a/lib/Accordion/Accordion.js
+++ b/lib/Accordion/Accordion.js
@@ -64,12 +64,22 @@ function getWrapClass(open) {
   );
 }
 
+/* The z-index requirements for accordions:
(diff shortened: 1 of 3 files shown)


lib/Accordion/Accordion.js
```diff
@@ -102,15 +114,28 @@ const Accordion = (props) => {
...
 
   const uncontrolledToggle = useRef(() => {
     updateOpen(current => !current);
   }).current;
 
-  const handleFocusIn = () => updateZIndex(accordionSet?.getNumberOfAccordions() || 2);
-  const handleFocusOut = () => updateZIndex(stackOrder);
+  const handleFocusIn = () => {
+    // only update the accordion z-index if it does not contain focus or if it's not
+    // already the highest z-index among other accordions.
```
zburke, line 126: Pedantic: `s/or/and/`

zburke (approved):
Thanks for the detailed technical explanation in the PR description. I'd love to see a bit of that documentation/explanation filter down into comments directly on Accordion.js itself so the next person looking at the code can, in fact, just look at the code, without needing to dig up this PR to understand why we had to do things this way.

PR #2255, opened 2024-03-27
STCOM-1274: Exclude invalid additional currencies

alisher-epam:
https://folio-org.atlassian.net/browse/STCOM-1274
Exclude invalid Uruguayan additional currencies

### DIFF
--- a/lib/CurrencySelect/tests/CurrencySelect-test.js
+++ b/lib/CurrencySelect/tests/CurrencySelect-test.js
@@ -32,7 +32,7 @@ describe('CurrencySelect', () => {
   });
 
   describe('utility functions', () => {
-    const CURRENCY_COUNT = 161;
+    const CURRENCY_COUNT = 159;
     it('expects currency maps to contain the same element counts', () => {
       expect(Object.keys(currenciesByCode).length).to.equal(CURRENCY_COUNT);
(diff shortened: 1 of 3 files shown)


BogdanDenis (approved):

usavkov-epam (approved):

lib/CurrencySelect/tests/CurrencySelect-test.js
```diff
@@ -32,7 +32,7 @@ describe('CurrencySelect', () => {
   });
 
   describe('utility functions', () => {
-    const CURRENCY_COUNT = 161;
+    const CURRENCY_COUNT = 159;
```
zburke, line 35: Man I hate myself for writing such a brittle, inadequate test 😧

zburke (commented):
Technically, this is fine, but I think we need better documentation than "this happened to not work with one banking provider" before we make this change, which is all that's provided in the ticket at present. I won't block this if folks want to move ahead with it, but this implementation feels ... insufficient**. e.g., even if everything here is technically correct, can we at least add some comments to `util/currencies.js` that describe how the backend API (_which_ backend API?) validates these values?

** My original impl, with its undocumented "exclude list" here, is also insufficient. This PR feels like an opportunity to improve documentation/understanding of how are APIs interact and it would be nice to take advantage of it.

PR #2269, opened 2024-04-16
STCOM-1282. Multiselection overlay overlapping input under certain conditions.

JohnC-80:
The positioning of `<MultiSelection/>`'s overlay element is completely controlled by the `popper` library. To affect the placement at the instance level, `popper`'s options include an object with 'modifiers' - a bunch of different adjustments that are applied/accounted for when the overlay element is rendered. One of them is 'preventOverflow' that keeps an element within the bounds of a certain scrollable parent/element. It defaults to the `scrollParent` of the input the overlay belongs to.

The case sited in [STCOM-1282](https://folio-org.atlassian.net/browse/STCOM-1282) has `<MultiSelection>` being used within a *short `<EditableList>`. The `scrollParent` - the `MCL`'s scrollable element - with only 3 items is not tall enough to render the overlay without it overlapping the `<MultiSelection>` control. The `renderToOverlay` prop is used as it should be on the `MultiSelection`, but `popper` still wants to stick to MCL's scrollable element as its boundary.

### Approach
This PR changes the setting applied to `popper` iin `<MultiSelection>` if the `renderToOverlay` prop is used, setting the `boundaryElement` of the `preventOverflow` modifier to the `div#OverlayContainer` rendered within the stripes UI.


artem-blazhko (approved):

zburke (approved):
I really appreciate the detail in the PR description. It would have been impossible to tell what was going, or why this is necessary without it. Thank you!

BogdanDenis (approved):

PR #2270, opened 2024-04-17
STCOM-1283 remove postcss-plugins as per STRWEB 111

JohnC-80:
This PR cleans up stripes-components' storybook webpack config as well as removes postcss-plugin dependencies as seen in https://github.com/folio-org/stripes-webpack/pull/142

Additionally, it removes imports of CSS variables across the repo, since those are no longer necessary - they are stripped out by `stripes-webpack` here-forward. The majority of the large number of files changed are simply removing the `@import '..variables.css'` or similar from CSS files.

An import of the CSS variables was also included within the test harness, as the entries (courtesy of `karma-webpack`) are not the same as our standard webpack config.

* This does require https://github.com/folio-org/stripes-webpack/pull/142 to be merged before tests will pass.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,9 +1,9 @@
 import React, { Component, Fragment } from 'react';
-import { IntlProvider } from 'react-intl';
 import pkg from '../package.json';
 import { themes } from '@storybook/theming';
 import IntlWrap from './IntlWrap';
 import '../lib/global.css';
+import '../lib/variables.css';
 
 /**
  * React intl support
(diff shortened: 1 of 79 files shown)


zburke (approved):
LGTM, and I appreciate the detailed explanation. I am curious about those test failures though. Do you think they could be an indication of similar failures we might see in ui-* repos after this change (in which case this PR _does_ change the public surface) or is it really just "we're adjusting internals and don't have everything quite aligned yet"?

JohnC-80:
Test failures here look to be primarily caused by `@media` queries not being present since they have been removed and are not injected in the version of `stripes-webpack` that's installing. The missing queries can cause differentiation where elements that were presented at a certain media range remain hidden. This is all due to the variable `@imports` being removed. Variables are filled in at the top of the test suite, but the custom-media-queries work differently and depend on the webpack configuration to inject them.

PR #2275, opened 2024-04-25
Upgrade yarn v1 to v4

JohnC-80:
https://folio-org.atlassian.net/browse/STCOM-1279


zburke:
We have closed this ticket as "Won't do". Details at [STRIPES-907](https://folio-org.atlassian.net/browse/STRIPES-907).

PR #2276, opened 2024-04-29
STCOM-744 - MCL - Virtualization bug with static data.

JohnC-80:
https://folio-org.atlassian.net/browse/STCOM-744

As displayed in the issue examples, MCL behaves strangely with virtualization enabled for a static list.

Virtualization isn't commonly used among FOLIO modules, but this functionality should be fixed anyhow in the hopes that one day we _might_ have better support virtualized lists, estimated length of query results, initial fetching of a page that exists in the middle of a long list, etc...

Investigation into this exposed fairly major bugs in MCL logic. If the usage of this feature were more common, I believe we would've heard more about this buggy behavior.


ViktorSoroka07:
@JohnC-80, glad to see progress on this. I thought the ticket eventually has been removed or something😎

JohnC-80:
@ViktorSoroka07 Glad to hear from you, even in brief! Yeah - not a great deal of requests for this in the application code, but one of these days, perhaps...

PR #2283, opened 2024-05-10
STCOM-1091 Upgrade Downshift dependency.

JohnC-80:
This PR updates our `downshift` dependency from v2 to v9 XD. At it's completion it will involve a refactor of `<Selection>`, `<MultiSelection`>, and `<AutoSuggest>`.


lib/Selection/Selection.js
```diff
@@ -1,27 +1,344 @@
...
+// of dataOptions, it's not grouped...
+const optionIsGrouped = (item, dataOptions) => {
+  return dataOptions.findIndex((i) => isEqual(i, item)) === -1;
+};
+
+const getControlWidth = (control) => {
+  if (control) {
+    return control.offsetWidth;
+  }
+  return undefined;
```
vashjs, line 35: is there any specific reason return underfined but not just do return? Maybe some eslint rules?

lib/Selection/Selection.js
```diff
@@ -1,27 +1,344 @@
...
+}
+
+const getItemClass = (item, i, props) => {
+  const { value } = item;
+  const { selectedItem, highlightedIndex, dataOptions } = props;
+  if (value === undefined) {
```
vashjs, line 41: can be simplefied to `if (!val) return;` or  `if (!val) return val;`

lib/AutoSuggest/AutoSuggest.js
```diff
@@ -1,217 +1,179 @@
...
+  renderOption = defaultRender,
+  renderValue = defaultRender,
+  required,
+  validationEnabled = true,
+  value,
+  valueKey = 'value',
+}) => {
+  const container = useRef(null);
+  const textfield = useRef(null);
+  const testId = useRef(id || uniqueId('autoSuggest-')).current;
```
vashjs, line 57: maybe we can use https://react.dev/reference/react/useId instead

vashjs (commented):
Hi @JohnC-80, first of all thank you for this change, Firebird team is waiting for this group supporting in Selection component. No question from code perspective, I've just added few minor comments/questions. 
One question about behavior related to grouped items, I see that now when items weren't found we still have groups available, just want to confirm that it's expected behavior? 

group example 1

I've checked some libraries, and here if there are no items under categories - categories titles not rendering at all, ex of react-select https://react-select.com/: 

group example 2

lib/AutoSuggest/AutoSuggest.js
JohnC-80: Hmm... it's not exactly a simple swap for `uniqueId`... since it's a hook, it can't be called conditionally. It's possible we can wrap it within our own hook to keep the component code concise. I'll make an update.

JohnC-80:
@vashjs Thank you for the review! I had the filtering issue you found fixed in an early version - I will re-implement that fix. Thanks!

vashjs (approved):

BogdanDenis (changes requested):

BogdanDenis (approved):
Please also update CHANGELOG
Other than that - changes look good!

PR #2285, opened 2024-05-13
STCOM-1291 avoid defaultProps in functional components

zburke:
`defaultProps` for functional components will be deprecated in React v19 in favor of ES6 default parameters.

Refs [STCOM-1291](https://folio-org.atlassian.net/browse/STCOM-1291)


zburke:
@JohnC-80 , there are a few test failures here that I can't suss. Can you take a look?

lib/Button/Button.js
```diff
@@ -31,57 +31,60 @@ export const propTypes = {
...
-  type: 'button',
-};
-
-const Button = React.forwardRef((props, ref) => {
+const Button = React.forwardRef((
+  {
+    align,
+    allowAnchorClick,
+    autoFocus,
+    bottomMargin0,
```
JohnC-80, line 39:
```suggestion
    allowAnchorClick,
    bottomMargin0,
```


Since `autoFocus` is being destructured out here, it isn't applied to the button when all of the `inputCustom` props are gathered up down in line 80.

zburke: 🤦

JohnC-80 (approved):

PR #2302, opened 2024-06-18
STCOM-1301 Color contrast in focus styles, Callout styles.

JohnC-80:
[STCOM-1301](https://folio-org.atlassian.net/browse/STCOM-1301)

An accessibility audit item from LOC mentioned the low contrast of current focus indicators among the main navigation buttons. This PR fixes that, making the border colors stand out on those links.
Styles were also adjusted for callouts to support *browsers that don't yet have full support for relative color syntax...

Before:
image

After:
image

Also adjusted:
Callouts in Firefox
image


lib/variables.css
```diff
@@ -2,8 +2,8 @@
   --color-text: #000;
   --color-text-p2: rgba(0 0 0 / 62%);
   --color-border-form: rgba(0 0 0 / 42%);
-  --color-border: rgba(0 0 0 / 20%);
-  --color-border-p2: rgba(0 0 0 / 10%);
+  --color-border: color-mix(in oklch, currentcolor 50%, transparent);
+  --color-border-p2: color-mix(in oklch, currentcolor 15%, transparent);
```
zburke, line 6: Above we moved from `oklch(from var(--error) 90% 11% h)` to `color-mix(in hsl, var(--error) 25%, var(--bg))`. Is it ok to use `oklch` notation here?

zburke (approved):
Contrast improvements look obviously good, but I'm unclear on why we are removing oklch in some places and adding it in others. Even if we agree it's best, if oklch support is not sufficiently widespread across browsers, don't we need to remove it everywhere?

```
stripes-components(master)
$ grep -r oklch ./lib
./lib/Card/Card.css:  background-color: color-mix(in oklch, var(--bg) 80%, var(--bg-hover) 20%);
./lib/variables.css:  --bg-hover: color-mix(in oklch, oklch(from var(--bg) calc(l - 60) c h) 10%, var(--bg) 90%);
./lib/Datepicker/Calendar.css:      background: color-mix(in oklch, var(--bg) 80%, var(--bg-hover-native) 20%);
./lib/Callout/Callout.css:    background-color: oklch(from var(--error) 90% 11% h);
./lib/Callout/Callout.css:    background-color: oklch(from var(--success) 90% 12% h);
./lib/Callout/Callout.css:    background-color: oklch(from var(--warn) 90% 9% h);
./lib/Button/Button.css:          border-bottom-color: oklch(from var(--primary) calc(l - 0.03) c h);
./lib/sharedStyles/form.css:    background-color: oklch(from var(--warn) l c h / 5%);
./lib/sharedStyles/form.css:    background-color: oklch(from var(--error) l c h / 5%);
./lib/sharedStyles/form.css:    background-color: oklch(from var(--success) l c h / 5%);
./lib/sharedStyles/form.css:    background-color: oklch(from var(--primary) l c h / 20%);
```

lib/variables.css
JohnC-80: Good question... let me play with the `hsl` ones some more and see if `oklch` will suffice there (it should) the relative color syntax `(from foo` was the main thing that FF didn't agree with. not the `oklch`

JohnC-80: Looks like this is one of those strange cases where `oklch` throws a bit of a loop :/  I'm switching these to `oklab` for now for consistency among the 3 colors there... possibly pushing in in that direction for variables that do more than grayscale tones...

IOW, it's not oklch alone, but the combination of `color-mix`, `oklch` and `#070` (green) to `#fff` (white) that's behaving strangely here...
To interpret the low percentages of green within white, `oklch` is wanting to pass through the `pink` end of the spectrum. 

`oklchwtf`:
image

JohnC-80: There is a difference between the handling of oklch in FF/Chrome :/ FF handles the above code fine and presents a nice, minty lightened green instead of the where-did-you-come-from-orange-pink....

PR #2306, opened 2024-06-27
STCOM:1305: Allow MCL prev-next pagination to display loading states for previous and next buttons

EthanFreestone:
feat: MCL PrevNext loading control

Added loading controls to MCL to allow direct loading access on prevNextPagination component via props:
`pagingCanGoNextLoading` and `pagingCanGoPreviousLoading`. This will apply a loading spinner to the relevant buttons in cases where the frontend is waiting to understand if a previous or next page is available.

The use case for ERM is that for performance reasons we needed to implement a `stats=false` call in a couple of places, so we don't know until we fetch a given page whether there is or is not data for it.

In our case, we "prefetch" a page or two ahead, and so while those are loading we need to make the user aware that there _could_ be a next page, but we don't know yet. Hopefully most of the time this will be under a second, but in extreme circumstances we could see a 10 second or longer wait, hence the need for user feedback


JohnC-80 (changes requested):
Please update the changelog and associate this with a JIRA ticket (if possible)

JohnC-80 (approved):

PR #2316, opened 2024-07-08
STCOM-1312 Update Selection with value from external changes.

JohnC-80:
If the component's `value` is set externally, the component would need to update its own internal state and call any coordinating change handlers. Added tests for this scenario.


zburke (approved):
Code changes LGTM. IIUC, this bug is likely the result of a missed use-case from the recent downshift update in [STCOM-1091](https://folio-org.atlassian.net/browse/STCOM-1091)/PR #2283; we should link up the tickets if so.

PR #2317, opened 2024-07-09
STCOM-1311 MultiSelection setting value prop after render should be reflected in component state.

JohnC-80:
https://folio-org.atlassian.net/browse/STCOM-1311

Round one of fixes addressed deduplication of selected items by using `isEqual` when determining if the chosen item is already selected or not, then it called downshift's own state management functions to handle addition or removal of the chosen item.

These problems (Multiselect not removing items, not resetting) ran a bit deeper.

Item removal was calling downshift's own 'removeSelectedItem', but the function within that was using strict equality against the result of a utility function (`itemToKey`) that needed to be overridden - otherwise it amounts to `object === object` which can be true but rarely and odd because all keys/values would be equal - so not intuitive for this. The override of `itemToKey` to use `itemToString` bakes the item down to a single string that strict equals is more likely to be affective with.

For the issue with affecting the value from outside of the component, the approach was to create a sentinel value so that the logic could differentiate between changes that happen outside of the component and changes that happened internally. Without it, having an empty value would always be different from the 'selectedItems' and always be reset before render so that no items could ever be selected or state couldn't change from its initial value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@
 * Use `isEqual` to dedupe multiSelection values list rather that `===`. Refs STCOM-1311.
 * Set `<MultiSelection>`'s popper modifiers to avoid overlap with the control when rendered within an Editable list. Refs STCOM-1309.
 * Conform `<Selection>`'s internal state when value prop changes after initial render. Refs STCOM-1312.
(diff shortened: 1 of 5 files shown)


zburke (approved):
I don't have good knowledge of Downshift's APIs, so the changes here are fairly opaque to me. The details in the PR description tell me you have great knowledge of those APIs and I approve this based on the strength of that description.

PR #2318, opened 2024-07-09
STCOM-1313 Update Selection internal state when dataOptions are received

JohnC-80:
https://folio-org.atlassian.net/browse/STCOM-1313

Since the `selectedItem` was originally determined early on, it formed a race condition if `dataOptions` still had yet to populate since they were being obtained dynamically. This PR makes it so that when the `dataOptions` do load, the state will be updated and the component will make another attempt to find the selected item.

Additionally - there as an error message mentioning that `selectedItem` was transitioning from 'uncontrolled' to 'controlled' - I believe this was due to the `getSelectedItem` code returning undefined (er, something that amounted to `undefined` like `arrayVar[-1]`) instead of `null` - let's not talk about the overbearing error messages... like it has an `onChange`, whadayawant?

Additionally, fixed a minor but very annoying message about the translation string being unable to be generated.

### DIFF
--- a/lib/Selection/Selection.js
+++ b/lib/Selection/Selection.js
@@ -1,4 +1,4 @@
-import React, { useState, useMemo, useCallback } from 'react';
+import React, { useState, useMemo, useEffect, useCallback, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { useIntl } from 'react-intl';
 import { useCombobox } from 'downshift';
@@ -121,6 +121,7 @@ const Selection = ({
   const { formatMessage } = useIntl();
   const [filterValue, updateFilterValue] = useState('');
   const [selectedItem, updateSelectedItem] = useState(value ? getSelectedObject(value, dataOptions) : null);
+  const dataLength = useRef(dataOptions?.length || 0);
   const controlRef = useProvidedRefOrCreate(inputRef);
   const options = useMemo(
     () => (asyncFilter ? dataOptions :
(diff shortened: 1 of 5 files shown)


lib/Selection/Selection.js
```diff
@@ -260,7 +270,7 @@ const Selection = ({
         })}
         onClick={() => {}}
         onChange={(e) => updateFilterValue(e.target.value)}
-        aria-label={formatMessage({ id: 'stripes-components.selection.filterOptionsLabel', values: { label } })}
+        aria-label={formatMessage({ id: 'stripes-components.selection.filterOptionsLabel' }, { label })}
```
zburke, line 273: This is so subtle. Good catch!

lib/Selection/utils.js
```diff
@@ -52,10 +52,13 @@ export const getSelectedObject = (value, dataOptions = []) => {
       const flattenedOptions = flattenOptionList(dataOptions);
       // in the case of RF, nothing selected, so cursor the 1st...
       const valueIndex = flattenedOptions.findIndex((o) => o.value === value);
+      if (valueIndex === -1) {
+        return null;
+      }
       return flattenedOptions[valueIndex];
```
zburke, line 58: This kind of API drives me bonkers. We're not writing C in 1978. `findIndex` should return a valid index, >= 0,  or throw an exception. Avoiding the conditional would make it harder to shoot yourself in the foot (-1 is truthy but invalid, 0 is falsy but valid, 1 is truthy and valid 🤯 ) and generally improve/clarify everything.

zburke (approved):
Lots of good fixes, and great explanations!, here. Thank you. At the same time, it makes me worry about what other edge cases we may have missed, whether in Selection or in other components involved in The Great Downshiftification that was #2283.

PR #2329, opened 2024-08-02
STCOM-1327: Datepicker: outputBackendValue={false} causes crash on date entry.

EthanFreestone:
(Branch was misnamed)

Fixes issue where outputBackendValue={false} would cause crash on date entry.
Also introduces a `formatInvalidDates` boolean to avoid needing to replicate all of DatePicker's internal logic around date formatting/parsing just to be able to validate whether a date is valid or not externally.

This is an issue when a Date field is non-required, as the default behaviour will set the input value to '' when an invalid date is typed, which for a non-required field is a valid entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change history for stripes-components
 
 ## 12.2.0 IN PROGRESS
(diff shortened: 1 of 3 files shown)


JohnC-80:
Closing in favor of #2331

PR #2331, opened 2024-08-07
Datepicker - invert validation logic. Provide props for validation workflow

JohnC-80:
Exporting a set of props for Datepicker - `datePickerAppValidationProps` That could be applied in the instance where validation logic will be implemented outside of the component for user feedback.

Example usage:

``` 
<Datepicker
  label="myDateField"
  value={someStateDateValue}
  onChange={someStateSetter}
  {...datePickerAppValidationProps }
/>
```


EthanFreestone (approved):
LGTM, thanks @JohnC-80 and apologies for the delays on my end

PR #2334, opened 2024-08-13
STCOM-1329 expose aria-label for SearchField Index <Select>

JohnC-80:
This will provide the ability for ui-modules to assign discernable, distinct labeling to their `<SearchField>`. 

UI modules with sub-navigations within the search & filter pain should supply distinct labels using the `aria-label` and `indexLabel` props of `<SearchField>`.


Dmytro-Melnyshyn (approved):

artem-blazhko (commented):
Could you update changelog?

PR #2341, opened 2024-08-28
STCOM-1334 Refactor Modals to not use react-overlays.

JohnC-80:
https://folio-org.atlassian.net/browse/STCOM-1334

This PR represents a step in the progress to eliminate the `react-overlays` dependency.
It implements focus-trapping using the `focus-trap` library, limiting focus to the `div#OverlayContainer` where all `<Modal>`s are rendered. This enables components within modals to use react portals to the Overlay container and still remain functional.

Additional features covered that react-overlays previously provided:
- Modal closure with the Escape key.
- Focus-management - moving focus into the modal, restoring focus to the modal trigger...
- Focus-trapping - tab focus cycles within elements of the modal.

This implementation resolves:
- Screen-reader trapping - while keyboard users were unable to tab out of the Modal before, screenreaders could still explore the DOM outside of the modal, which is considered an a11y flaw. This is implemented by applying the `inert` prop to FOLIO's main navigation and any siblings of the `div#OverlayContainer`, encompassing the interactive parts of the UI exist outside of the `OverlayContainer`.
- labelling issues. In most cases, Modals were incorrectly labeled or missing a reference to their internal Headline component.

Any/all ui-modules that use the `enforceFocus={false}` setting should be able to remove it. Components with Portal rendering will no longer break/be broken due to previous focus-trapping. This is the primary impetus for this PR, as this issue prevents our Modals from upholding accessibility across the board.

image

Additional:
Platform-wide id's/selectors are exported as consts from `stripes-components` these will be ingested by `stripes-core` in a future PR to that repo.

Future work.
Consolidate usage of the `OverlayContainer` string to exports from the single `consts.js` file created here.
We still have usage of react-overlays within our `useClickOutside` utility and usage of their `Portal` component within `Selection` to eliminate.
These are slightly lower priority as they don't place a burden on accessibilty as the problems with modals do.


lib/Modal/useOverlayContainer.js
```diff
@@ -0,0 +1,13 @@
+// returns or creates a div#OverlayContainer
+import { useRef } from 'react';
+
+export default () => {
+  const element = useRef(document.getElementById('OverlayContainer'));
```
zburke, line 5: Should we `export const OVERLAY_CONTAINER = 'OverlayContainer';` from somewhere? It's kinda terrifying how often that string shows up in this repo 😬 .

lib/Modal/Modal.js
```diff
@@ -96,115 +99,280 @@ const Modal = forwardRef((props, ref) => {
...
+    let modalContained = false;
+    // collect modals... check among multiple to prevent multiple modals from fighting over focus.
+    const modals = document.querySelectorAll('[class^=modalRoot--]');
+    if (modals.length === 0) {
+      modals.forEach((m) => {
+        if (m.contains(document.activeElement)) {
+          modalContained = true;
+        }
+      })
+    }
```
zburke, line 162: `s/===/>/` I think? There's no value in iterating if length is 0.

lib/Modal/useOverlayContainer.js
JohnC-80: That's a great suggestion. Will include a utility with this PR and set up future work for a conversion across the library.

lib/Modal/Modal.js
JohnC-80: Necessary refactor overall here... thank you!

lib/Modal/Modal.js
```diff
@@ -1,23 +1,65 @@
...
+  ));
+
+  elementList.current = [mainNav, ...overlaySiblings];
+
+  const apply = useCallback(() => {
+    elementList.current?.forEach(el => el?.setAttribute?.('inert', ''));
+  }, [elementList])
+
+  const release = useCallback(() => {
+    elementList.current?.forEach(el => el?.removeAttribute?.('inert'));
```
BogdanDenis, line 56: In `useEffect`, after `elementList` changes it will be a new value.
So if we set the attribute to one element, then that element was removed from DOM, `useEffect` would be called with an empty list and we wouldn't be able to remove them.
But it's DOM and upon removal it will be cleared anyways...maybe I'm overthinking this

BogdanDenis (commented):
Hey @JohnC-80, the good news is that it looks like the `useClickOutside` hook is not used anywhere
https://github.com/search?q=org%3Afolio-org%20useClickOutside(&type=code

lib/Modal/Modal.js
JohnC-80: It's possible we could just find the elements within each function. Not quite as DRY as we'd like it, but it's also not very many elements. Thanks for mentioning it!

JohnC-80:
> Hey @JohnC-80, the good news is that it looks like the `useClickOutside` hook is not used anywhere https://github.com/search?q=org%3Afolio-org%20useClickOutside(&type=code

I think I mistyped the utility. `useRootClose` is the last thing that's in use and comes from `react-overlays`. Not `useClickOutside`. `useRootClose` is at least imported in our dropdown menus.

lib/Modal/Modal.js
JohnC-80: I've modified the function to just select its elements at runtime.

BogdanDenis (approved):

PR #2345, opened 2024-09-11
STCOM-1340 Improve rendering performance of Selection with long lists.

JohnC-80:
In the particular eholdings instance, the `Selection` was ingesting 2000 options and taking a while to render. This PR implements a hook to take the more expensive calculations that have to loop through the array and executes them within a promise, resolving with the results. Large data-sets are diced into 100 item chunks and cycled through the functions such as `onFilter` that handle them - so we can keep the existing logic. Additional changes were included to ensure the list doesn't render when the Selection dropdown is closed.


JohnC-80:
closing in favor of #2346

PR #2349, opened 2024-09-19
STCOM-1334: Modal - add 'open' attribute to dialog.

Dmytro-Melnyshyn:
## Description
Currently, all tests that use `getByRole` to find elements in `Modal` fail (ui-quick-marc, ui-inventory,...). This is due to the lack of the `open` attribute on the `dialog` tag. The `open` property indicates whether the `dialog` is available for interaction.

### DIFF
--- a/lib/Modal/Modal.js
+++ b/lib/Modal/Modal.js
@@ -191,6 +191,7 @@ const Modal = forwardRef((props, ref) => {
           onClick={handleBackgroundClick}
         />
         <dialog
+          open={open}
           className={classNames(css.modalRoot, css[status])}
           ref={modalElem}
           onFocus={focusEntered}


BogdanDenis (approved):

artem-blazhko (approved):

zburke (commented):
The practical nature of this fix notwithstanding, it specifically goes against [the `<diaglog>` documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog): 

> It is recommended to use the `.show()` or `.showModal()` method to render dialogs, rather than the `open` attribute. If a `<dialog>` is opened using the `open` attribute, it is non-modal.

It's possible we introduced a problem in the refactor in PR #2341 and we need to fix it, but it isn't clear this is the right fix. It's also possible I'm overthinking this fix/under-understanding our implementation (e.g. our impl appears to duplicate native functionality like keyboard handling, and maybe other things too) and the change here is both necessary and sufficient. 

@JohnC-80 , can you provide the context I lack?

PR #2351, opened 2024-09-20
STCOM-1335 add `StripesOverlayContext` to Modal, MCL... implement in Popper

JohnC-80:
## [STCOM-1335](https://folio-org.atlassian.net/browse/STCOM-1335)

We've seen a common issue with 'Overlay' components (`Selection`, `Datepicker`, etc) where they end up overlapped by the boundaries of their scrollable containers. Common among these, a `usePortal` boolean prop can be used to render them into the centralized 'OverlayContainer' element, but it isn't always so ergonomic to add or to know when it should be added.

The functionality in this PR adds a context to common containers where overflow can be problematic: Modals and MCL's. The main implementation exists within the `Popper` component, which is consumed and used by all of the 'Overlay' components. When any of the 'Overlay' components are used within a Modal or an MCL, they will automatically behave as thought a `usePortal` prop is applied.

### Approach
Wrap `Popper` in an HOC (it's still a class-based component) that reads the `StripesOverlayContext`. When we do upgrade from the PopperJS library to `floating-ui`, we'll refactor to a functional version which will clean things up more.

Additonally, since the `useOverlayContainer` hook has a necessity across components, it's moved into the `hooks` folder now.

### DIFF
--- a/tests/helpers/getHookExecutionResult.js
+++ b/tests/helpers/getHookExecutionResult.js
@@ -1,4 +1,6 @@
-import React from 'react';
+import React, { useEffect, useState, Fragment } from 'react';
+import isEqual from 'lodash/isEqual';
+import noop from 'lodash/noop';
 import { mountWithContext } from '../helpers';
 
 /**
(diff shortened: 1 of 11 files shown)


tests/helpers/getHookExecutionResult.js
```diff
@@ -1,6 +1,22 @@
...
 import { mountWithContext } from '../helpers';
 
+
+const objectCompare = (current, candidate) => {
+  if (current === candidate) {
+    return true;
+  }
+
+  if (typeof current === 'object' && typeof candidate === 'object') {
+    return JSON.stringify(current) === JSON.stringify(candidate);
```
BogdanDenis, line 13: Why not use `lodash.isEqual`? `JSON.stringify` is sensitive to order of properties

BogdanDenis (approved):

PR #2354, opened 2024-09-30
STCOM-1348 Keyboard shortcut modal missing scrollbar.

JohnC-80:
## [STCOM-1348](https://folio-org.atlassian.net/browse/STCOM-1348)

During refactor of our modals, an inline style was added, hiding the overflow of the content div. This PR just removes that inline style. Tested with the `ui-plugin-find-users` modal as well as the applicable keyboard shortcuts modal.


Dmytro-Melnyshyn (approved):

zburke (approved):
@JohnC-80, this PR explains the solution (remove the offending `style={{ overflow: 'hidden' }}` line) but it would be nice to know why that line was added in the first place  in STCOM-1334/PR #2341. Indisputably, it causes the scroll problem, but presumably it solved some problem as well. Will that problem now resurface or are you confident it's just cruft?

JohnC-80:
I'm pretty confident of cruft here. It was not present on the modal content pre-refactor and conflicts with [the overflow: auto built-in styling of ModalContent](https://github.com/folio-org/stripes-components/blob/master/lib/Modal/Modal.css#L73) Likely, it was added for testing/ensuring Overlay problems/functionality in a pre- #2351 world.

PR #2355, opened 2024-10-01
CVE-2022-37601 bump it up

zburke:
There's a slew of CVEs that stem from outdated versions in `yarn.lock` and some cruft in `package.json` that should've been removed ages ago:

* `hoist-non-react-statics` hasn't been necessary since we yanked out the `injectIntl` hoc in 2018 in PR #696
* `@mdx-js/loader` I think is leftover from some early storybook work but storybook builds happily without it so :shrug:

Refs CVE-2022-37601


JohnC-80:
> * `hoist-non-react-statics` hasn't been necessary since we yanked out the `injectIntl` hoc in 2018 in PR [Remove injectIntl() fork #696](https://github.com/folio-org/stripes-components/pull/696)

`stripes-core` is inquiring and wants to know since it actually does use `hoist-non-react-statics`...

> * `@mdx-js/loader` I think is leftover from some early storybook work but storybook builds happily without it so 🤷

Indeed... this was all storybook.

JohnC-80 (approved):

PR #2358, opened 2024-10-02
STCOM-1345 `<Selection>` - Call `onFilter` in side-effect for `asyncFilter`

JohnC-80:
## [STCOM-1345](https://folio-org.atlassian.net/browse/STCOM-1345)

Shoring up this component feature on this side of the refactor. The boolean prop `asyncFilter` will let the component know that option filtering will happen via request/not within the component. So it will just render the `dataOptions` prop as it is provided. Emptying the list and setting a `loading` prop are the responsibilities of the application.


Dmytro-Melnyshyn (approved):

translations/stripes-components/en.json
```diff
@@ -104,6 +104,7 @@
   "button.duplicate": "Duplicate",
   "button.new": "New",
   "exportToCsv": "Export to CSV",
+  "selection.enterFilter": "Enter filter to load possible options",
```
zburke, line 107: Would "Enter a value to load..." be more clear?

zburke (approved):
Code changes LGTM but please fix the typo in the docs

PR #2363, opened 2024-10-08
STCOM-1366 - Selection aggressive 'required' validation.

JohnC-80:
## [STCOM-1366](https://folio-org.atlassian.net/browse/STCOM-1366)

React Final Form and redux form validate values using an onblur event. This can cause unwanted validation messages in the UI if the control is shifting focus to  its open options list, when the user hasn’t yet had a chance to select a value yet.

### Approach
Using refs for the options list containing element and the control itself, check to see if focus is still within the bounds of the component. If it isn't, trigger the provided `onBlur` if any, and close the menu if necessary.

Before: The control value is validated before any value has been selected, when the menu is opened.
onblur-val-problem

After: Validation only occurs when the user 
onblur-val-fix

When the filter input is blurred naturally (tab key) the menu is closed and the `onBlur` handler is triggered.

Additionally - I noticed issues with keyboard navigation of items in the list and updated dependencies on rendering functions accordingly.


zburke (approved):
Thank you for the helpful comments explaining what the problem is, how the solution works, and especially the details of why the solution is written as it is with setTimeout. It totally makes sense but is not obvious at first. ❤️

PR #2371, opened 2024-10-21
STCOM-1341 Change Repeatable fields component's focus behaviour

vashjs:
Since the mechanism for focusing elements in repeatable fields is tied to a state change, in some parts of the application this caused a problem. For example, in edit forms, when filling in the initial values ​​of the form - elements could be located far at the bottom of the page, which provoked focusing of these elements and scrolling to them. Refs [STCOM-1341](https://folio-org.atlassian.net/browse/STCOM-1341)

To solve this problem, 2 updates were added:

- prevent scrolling to the focused element. This is a mechanism that should have been made in the original implementation, since when working with repeatable fields there is no need to scroll anywhere.
- focus added element if it's already ony element inside component focused (add button, etc)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,8 @@
 * Use `<Selection>`'s formatter for rendering the selected value within the field and prevent overflow of text. Refs STCOM-1344.
 * Export `<StripesOverlayWrapper>` that will automatically apply the `usePortal` behavior to nested overlay components. Refs STCOM-1353.
 * Properly pass `readOnly` prop to `<RadioButton>`'s internally rendered `<Label>`. Refs STCOM-1367.
-* Change repeatable fields focus behaviour. Refs STCOM-1341.
+* `TextArea` - move focus to the field after clearing the field by clicking on the `x` icon. Refs STCOM-1369.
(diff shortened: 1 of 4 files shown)


JohnC-80:
@vashjs Would the 'isFocused' check alone be enough to accomplish this - as in, could you just prevent focus from moving to the component if focus isn't already within the body of the component?

lib/RepeatableField/RepeatableFieldContent.js
```diff
@@ -6,13 +6,12 @@ import { getFirstFocusable } from "../../util/getFocusableElements";
...
+
   const callbackRef = useCallback((node) => {
-    if (node) {
+    if (node && isFocused) {
       const elem = getFirstFocusable(node, true, true);
 
-      if (isFocused) {
-        elem?.focus();
-      }
+      elem?.focus({ preventScroll: true });
```
JohnC-80, line 14: The new changes look good, @vashjs -
`preventScroll: true` is probably unnecessary here. Best to remove this to uphold typical browser behavior for when it is necessary.

JohnC-80 (approved):

PR #2389, opened 2024-11-12
STCOM-1382 Assign modal escape key handler to modal element rather than document

JohnC-80:
## [STCOM-1382](https://folio-org.atlassian.net/browse/STCOM-1382)

See [UIQM-732](https://folio-org.atlassian.net/browse/UIQM-732) for the module-level bug.

Modal hangs its escape key listener on the `document` rather than one of its owned elements. this can be problematic if an app has a `escape` shortcut key that's set up the same way. See the UI-QuickMarc bug for details. TLDR: `escape` key in dirty form should trigger the confirmation modal, but instead it opens the modal and immediately closes it, exiting the form without any confirmation.


BogdanDenis (approved):
Thanks @JohnC-80!

zburke (approved):

zburke:
Please update the commit subject with the JIRA slug when you merge. Release scripts use the commit subject, not the PR title, to find tickets in Jira and set the fixVersion field. Thank you!

PR #2393, opened 2024-11-19
STCOM-1385 Clear filter value after an action is chosen from MultiSelection

JohnC-80:
## [STCOM-1385](https://folio-org.atlassian.net/browse/STCOM-1385)

Known locations affected: Tags (linked in main JIRA), [Agreements->Settings ->Supplementary properties](https://folio-org.atlassian.net/browse/STSMACOM-877) (closed as duplicate)

### Approach
Multiselection needs to reset the filter value when an option's `onSelect` handler is called (actions API for Multiselection) Otherwise, the filter value will not be reset as it usually is through typical value selection logic.


zburke (approved):
Code changes LGTM. Is it possible to capture this behavior in a new test, e.g. to assert that the filter is empty after one of thes events on lines 263/264/265 fires? 

Unrelated, I'm not usually a fan of "while you're in there..." changes, but is it maybe worth fixing the comment on line 276/277? That hardly feels worth it's own PR, but it's definitely wrong.

OleksandrHladchenko1 (approved):

PR #2395, opened 2024-11-22
STCOM-1386 Paneset: check for existing id before registering pane

JohnC-80:
Panes can exhibit strange behavior in the dev environment that just doesn't happen in production. I believe this is caused by pane registration misbehaving in React StrictMode (where mounting happens twice). This PR dedupes the panes by checking if a pane id already exists in the paneset - and if it does, the pane is not added.

Problematic behavior can be seen specifically with panes that have `defaultWidth: "fill"` - they'll be registered twice, and their width calculation (which divides the remaining width among other `fill` panes) will leave them being half the size they should be. Hiding a search/filter pane can leave the results view at only 50% of its width, when it should be the full width of the view when it's the only pane.


zburke (commented):
Your description of the symptom sounds correct. Bravo! I have seen this and been totally mystified by it. 

Are you confident this the Most Correct fix? I worry a little bit that we are treating the symptom here (double registration) instead of a root problem (missing cleanup function).

zburke (approved):

JohnC-80:
The react-best practice tenant is applicable here - ` ensure mount/setState function can run twice w/o causing issue.`  Deduplication as implemented here seems like the best way. Other techniques such as avoiding the `setState` callbacks and pushing the calls into the `setState` function itself still resulted in duplicate items in state passed to code that would actually calculate the necessary style updates. This way, that never happens.

zburke (approved):

PR #2400, opened 2024-12-02
STCOM-1387 ExportCSV download link not working in Modals

JohnC-80:
## [STCOM-1387](https://folio-org.atlassian.net/browse/STCOM-1387)

The focus-trapping logic of Modal (via `focus-trap`) causes the `click()` event not to trigger on the download link rendered by `ExportCSV` because the link is appended outside of the focus-trapped element (in the body.)

Approach: 
Rendering the link within the `div#OverlayContainer` places it in within the `focus-trapping` boundary. Since this element is always present in the stripes UI, it's okay to render here. For tests where the `OverlayContainer` may not exist, the logic falls back to the body.

Additionally, the `exportCSV` logic would simply click the link and then remove it. This may not cause the focus to move, but it's a safe accessibility measure to return focus to the element it was originally on prior to clicking the download trigger.

A similar fix is required in `stripes-util` since we're not sure how many ui modules use THAT version of `ExportCSV`... That PR: https://github.com/folio-org/stripes-util/pull/91


UladzislauKutarkin (approved):

zburke (approved):
Thank you, thank you for the clear explanation of the problem, its solution, and the attendant a11y concerns and mitigations. It's so helpful to have that extra context in the comments.

PR #2405, opened 2024-12-20
STCOM-1392 Use `h3` instead of `h1` for menu.

JohnC-80:
We saw it coming. We did. Zak should've pressed me. We should have changed the default `labelTag` back then, in [STCOM-1191](https://folio-org.atlassian.net/browse/STCOM-1191). Despite so much documentation around the internet regarding the `<section>` tag and its contained heading supporting `h1-h6`, etc... there's a scarce amount of connection to the safety-first best-and-long-practiced pattern of ONE SINGLE H1 PER-PAGE! (regardless of visiblity status.) Regardless of what apparent outcome of articles like this ([`h2`  is the apparent ceiling for `sections`, period](https://www.webucator.com/article/html-heading-levels-and-sectioning-content/)) Having multiple `<H1>`'s show up to any validator and expecting validators/audit tools of any kind to read the minds of browsers/browser developers/spec writers/developers 
 everywhere... just set it to make sense if you view the mark-up. Since dropdowns are commonly used within panes (which conventionally hold `h2`'s as their header) defaulting to `h3` fits with the convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Wrap `<Card>`'s render output in `<StripesOverlayContext>` to facilitate ease with overlay components. Refs STCOM-1384.
 * Clear filter value after an action chosen from `MultiSelection` menu. Refs STCOM-1385.
 * ExportCSV - fix usage within `<Modal>`s by rendering the download link to the `div#OverlayContainer`. Refs STCOM-1387.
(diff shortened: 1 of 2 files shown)


zburke (approved):
LGTM. See #2330 for the original STCOM-1191 change that replaced `<div>` (whoops, a fuggle) with `<h1>` (a defuggle). This issue is now properly undefuggled 😆 🤝

PR #2409, opened 2024-12-30
STCOM-1195: `<AdvancedSearchRow>` - change `aria-label` for the input box to enter a search query and the Boolean operator dropdown.

Dmytro-Melnyshyn:
## Description
 `<AdvancedSearchRow>` - change `aria-label` for the input box to enter a search query and the Boolean operator dropdown.

## Related PRs
https://github.com/folio-org/stripes-testing/pull/4592

## Issues
[STCOM-1195](https://folio-org.atlassian.net/browse/STCOM-1195)

## Screenshots
image


JohnC-80 (approved):

JohnC-80:
@Dmytro-Melnyshyn Minor - but the labels in the tests will have to be updated as well.

mariia-aloshyna (approved):

BogdanDenis (approved):

Dmytro-Melnyshyn:
@JohnC-80, I updated `"@folio/stripes-testing"` from `^4.7.0` to `^4.9.0` to pull new changes into `yarn.lock` file. Do you have any objections?

JohnC-80:
@Dmytro-Melnyshyn If these changes are needed immediately, I would make the changes locally by extending the interactor in the way that you need in the test file for now... it's not certain that a stripes-testing release will be published when it comes time to publish this fix. Thank you for doing the  work within stripes-testing, though.

It looks like there's much more changing within the `yarn.lock` file aside from just the version of `stripes-testing`... I think we should hold off on changing that for now.

Dmytro-Melnyshyn:
@JohnC-80 This is a Sunflower release change, so the `stripes-testing` changes will be present. I've now manually modified `yarn.lock` to avoid other changes.

JohnC-80:
@Dmytro-Melnyshyn Ok sounds good :) Thank you!

JohnC-80 (approved):

zburke (commented):
@Dmytro-Melnyshyn, @JohnC-80, if this change requires `@folio/stripes-testing` `v4.9.x`, then you must change `package.json` to depend on `^4.9.0` instead of `^4.7.0`. It is not enough just to update `yarn.lock`. 

stripes-testing is just a dev-dep so the impact in this PR is insignificant, but it's important to understand how dependency resolution works. Say you are installing A, A depends on B, and B depends on C. The minimum dependencies specified in the `package.json` files of A and B and C will all be consulted, but the only relevant `yarn.lock` file is A's. IOW, if A has a dependency like `"C": "~4.7.0"` and B has a loose dependency on the same thing like `^4.7.0` but a `yarn.lock` entry like `4.9.1`, when you install A, you're going to get C `v4.7.0` because it satisfies all the `package.json` constraints. 

It is not necessarily surprising that updating a single dep like stripes-testing can have a wide impact on `yarn.lock` since dependency resolution is hierarchical.

PR #2411, opened 2025-01-02
STCOM-1391 Tooltip - check for hover on tooltip content before closing the tooltip.

JohnC-80:
According to [WCAG](https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus), content that's available/visible on hover should remain visible when the content itself is hovered... mouse-readers everywhere unite!

### Approach
Tooltip already uses timeouts to provide a brief gap between event listeners and state updates. This implementation added a timeout to hiding the tooltip where we only hide the tooltip if neither the trigger nor the tooltip element itself are hovered by the mouse (using `matches(':hover')`). The CSS of the tooltip class had to be affected to observe mouse events since they are all placed in the `#OverlayContainer` by default.


Dmytro-Melnyshyn (approved):

mariia-aloshyna (approved):

lib/Tooltip/Tooltip.js
```diff
@@ -99,10 +100,22 @@ export default class Tooltip extends Component {
     const disable = hideOnTouch && isTouch;
     clearTimeout(this.timeout);
 
-    if (bool !== open && !disable) {
-      this.setState({
-        open: bool,
-      });
+    // Accessibuilty - When hiding the tooltip, ensure that the mouse is not hovered over the tooltip
```
zburke, line 103: Pedantic: `s/Accessibuilty/Accessibilty/`

zburke (approved):
LGTM. There is a reason I only every type a11y 😆

PR #2423, opened 2025-02-10
STCOM-139 - Popper - hide popper overlay when target element is scrolled outside of the view.

JohnC-80:
### Approach

Popper already adds an attribute (via its default set of modifiers) - the hiding outcome just has to be connected via CSS. If the anchoring element control is visible within the view, then the popper/overlay will be visible as well. If not, it disappears.

### Result
hidePopup

### DIFF
--- a/lib/Popper/Popper.css
+++ b/lib/Popper/Popper.css
@@ -1,7 +1,7 @@
 .overlay {
   z-index: 9999;
 
-  /*  popper-appied element property when the 'anchor' element is hidden
+  /*  popper-applied element property when the 'anchor' element is hidden
       or scrolled out of the view. This rule effectively hides the overlay if the anchor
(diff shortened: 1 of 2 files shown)


Dmytro-Melnyshyn (approved):

lib/Popper/Popper.css
```diff
@@ -1,3 +1,13 @@
 .overlay {
   z-index: 9999;
+
+  /*  popper-appied element property when the 'anchor' element is hidden
```
zburke, line 4:
```
popper-applied
          ^
```
OMG, so pedantic

zburke (approved):
Typo in the comment but code LGTM. And THANK YOU for the comments explaining why this is necessary and how it works. If only all our code were so commented. A boy can dream.

PR #2425, opened 2025-02-12
STCOM-1408: Add marginTop0 class to MessageBanner component

mariia-aloshyna:
Introduce the marginTop0 class to the MessageBanner component to remove the default top margin.

## Refs
https://folio-org.atlassian.net/browse/STCOM-1408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Paneset - deduplicate panes via `id` prior to registration. Refs STCOM-1386.
 * Calendar - improved color contrast of edge month days, as per WCAG standards. Changed hover bg color of edge/month days. Increased weight of day numbers overall. Refs STCOM-1390.
 * *BREAKING* Update `react-intl` to `^7`. Refs STCOM-1406.
+* Add marginTop0 prop to the MessageBanner component. Refs STCOM-1408.
 
 ## [12.2.0](https://github.com/folio-org/stripes-components/tree/v12.2.0) (2024-10-11)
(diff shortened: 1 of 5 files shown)


zburke (commented):
I'm lost here -- I see the prop and the style, but not where anything actually happens with that prop. How does this work?

mariia-aloshyna:
> I'm lost here -- I see the prop and the style, but not where anything actually happens with that prop. How does this work?

Oh hi @zburke I am going to introduce it in this [PR](https://github.com/folio-org/ui-inventory/pull/2742) you can see from the screenshot that there is an extra space between message banner and top border

zburke:
Sorry my question wasn't clear, @mariia-aloshyna. What I meant was, I see this new prop on line 49, but it's never applied anywhere. The remaining props are collected into `rest` (line 50) which is then spread onto `<Element>` (line 110) but I don't see where `marginTop0` is applied to a DOM element. What's the magic ✨ ?

mariia-aloshyna:
> Sorry my question wasn't clear, @mariia-aloshyna. What I meant was, I see this new prop on line 49, but it's never applied anywhere. The remaining props are collected into `rest` (line 50) which is then spread onto `<Element>` (line 110) but I don't see where `marginTop0` is applied to a DOM element. What's the magic ✨ ?

No magic, I just missed a piece of code :woman_facepalming:. Updated the PR, thank you for catching it, now it works as expected.

zburke (approved):
No worries, PR review working exactly like it supposed to 🤝 

How did that test ever pass though? 🤔

PR #2429, opened 2025-02-20
STCOM-1336 remove react-overlays dependency

JohnC-80:
## [STCOM-1336](https://folio-org.atlassian.net/browse/STCOM-1336)

This implements [STCOM-1339](https://folio-org.atlassian.net/browse/STCOM-1339), as it was a sub-task of this story.

`react-overlays` hasn't been published in 2 years.

After the modal refactor, the remaining bits were the 
`useRootClose` hook and an instance of `<Portal>`

- <Portal> was a simple refactor - `React-DOM`'s base functionality for portals is all that's necessary there.

- ~~I opted to pull in a new dependency - `usehooks-ts` instead of re-implementing `useRootClose` locally. [usehooks-ts](https://usehooks-ts.com/introduction) is a minimal, no-dependency lib with active development and a nice dev experience curtesy of TS! It is pre-transpiled to commonjs, of course... `useClickOutside` nicely replaced `onRootClose`.~~
-  our own `useClickOutside` is implemented a bit differently, but served purpose just fine rather than taking on any dependency - and the logic is quite similar to that of the compound hook within `usehooks-ts`.

Additionally - updated shared workflows to 1.9 so that tests can run!


zburke (approved):
👏 👏 👏

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@
 * *BREAKING* remove deprecated props. Refs STCOM-1398.
 * Add marginTop0 prop to the MessageBanner component. Refs STCOM-1408.
 * Popper - hide overlay if popper anchor is scrolled out of the view. Refs STCOM-1386.
+* Switch `useRootClose` hook to `useOnClickOutside` from `usehooks-ts`. Refs STCOM-1339.
```
zburke, line 29: Omit/update this line?

lib/DropdownMenu/DropdownMenu.js
```diff
@@ -32,17 +33,15 @@ const DropdownMenu = ({ overrideStyle = {}, ...rest }) => {
...
     width
   } = props;
 
   const ref = useRef(null);
-
-  useRootClose(ref, onToggle, {
-    disabled: !open
-  });
+  const handleOutsideClick = useCallback((e, outside) => outside && onToggle(), [onToggle]);
+  useClickOutside(ref, open ? handleOutsideClick : noop);
```
zburke, line 44: Any value in having parity in the names here, e.g. handleClickOutside/useClickOutside (since `useClickOutside` is the one that already exists)?

zburke (approved):
True confession: I barely understand this, so I'm relieved that you do. Thank you for the clear explanation in the PR description. As a rule, I always favor more documentation: both "how this works" and "why it's useful". You can often figure out the former from reading the code, but the why is hard to divine without an explanation.

PR #2436, opened 2025-03-06
STCOM-1416: AuditLog: Add original version card

mariia-aloshyna:
## Purpose
When a card has action 'CREATE' and no field level changes info then it is required to show it, but with `Original version` label

## Approach
- Added `isOriginal` prop to `AuditLogCard`, get it from the `versions` list
- Updated `readme`
- Wrapped text content inside cards in `p` tag and apply styles

## Refs
https://folio-org.atlassian.net/browse/STCOM-1416

## Screenshot
Screenshot 2025-03-06 at 13 31 43

### DIFF
--- a/lib/AuditLog/AuditLogCard.js
+++ b/lib/AuditLog/AuditLogCard.js
@@ -16,6 +16,7 @@ const AuditLogCard = ({
   fieldChanges,
   fieldFormatter,
   fieldLabelsMap,
+  isOriginal,
   source,
   userName,
 }) => {
@@ -39,6 +40,14 @@ const AuditLogCard = ({
       />
     </>
   );
+  const originalVersionLabel = (
+    <b>
+      <FormattedMessage
+        id="stripes-components.versionHistory.card.originalVersion"
+        tagName="em"
+      />
+    </b>
+  );
 
   const onChangeButtonClick = () => {
     setIsModalOpen(true);
@@ -54,12 +63,15 @@ const AuditLogCard = ({
         roundedBorder
       >
         {fieldChangesSource}
-        <AuditLogChangedFieldsList
-          actionsMap={actionsMap}
-          fieldChanges={fieldChanges}
-          fieldLabelsMap={fieldLabelsMap}
-          onChangeButtonClick={onChangeButtonClick}
-        />
+        {isOriginal && <p>{originalVersionLabel}</p>}
+        <p>
+          <AuditLogChangedFieldsList
+            actionsMap={actionsMap}
+            fieldChanges={fieldChanges}
+            fieldLabelsMap={fieldLabelsMap}
+            onChangeButtonClick={onChangeButtonClick}
+          />
+        </p>
       </Card>
       <AuditLogModal
         contentData={fieldChanges}
@@ -82,8 +94,9 @@ AuditLogCard.propTypes = {
   fieldChanges: PropTypes.arrayOf(PropTypes.object).isRequired,
   fieldFormatter: PropTypes.object,
   fieldLabelsMap: PropTypes.object,
-  source: PropTypes.oneOf([PropTypes.string, PropTypes.node]).isRequired,
-  userName: PropTypes.oneOf([PropTypes.string, PropTypes.node]).isRequired,
+  isOriginal: PropTypes.bool,
+  source: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+  userName: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
 };
 
 export default AuditLogCard;
--- a/lib/AuditLog/AuditLogPane.js
+++ b/lib/AuditLog/AuditLogPane.js
@@ -59,6 +59,7 @@ const AuditLogPane = ({
 
   const versionCards = useMemo(() => {
     return versions.map(({
+      isOriginal,
       eventId,
       eventDate,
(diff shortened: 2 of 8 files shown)


lib/AuditLog/AuditLogCard.js
```diff
@@ -54,12 +63,15 @@ const AuditLogCard = ({
...
-        />
+        {isOriginal && <p>{originalVersionLabel}</p>}
+        <p>
+          <AuditLogChangedFieldsList
+            actionsMap={actionsMap}
+            fieldChanges={fieldChanges}
+            fieldLabelsMap={fieldLabelsMap}
+            onChangeButtonClick={onChangeButtonClick}
+          />
+        </p>
```
Dmytro-Melnyshyn, line 74: Let's hide the `AuditLogChangedFieldsList` component if `isOriginal` true, MARC records have `fieldChanges.length` even when the action is created.

lib/AuditLog/AuditLogPane.js
```diff
@@ -69,6 +70,7 @@ const AuditLogPane = ({
         <AuditLogCard
           key={eventId || i}
           date={eventDate}
+          isOriginal={isOriginal}
```
Dmytro-Melnyshyn, line 73: Let's put the `isOriginal` in the dependencies of `useMemo`.

lib/AuditLog/AuditLogCard.js
Dmytro-Melnyshyn:
```
        {isOriginal
          ? <p>{originalVersionLabel}</p>
          : (
            <p>
              <AuditLogChangedFieldsList
                actionsMap={actionsMap}
                fieldChanges={fieldChanges}
                fieldLabelsMap={fieldLabelsMap}
                onChangeButtonClick={onChangeButtonClick}
              />
            </p>
        )}

**Review by Dmytro-Melnyshyn — APPROVED** [review]

File: lib/AuditLog/AuditLogPane.js [hunk 3]
```diff
@@ -133,8 +135,11 @@ AuditLogPane.propTypes = {
     PropTypes.shape({
       eventDate: PropTypes.string,
       eventId: PropTypes.string,
+      eventTs: PropTypes.number,
```
zburke, line 138: I'm late to the party, but why oh why do we have `eventTs` and `eventDate`? That suggests a broken API IMO. Given `eventDate` is already present and looks to be an RFC-3339 string like most other date-time values in FOLIO, let's ignore `eventTs` and file a bug to have it removed from the API.

translations/stripes-components/en.json
```diff
@@ -53,6 +53,7 @@
   "versionHistory.pane.sub": "{count, plural, one {# version} other {# versions}}",
   "versionHistory.pane.loadingLabel": "Loading",
   "versionHistory.card.changed": "Changed",
+  "versionHistory.card.originalVersion": "Original version",
```
zburke, line 56: petty: should we consider changing the translation strings to be keyed off `auditLog` instead of `versionHistory` to match the name of the component that uses them?

lib/AuditLog/AuditLogCard.js
```diff
@@ -39,6 +40,14 @@ const AuditLogCard = ({
       />
     </>
   );
+  const originalVersionLabel = (
+    <b>
+      <FormattedMessage
+        id="stripes-components.versionHistory.card.originalVersion"
+        tagName="em"
+      />
+    </b>
```
zburke, line 49: If you want this wrapped in both `<b>` and `<em>` then do both the same way, e.g. 
```
    <b>
      <em>
        <FormattedMessage
          id="stripes-components.versionHistory.card.originalVersion"
        />
      </em>
    </b>
```
But, also why? And why mix b/em instead of b/i or strong/em? Zak can haz OCD? Yes! Zak have OCD!

lib/AuditLog/AuditLogCard.js
```diff
@@ -54,12 +63,19 @@ const AuditLogCard = ({
...
         {fieldChangesSource}
-        <AuditLogChangedFieldsList
-          actionsMap={actionsMap}
-          fieldChanges={fieldChanges}
-          fieldLabelsMap={fieldLabelsMap}
-          onChangeButtonClick={onChangeButtonClick}
-        />
+        <p>
+          {isOriginal
+           ? originalVersionLabel
```
zburke, line 68: Can we get a test that validates the presence of this label?

zburke (approved):
I have a couple of nits but nothing major, though I would like to see a test for how the display changes with `isOriginal` and don't understand why CI didn't ding us for that :/

lib/AuditLog/AuditLogPane.js
mariia-aloshyna: so talking from UI perspective - `eventDate` is used to render the header label of cards and modal window, but `eventTs` is used for pagination, when we need more curds to be displayed - we send additional param to the server, `eventTs` of the last rendered card.
Maybe @psmagin could add more details

mariia-aloshyna:
thank you for review @zburke all comments are fixed

lib/AuditLog/AuditLogPane.js
zburke: I don't expect the API to change at this point, but I don't see what prevents you from using `eventDate` for pagination too if it's always stored in UTC. 🤷

PR #2437, opened 2025-03-06
STCOM-1415: `AuditLog` - Show current version for the first card.

Dmytro-Melnyshyn:
## Purpose
Show the current version for the first card.

## Issues
[STCOM-1415](https://folio-org.atlassian.net/browse/STCOM-1415)

## Screenshots
image


lib/AuditLog/AuditLogCard.js
```diff
@@ -16,6 +16,8 @@ const AuditLogCard = ({
   fieldChanges,
   fieldFormatter,
   fieldLabelsMap,
+  isOriginal,
+  isFirst,
```
mariia-aloshyna, line 20: so is this actually the first or the last change? :thinking:

Dmytro-Melnyshyn: This flag indicates whether this is the first card or not. It is not about changes.

Dmytro-Melnyshyn: Maybe `isLatest` would be better, I will change it.

zburke: Or `isFirstCard` or `isLatestCard` since the confusion seems to be that this value refers to the card (among a list of cards) rather than the changes (among a list of changes).

lib/AuditLog/AuditLogChangedFieldsList.js
```diff
@@ -26,23 +27,33 @@ const AuditLogChangedFieldsList = ({
...
         onClick={onChangeButtonClick}
       >
-        <FormattedMessage
-          id="stripes-components.versionHistory.card.changed"
-          tagName="strong"
-        />
+        {isFirst
+          ? (
+            <b>
+              <em>
```
zburke, line 38: Petty, again: avoid mixing presentational and semantic markup

zburke (approved):
I have the same "use markup consistently" nits as on #2436 but no major complaints

mariia-aloshyna (approved):

PR #2438, opened 2025-03-06
STCOM-1417 - adjust styles for printing full pane content.

JohnC-80:
## [STCOM-1417](https://folio-org.atlassian.net/browse/STCOM-1417)

More to add here as cleanup continues.

For printing, we hide all panes except for the last one and set it to static positioning. The strange `min-height: 1px` is for `autosize` which requires a non-zero height.

https://github.com/folio-org/stripes-core/pull/1607 required

MCL in print dialog:
image


lib/MultiColumnList/MCLRenderer.css
```diff
@@ -422,4 +423,25 @@
...
 }
 
+ /* .printFullContents { */
+ @media print {
+  .mclContainer {
+    overflow: visible !important;
+  }
+
+  .mclRowContainer {
+      overflow: visible;
```
zburke, line 433: OMG. I'm hyperventilating. The whitespace is off here. OK, ok. Breathe. BREATHE! Lint, lint, where are you lint? Did we mess up lint in #1941 and never fix it? AFAIK we do not support v8 😬

zburke (approved):
I appreciate the comments WRT autosizer. Would any additional comments about the print styles be helpful or do you think they are self-explanatory?

CHANGELOG entry?

lib/MultiColumnList/MCLRenderer.css
JohnC-80: 1) Seems like this would fall into `style-lint`'s court since CSS... probably need future work to be sure that's working.
2) Are you okay?

PR #2443, opened 2025-03-18
STCOM-1421 Make `dayjs.contains` include start end end dates

BogdanDenis:
## Description

`moment.contains` included start and end dates by default, so 01/01/2025 would be considered included in 01/01/2025-12/31/2025 range
`dayjs` needs some additional configuration to replicate this behaviour
https://day.js.org/docs/en/plugin/is-between

## Issues
[STCOM-1421](https://folio-org.atlassian.net/browse/STCOM-1421)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * `AuditLog` - Show current version for the newest card. Refs STCOM-1415.
 * CSS Support for printing of results list content. Refs STCOM-1417.
 * `AuditLog` - add `modalFieldChanges` to display field change in card modal window, make modal large, add totalVersions. Refs STCOM-1419.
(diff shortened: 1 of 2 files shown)


zburke (approved):
LGTM. Thank you for the documentation reference! Consider adding a comment about the granularity and inclusiveness of this function to its comments. 

Other functions could use this commentary too if you are willing.

BogdanDenis:
Added a comment next to the call.

> Other functions could use this commentary too if you are willing.

What do you mean here by "other functions"?

zburke:
"other functions" meaning the other methods on this class would also benefit from more documentation. That's clearly outside the scope of work for this PR, but if you're updating documentation and feeling generous, of course I wouldn't object :)

zburke (approved):

OleksandrHladchenko1 (approved):

PR #2445, opened 2025-03-24
STCOM-1424: `AdvancedSearch` - replace `Cancel` button with `Reset all` in modal window.

Dmytro-Melnyshyn:
## Purpose
Add the ability to reset Advanced search input from modal

## Approach
Expose the `isPristine` property that compares the initial state of rows with the current one and use it to enable the `Reset all` button. Use the existing `resetRows` function to reset all states.

## Refs

## Screencasts

https://github.com/user-attachments/assets/04f0d79d-126c-4dba-9bfb-c6f487299ca9


zburke (approved):
LGTM. @JohnC-80 , any objections?

artem-blazhko (approved):

PR #2447, opened 2025-03-25
STCOM-1426 make AutoSuggest::defaultIncludeItem() null-safe

zburke:
`defaultIncludeItem(haystack, needle)` assumes `haystack` is shaped like `{ value }` and returns true if the `value` attribute is a string containing `needle`.

The previous incarnation did not validate whether `haystack` or `haystack.value` were null, leading to NPEs if `haystack` was `null` or if it did not contain an attribute named `value`.

CC: @bbarnetteebsco

Refs [STCOM-1426](https://folio-org.atlassian.net/browse/STCOM-1426)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * `AuditLog` - add `modalFieldChanges` to display field change in card modal window, make modal large, add totalVersions. Refs STCOM-1419.
 * Make `dayjs.contains` include start end end dates. Refs STCOM-1421.
 * Add loading indicator to `AuditLogPane` when data is initially loading. Refs STCOM-1422.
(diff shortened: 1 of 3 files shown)


NikitaSedyx:
I don't think it will help, it puts object instead of string in `valueKey` as a result form value becomes `{ field: { field: value } }` instead of just `{ field: value }`

with the steps described in the ticket it's not reproducible, our QA will add comment with exact steps to produce this issue

zburke:
@NikitaSedyx , I don't follow. The shape of the return values doesn't change here. Can you show me what you think the problem is?

zburke:
Replaced by #2449

PR #2449, opened 2025-03-27
STCOM-1426 Restore onSelect support for AutoSuggest component

NikitaSedyx:
purpose: https://folio-org.atlassian.net/browse/STCOM-1426

description: `useCombobox` doesn't support onSelect option (that was supported by previous AutoSuggest implementation), instead they have `onSelectedItemChange` https://github.com/downshift-js/downshift/blob/master/src/hooks/useCombobox/README.md#onselecteditemchange

The downshift upgrade for AutoSuggest switched from a component implementation to the `useCombobox` hook from `downshift` - The API change - as mentioned is covered in this change and kept the `onSelect` API intact. Additionally, I was unable to click on the option in the dropdown with the mouse - this is due to the way our `div#OverlayContainer` is setup when the dropdown is rendered it via React portal. Any element rendered directly there needs `pointer-events: all` to allow for mouse hovers/clicks to function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 13.0.3 IN PROGRESS
 
 * Display `Changed from - "-"` and `Changed to - "false"` values for boolean fields in AuditLogModal. Fixes STCOM-1427.
+* Restore onSelect support for AutoSuggest component. Refs STCOM-1426.
 
 ## 13.1.0 IN PROGRESS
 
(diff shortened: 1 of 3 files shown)


JohnC-80:
@NikitaSedyx Thank you for this! Along with your `stripes-acq-components` change, this should resolve... 

Added `pointer-events: all` to CSS to restore click functionality to the value list - this was broken since auto-portal behavior was added to `<AutoSuggest>` in `RepeatableFields`...

zburke (approved):
LGTM

@JohnC-80, can you explain a little more about how fixing the CSS and _removing_ the `onSelect` prop that was previously passed to `useCombobox` works to restore the click-to-select behavior? The PR description is not very detailed at present.

PR #2453, opened 2025-04-01
STCOM-1427: Display `false` for `Changed from` column also (follow-up)

OleksandrHladchenko1:
image


zburke (approved):
Please also remove `showFalsyValue ` from `AuditLogModal.js#L49` before you merge.

artem-blazhko (approved):

PR #2456, opened 2025-04-04
STCOM-1424: Use empty rows state for `isPristine` and for resetting rows state; add clear icon for inputs; disable search button (follow-up PR).

Dmytro-Melnyshyn:
## Purpose
1. The `Reset all` and `Search` buttons should be disabled with no values/selections in the advanced search modal.
2. A clear icon should be added for text fields.

## Issues

## Screencasts

https://github.com/user-attachments/assets/0de86b8f-8736-4bfb-b153-6cf08f86ace1

### DIFF
--- a/lib/AdvancedSearch/tests/AdvancedSearch-test.js
+++ b/lib/AdvancedSearch/tests/AdvancedSearch-test.js
@@ -338,9 +338,7 @@ describe('AdvancedSearch', () => {
 
     describe('resetting from the outside', () => {
       beforeEach(async () => {
-        await closeModalIcon.click();
         await ButtonInteractor('Reset').click();
-        await advancedSearchButton.click();
       });
 
       it('renders updated query values', async () => {
--- a/lib/AdvancedSearch/tests/AdvancedSearch-test.js
+++ b/lib/AdvancedSearch/tests/AdvancedSearch-test.js
@@ -24,6 +24,7 @@ import { mountWithContext } from '../../../tests/helpers';
 import AdvancedSearch from '../AdvancedSearch';
 import {
   BOOLEAN_OPERATORS,
+  DEFAULT_SEARCH_OPTION,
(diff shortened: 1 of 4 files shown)


BogdanDenis (approved):

lib/AdvancedSearch/tests/AdvancedSearch-test.js
```diff
@@ -338,9 +338,7 @@ describe('AdvancedSearch', () => {
 
     describe('resetting from the outside', () => {
       beforeEach(async () => {
-        await closeModalIcon.click();
         await ButtonInteractor('Reset').click();
-        await advancedSearchButton.click();
       });
 
       it('renders updated query values', async () => {
```
zburke, line 344: I don't understand how this test works now :/ previously, the `beforeEach` would close the AS modal, click reset, and open the AS modal. How can you click the reset button if the modal is visible? And you can you test components in the AS modal if you haven't clicked the AS button? I'm so confused.

Dmytro-Melnyshyn: @zburke this is not my test. I changed it but then reverted the changes because I changed the implementation. It somehow allows clicking. It looks like you are not looking at the latest changes.

Dmytro-Melnyshyn: @zburke there are two reset buttons, one in the main search (outside the modal) and the second reset button is inside the advanced search modal. This test describes the outer reset button click.

zburke (approved):

lib/AdvancedSearch/tests/AdvancedSearch-test.js
zburke: Thanks for taking the time to explain this to me, @Dmytro-Melnyshyn . I appreciate it.

PR #2457, opened 2025-04-04
STCOM-1403 - Empty MCL's container non-interactive tabIndex.

JohnC-80:
https://folio-org.atlassian.net/browse/STCOM-1403

Simply removes the static tabIndex from MCL's empty message wrapping element. Added an id to the element so that if it needs to be referenced/linked to some other element, it can be.


zburke (approved):
MCL change LGTM. 

Is the Timepicker storybook change included here deliberately?

JohnC-80:
@zburke nope, thanks! MCL-only... removed Timepicker storybook changes.

PR #2458, opened 2025-04-07
Move changelog items from 13.0.3 to 13.1.0

JohnC-80:
Multiple in-progress entries in the list - probably due to PR's spanning the version bump/merges of master... but confusing for incoming changes.

On master, they should all go to the latest version.. If any are for a previous, they would be cherry-picked/added to the necessary release branch.


zburke (approved):
Thanks for the clean up!

PR #2459, opened 2025-04-08
STCOM-1424: Prevent default behavior for search shortcut in `AdvancedSearchModal` (follow-up).

Dmytro-Melnyshyn:
## Purpose
Prevent the advanced search modal opening after hitting the enter key.

## Description
When the advanced search modal is open and there is any value, after hitting the Enter key the modal should close and perform the search, but currently, it automatically opens again due to the default action. This has been the behavior since the beginning. Adding `e.preventDefault` will prevent the modal from opening again ([comment](https://folio-org.atlassian.net/browse/STCOM-1424?focusedCommentId=259244)).

## Issues


Dmytro-Melnyshyn:
@zburke, when the advanced search modal is open and there is any value, after hitting the Enter key the modal should close and perform the search, but currently, it automatically opens again due to the default action. This has been the behavior since the beginning. Adding `e.preventDefault` will prevent the modal from opening again ([comment](https://folio-org.atlassian.net/browse/STCOM-1424?focusedCommentId=259244)).

zburke (approved):
Thanks for the explanation, @Dmytro-Melnyshyn. It would be great to include those details in a comment in the source so that folks who run across it understand why `e.preventDefault()` is there. Consider updating the description in the PR/in the comment when you squash-and-merge as well.

BogdanDenis (approved):

PR #2460, opened 2025-04-08
[STCOM-1438] Support for 23 minority languages

MikeTaylor:
These have been added to the list in `util/languages.js`, and their names to the translation file `translations/stripes-components/en.json`. They are required by the National Library of Sweden.


MikeTaylor:
When running the tests locally, they barf a vast amount of output including many errors (not just one), then apparently go into an endless loop — I got bored and interrupted it after three quarters of an hour. So my only hope of finding the test error is from the output of CI.

MikeTaylor:
False alarm, I managed to run the test locally. It needed a fresh build not in the context of a Yarn workspace, which is fair enough. The failing test is:
```
SUMMARY:
✔ 1565 tests completed
ℹ 8 tests skipped
✖ 1 test failed

FAILED TESTS:
  DateRangeWrapper
    Renders children
      changing the startDate picker
        changing the end date picker
          excluded days for end datepicker
            ✖ should be before the start date
              Chrome 135.0.0.0 (Mac OS 10.15.7)
            FilterNotMatchingError: calendar widget does not match filters:
            
            ╒═ Filter:   excludedDays
            ├─ Expected: ["1","2","3","4","5","6","7","8"]
            └─ Received: ["1","2","3","4","5","6","7","8","9"]
```
Which is hard to imagine being related the the present change.

zburke (approved):

PR #2462, opened 2025-04-09
STCOM-1429 Unable to close popover

JohnC-80:
## [STCOM-1429](https://folio-org.atlassian.net/browse/STCOM-1429)

When clicking an open Popover's target, the overlay will close and then re-open. This was caused by the `<RootCloseWrapper`'s handler triggering (closing the overlay) followed by the actual click handler (toggling the open status closed -> open).

Approach:
Check for the target at the `<RootCloseWrapper>`'s handler and allow Toggle to 'toggle' as it should.

Additional: Focus styles for the overlay since there wasn't any indication, and adding `toggle` functionality to the overlay's  escape key handler so that it actually closes the dialog (otherwise it was empty, aside from stopping propagation.)

Comments added for future us.


zburke (approved):
The code change LGTM; thank you!

I appreciate the comments but, well, they confuse me. Maybe describe the full flow of control in the comments on `handleClickOutside`, both when the target is within the trigger and when the target is outside it? 

Please add a CHANGELOG entry and update the PR title/commit title to reflect what this change does, e.g. sth like "Clicking control for open Popover now closes it".

zburke (approved):

PR #2471, opened 2025-05-09
STCOM-1448 Adjust default maxHeight of MultiSelection to accommodate 6 whole options.

JohnC-80:
## [STCOM-1448](https://folio-org.atlassian.net/browse/STCOM-1448)

A detail picked up by Kimie - MultiSelection would be *just that more useable if the options weren't just cut off by the edge of the menu.

## Before:

Screenshot 2025-05-07 at 11 13 43 AM


zburke (approved):
Sometimes it's the little things.

PR #2472, opened 2025-05-09
STCOM-1442 itemToView bug.

JohnC-80:
## WIP for [STCOM-1442](https://folio-org.atlassian.net/browse/STCOM-1442)

Cleanup needed... already improved behavior in the storybook item-to-view story.

## Approach thus far:
1) previous item capturing was done from `event.target` rather than 'item' aka 'row'. Sometimes `target` might be a `<Checkbox>` that's aligned middle, so measurements are different from expected.

2) previous item capturing happened on focus. While this is fine when a row is selected, it's awkward because it won't happen again while the row is focus - and interactions like checkbox checks and stuff can happen again in the focus state, so the item capture wouldn't be updated correctly.

### DIFF
--- a/lib/MultiColumnList/MCLRenderer.js
+++ b/lib/MultiColumnList/MCLRenderer.js
@@ -662,7 +662,7 @@ class MCLRenderer extends React.Component {
       }
 
       if (!virtualize && columns.length === Object.keys(stateColumnWidths).length) {
-        this.scrollToItemToView();
+        // this.scrollToItemToView();
       }
 
       // update header row and check for vertical scrollbar...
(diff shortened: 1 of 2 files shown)


lib/MultiColumnList/MCLRenderer.js
```diff
@@ -1881,6 +1896,24 @@ class MCLRenderer extends React.Component {
...
+      this.itemToView = {
+        selector: `[aria-rowindex="${rowIndex}"]`,
+        localClientTop: rowRect.top - mclRect.top,
+      };
+      this.itemToViewIsStale = false;
+      if (this.props.onMarkPosition) {
+        this.props.onMarkPosition(this.itemToView);
+      }
+    }
+  }
```
zburke: It would be helpful to have a header comment explaining (a) what this function does and (b) why it needs to do that work.

zburke (approved):
LGTM. Comments on new and updated functions would be a welcome addition. This stuff is pretty arcane and it's not immediately obvious what all the calculations do or why they are necessary.

PR #2485, opened 2025-07-18
STCOM-1451 correctly handle empty codes in formattedLanguageName

zburke:
1. Return "undetermined" from `formattedLanguageName()` given an empty code. Previously, empty codes matched the first value in the languages table that lacked a two-character code. Whoops!
2. Correctly sort entries according to current locale in `languageOptions()`. The previous implementation used `lodash.sortBy`, an implementation that does not correctly handle localized data.

Refs [STCOM-1451](https://folio-org.atlassian.net/browse/STCOM-1451)


JohnC-80 (approved):
Code looks good! Thank you for comments and tests 😍 😻

PR #2488, opened 2025-07-23
STCOM-1458 ButtonGroup spacing bug.

JohnC-80:
The tale of two selectors and their associated margin...

`.button--narrff + .button--narrff    0.25rem;`

`.buttonGroup--plz > :not(:first-child)    -1px;`

`ButtonGroups` selector is a bit more complex, more `pseudo-y` than `Button` - which means potentially fraught as browsers make adjustments for handling...

This PR adds `.buttonGroup--plz [class^=button]+[class^=button]` - No more pseudos, and higher specificity since the parent is specified. `ButtonGroup` takes ownership of its children as it should.

Also,
`!important` is the thing that Meatloaf won't do for love. Get outta here with that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,1287 +2,1387 @@
 
 ## 13.1.0 IN PROGRESS
 
-* Introduce `<AuditLog>` component. Refs STCOM-1412.
-* `AuditLog` - Add original version card. Refs STCOM-1416.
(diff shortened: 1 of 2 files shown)


JohnC-80:
!!important probably would've been another way to solve this - but not a good one. I don't think the pseudos are at fault and yeah, we'd want to keep them.. people like to put dropdowns in these things too.

PR #2495, opened 2025-09-11
STCOM-1460 Add SessionConfirmationModal component

JohnC-80:
A confirmation modal with a built-in suppression checkbox.
Refs [STCOM-1460](https://folio-org.atlassian.net/browse/STCOM-1460)

### Approach
Session value is updated in the `onConfirm` handler of the `ConfirmationModal` in case the user has second thoughts and wants to uncheck..

If the suppression session value is `true` the component just returns `null`.

Adds the `rehooks-ts` dependency for a `sessionStorage` hook.

Additional: cleaned up some of the old style code in ConfirmationModal... `props.foo` everywhere!

- [x] component
- [x] Storybook example
- [x] [Documentation](https://github.com/folio-org/stripes-components/blob/STCOM-1460/lib/ConfirmationModal/readme.md#sessionconfirmationmodal)
- [x] Tests

### Usage 

```
<SessionConfirmationModal
  open={this.state.confirming}
  heading="Please confirm!"
  message="Description of the thing that needs confirming"
  onConfirm={keepGoing}
  onCancel={stopGoing}
  storageKey = {'my-specific-storage-key'}
/>
```

image


translations/stripes-components/en.json
```diff
@@ -28,6 +28,7 @@
   "previous": "Previous",
   "next": "Next",
   "selectDay": "Choose {dayOfWeek}, {date}",
+  "ConfirmationModal.suppressMessage": "Do not show this message again",
```
zburke, line 31: Are we happy with "Do not show this message again" or do we want sth more verbose but also more accurate like "Automatically confirm future actions of this time for the duration of this session?" or "Do not require confirmation of this activity again for this session"?

lib/ConfirmationModal/readme.md
```diff
@@ -38,14 +38,41 @@ handleSubmit() {
...
+/>
+```
+
+### Additional props for SessionConfirmationModal
+
+Name | type | description | default | required
+--- | --- | --- | --- | ---
+`sessionKey` | string | Key for value in `SessionStorage` | randomized in ref |
+`checkboxLabel` | node | Label for suppression checkbox. | "Do not display this message again." |
+`confirmIfSuppressed` | bool | if `true` will execute the `onConfirm` handler if the modal is suppressed |
```
zburke, line 78: I don't understand `confirmIfSuppressed`. What would be the point of the checkbox if this were false?

JohnC-80: Good question @zburke . The hiding of the modal is one part of this - the `confirm()` logic is another. The prop is simply a 'logic does/doesn't need to run' setting - if the confirm handler just hides the window, it wouldn't have to be run - but that's probably a limited case.

If we're sure that we always want `confirm()` to be called if the modal is suppressed, we could remove this. @ryandberger would that be expected behavior? Hiding the modal =  calling `confirm()` for rest of session...

JohnC-80: Alright. Prop/accompanying logic removed!

PR #2496, opened 2025-09-23
STCOM-814 Leverage sourcemaps in ErrorBoundary

zburke:
## Summary

If a sourcemap can be loaded and parsed, show the mapped stacktrace instead of the original (pretty much useless) stacktrace. 

## Details

* March 2021: We merged #1512 even though the reporting wasn't great 😑 
* ... ~200 "Something went wrong" tickets are filed 😱 
* October 2024: Ugh, STCOM-1355, we should purge the useless reporting 😢 
* ... ~70 more "Something went wrong" tickets are filed 🙀 
* September 2025, [WOLFCon's "How do we figure out what the problem is?"](https://wolfcon2025.sched.com/event/25ldl/how-do-we-figure-out-what-the-problem-is) reminds me how annoying it is that we couldn't do better, which led to more searching, which led to https://github.com/novocaine/sourcemapped-stacktrace, which has been around since 2018 so I don't know why we didn't find it earlier, but we didn't, and then we did, and here we are. So even though STCOM-814 was closed as "Won't do" years ago, sometimes ya just gotta [do it anyway](https://www.youtube.com/watch?v=mEyrfFwf3rI) 🔥 .

### Before

Screenshot 2025-09-23 at 9 52 28 AM

### After

Screenshot 2025-09-23 at 9 51 40 AM

Refs [STCOM-814](https://folio-org.atlassian.net/browse/STCOM-814), [STCOM-1355](https://folio-org.atlassian.net/browse/STCOM-1355)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * `<AuditLog>` - change styling of "Current version" and "Changed" labels. Refs STCOM-1450.
 * Correctly handle empty language-codes in `formattedLanguageName()`. Refs STCOM-1451.
 * Remove spacing from `<ButtonGroup>`. Refs STCOM-1458.
+* Use mapped stacktraces in `<ErrorBoundary>` when sourcemaps are present. Refs STCOM-814.
 
 ## [13.0.0](https://github.com/folio-org/stripes-components/tree/v13.0.0) (2025-02-24)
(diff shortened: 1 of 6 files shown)


BogdanDenis (approved):
Thank you, thank you, thank you for this!!! I'm so happy we won't be seeing these useless bundle stacktraces in bug reports anymore

PR #2497, opened 2025-09-29
Revert "STCOM-1460 Add SessionConfirmationModal component"

JohnC-80:
Reverts folio-org/stripes-components#2495


zburke:
Replaced by #2499 and #2498

PR #2530, opened 2026-03-27
prop-type of searchableIndexes[] changed to to node

MikeTaylor:
When this is provided as `<FormattedMessage>`, a prop-types warning is emitted on the console even though this works just fine. We should get rid of the false-positive complaint.


zburke (approved):
`string` is the wrong thing for the `<FormattedMessage>` reason you identified, but I think the right thing is `node`. Moving to `any` just trades the prop-types warning for a lint warning since lint does not like `any`.

That said, lint is totally hosed in this repo, and I'm fine accepting this as a bug-fix even if it's questionable as an improvement. So, ultimately, your call.

JohnC-80 (approved):

PR #2532, opened 2026-04-02
STCOM-1489 Adding `dompurify` dependency, wrapping Editor with it.

JohnC-80:
## [STCOM-1489](https://folio-org.atlassian.net/browse/STCOM-1489)

Short of actually picking/implementing a new WYSIWYG editor, our security fallback would be to sanitize whatever content goes in and comes out of it and not EVEN to trust a thing that holds the sole purpose of editing some mark-up under the guise of a mini word-processor in a browser.

All editor functionality is retained, with links behaving as they should with the default `santizeConfig`. In case other `formats` are added, sanitization can be adjusted - although we see very little actual use of this component.

Tests added.

### DIFF
--- a/lib/Editor/Editor.js
+++ b/lib/Editor/Editor.js
@@ -28,13 +28,15 @@ const defaultSanitizeConfig = {
 };
 
 const sanitize = (value, config = defaultSanitizeConfig) => {
-  // Preserve original HTML when DOMPurify reports no actual removals.
   let resultValue = DOMPurify.sanitize(value, config);
 
   if (value !== resultValue) {
-    const removed = DOMPurify.removed.map((item) => item.attribute?.name || item.element?.outerHTML);
+    const removed = DOMPurify.removed.map((item) => item.attribute?.name || item.element?.outerHTML) || [];
 
-    if (removed && removed.length === 0) {
+    // Preserve original HTML when DOMPurify reports no actual removals.
+    // This addresses cases where DOMPurify may have reversed the order of attributes.
+    // If the editor picks up a change in its value, it will shift the cursor to the beginning of the text.
+    if (removed.length === 0) {
       resultValue = value;
     }
   }
--- a/lib/Editor/Editor.js
+++ b/lib/Editor/Editor.js
@@ -31,12 +31,10 @@ const sanitize = (value, config = baseSanitizeConfig) => {
   let resultValue = DOMPurify.sanitize(value, config);
 
   if (value !== resultValue) {
-    const removed = DOMPurify.removed.filter((item) => item.attribute?.name || item.element?.outerHTML);
-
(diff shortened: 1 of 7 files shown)


lib/Editor/Editor.js
```diff
@@ -20,6 +21,27 @@ const defaultModulesConfig = {
...
+const sanitize = (value, config = defaultSanitizeConfig) => {
+  // Preserve original HTML when DOMPurify reports no actual removals.
+  let resultValue = DOMPurify.sanitize(value, config);
+
+  if (value !== resultValue) {
+    const removed = DOMPurify.removed.map((item) => item.attribute?.name || item.element?.outerHTML);
+
+    if (removed && removed.length === 0) {
+      resultValue = value;
+    }
```
zburke: Did you mean `DOMPurify.remove.filter()`? With `.map()`, the output array will always match the length of the input array, so you could just directly evaluate `DOMPurify.removed?.length === 0` if you are only interested in "Did DOMPurify remove anything?"

lib/Editor/Editor.js
```diff
@@ -99,6 +123,29 @@ class Editor extends Component {
     );
   }
 
+  getSanitizeConfig() {
+    const { sanitizeConfig } = this.props;
+    const customAttrs = sanitizeConfig && Array.isArray(sanitizeConfig.ADD_ATTR)
```
zburke:
```
Array.isArray(sanitizeConfig?.ADD_ATTR)
```

package.json
```diff
@@ -96,6 +96,7 @@
     "classnames": "^2.2.5",
     "currency-codes": "2.1.0",
     "dayjs": "^1.11.10",
+    "dompurify": "^3.1.3",
```
zburke: This is resolved to `3.3.3` in the `yarn.lock` so let's specify that version here.

lib/Editor/Editor.js
```diff
@@ -62,6 +85,7 @@ class Editor extends Component {
     readOnly: false,
     disableEditorTab: true,
     modules: {},
+    sanitizeConfig: {},
```
zburke, line 88: I see how this works now, but it confused me at first: "OK, we set `defaultSanitizeConfig` to XYZ, and then we set the default `sanitizeConfig` to an empty object. Wait, what?? 😕"

What if we rename `defaultSanitizeConfig` to `baseSanitizeConfig`? IDK, maybe nobody else will be confused by this.

zburke (approved):
I got tripped up by multiple default-y things on the first read, but functionally this looks great. Thank you for updating the readme and for the thorough new tests.

Please address Sonar's optional chaining gripe before you merge 😒

lib/Editor/Editor.js
JohnC-80: Reduced to just looking at _any removals rather than filtering.

PR #2533, opened 2026-04-03
STCOM-1481 - Add ability to keep disabled buttons focusable

JohnC-80:
## [STCOM-1481](https://folio-org.atlassian.net/browse/STCOM-1481)

From readme of this PR:
> Disabling a button takes it out of the accessibility tree, making the element non-focusable. For assistive technology, this is as if the button were removed entirely. 95% of the time, this happens in a situation
when another reasonable focus-target is within proximity where focus can effectively be moved to. With the
other 5% of cases, there's no focusable element in reasonable range. This is where using `aria-disabled=true` is
useful, so that focus *can remain on the button itself without moving it far away from a reasonable focus order.

>Since the button is not formally `disabled`, it can still be triggered with a keypress. That said, click handlers will need to include
a defensive conditional to check the `aria-disabled` attribute before executing the rest of the function, like the following example:

```
const handleClick = (event) => {
  // Check if aria-disabled is explicitly set to "true"
  if (event.currentTarget.getAttribute('aria-disabled') === 'true') {
    event.preventDefault(); // Stop default browser actions
    return;                 // Exit early to prevent custom logic execution
  }

  // Your custom click logic here...
  console.log("Action performed!");
};
```

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 * Provide `isLoading` prop for `<MetaSection>` components to reflect loading status. Refs STCOM-1467.
 * Expose the `dedupe` property for `<Callout>` to prevent duplicate callouts with the same message and type from being displayed. Refs STCOM-1469.
 * `<MultiSelection>` - add a new `filterProps` prop to control option filtering. Refs STCOM-1473.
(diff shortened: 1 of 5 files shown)


zburke (approved):
I really appreciate the updated documentation, both in the readme and here in the PR description ❤️

PR #2535, opened 2026-04-09
STCOM-1501 - Add id attribute to Layer div.

JohnC-80:
## [UIDATIMP-1750](https://folio-org.atlassian.net/browse/UIDATIMP-1750)

A Tale of two Layers - 
So prior to this, Layer had no id... When the panesetResizeContainer would try and see if a removed node matched the blocker, two separate layers would look the same, since their id, classname, style was "", "LayerRoot---* "" - no discernable difference, so it would think that the Layer for the detail was being removed as the Edit Layer was being added.


zburke (approved):
Wow, brilliant debugging, and such a simple fix in the end! Ain't it always that way? Nice work, detectives! Please add a note to the CHANGELOG

mariia-aloshyna (approved):

PR #2537, opened 2026-04-16
STCOM-1503 - upgrade storybook - v7 -> v10

JohnC-80:
Following experiments will be to build up this content with information from the old ux.folio.org website.

This pull request updates the Storybook configuration and related files to modernize imports, streamline the story and addon setup, and improve documentation consistency. The changes primarily focus on updating import paths for Storybook packages, simplifying configuration files, and renaming documentation files for clarity.

Noteable Changes:
- changing `.stories.mdx` to `.mdx`.
- removing README imports - those will be handled in a future PR, when we FINALLY incorporate all of that great content ++ into our own storybook (including content from the old ux site)
- import syntax conversion for ESM... `require()` -> `import`


lib/PaneFooter/stories/BasicUsage.js
```diff
@@ -3,7 +3,7 @@
  */
 
 import React from 'react';
-import { action } from '@storybook/addon-actions';
+import { action } from 'storybook/actions';
```
zburke, line 6: I'm skeptical of this an other changes where the `@` is being stripped from the package-name because (a) it's inconsistent (e.g. `storybook/actions` but `@storybook/addon-docs/blocks`) and more importantly (b) it doesn't line up with the package-names in `package.json`. Are you confident this is correct?

JohnC-80: I thought so at first also. For more recent versions, these addons were added to the main `storybook` package. The serve/build functionality doesn't fail or seem broken. I will triple-check... 
But it is legit: 
https://github.com/storybookjs/storybook/blob/next/code/core/package.json#L49

JohnC-80: Some of the addons changed their own shape as well...  implementation of imports is correct as per https://storybook.js.org/docs/writing-docs/doc-blocks#within-mdx

zburke: 👌 🤢

zburke (approved):
Thank you for this!

I'm excited to configure [Storybook Controls](https://storybook.js.org/docs/essentials/controls/?renderer=react&ref=guide) in a future PR. This has been a major gripe of mine for a long time: our stories do a great job of showing us how components will render in a browser, but don't show how to actually use them in code 😖.

PR #2539, opened 2026-05-04
STCOM-1258 FINALLY -- Readme's HAVE come BACK to storybook!

JohnC-80:
## [STCOM-1258](https://folio-org.atlassian.net/browse/STCOM-1258)

Readme's have always had this layer of awkwardness with our storybook. Prior to our v7 upgrade, they existed off in a "readme" tab, apart from where the real action is and disconnected from the main storybook navigation. Story book doesn't *just support/incorporate markdown files by default, they have to be specially marked as stories, globbed into the build, etc etc. MDX is very nice - allowing both static readme content AND interactive examples of react components. We'll see more usage of MDX in future PR's as we make our documentation even more intuitive.

This PR:
- Sets navigation pathing in each of our existing stories to organize them in the storybook navigation.
- adds a script to take existing `Readme.md`'s and generate mdx files for the storybook build - `generate-readme-mdx.js`
- adds another script for running storybook that will generate, run storybook, and **Clean up the generated files** when storybook exits.
- adds `/lib/*/**.mdx` to `gitignore` so that generated files are not included in commits/PR's. 

We add some additional `devDependencies` to aid in rendering the markdown of the readme files to the `.mdx` files:
- react-markdown - Renders README markdown text to React elements inside generated MDX docs.
- remark-gfm - Enables GitHub-flavored markdown features (tables, strikethrough, task lists).
- rehype-raw - Allows inline HTML found in README files to be rendered in docs output.

image


zburke (approved):
Thank you for this! It has long been a pet peeve of mine that our storybook showed how components render, but not how to use them! What a relief to finally have both the display and the instructions in the same place!

Do we still need the `.circleci` directory? I gotta believe it is just an artifact of a long-ago FrontSide project when we were first getting started with CI. Nice of your agent to update it though 😆

zburke:
Why is `Calendar` stuck down at the bottom of the list? Everything else is nicely alphabetized 🤔

JohnC-80:
@zburke Properly alphabetized, and applied explicit control of the categories.

PR #2540, opened 2026-05-07
STCOM-1504: Pane support for disabled action menu and tooltip

slaemmer:
[STCOM-1504](https://folio-org.atlassian.net/browse/STCOM-1504)

### Summary

Adds `actionMenuDisabled` and `actionMenuDisabledTooltip` props to `Pane` and forwards them through `PaneHeader`. Lets consumers render the action-menu trigger in a disabled state with an optional tooltip explaining why, instead of replacing the real action menu with a per-consumer fake disabled `<Button>` workaround.

Originating discussion: [folio-org/ui-users#3035, comment 3190965080](https://github.com/folio-org/ui-users/pull/3035#discussion_r3190965080) (UIU-3388 review).

### Behaviour

- `actionMenuDisabled: bool`: disables the action-menu trigger so it cannot be opened.
- `actionMenuDisabledTooltip: string | <FormattedMessage>`: tooltip content shown on hover/focus while the menu is disabled. Ignored when `actionMenuDisabled` is `false`.
- The disabled trigger is wrapped in a `<Tooltip>` with `aria-describedby` linking to the tooltip text, so the explanation is reachable to screen-reader users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## IN PROGRESS
 
 * Supply Personal Data Disclosure form. Refs STCOM-1449.
+* `Pane`/`PaneHeader` - Add `actionMenuDisabled` and `actionMenuDisabledTooltip` props for rendering a disabled action menu with an optional tooltip. Refs STCOM-1504.
 
 ## [13.1.0](https://github.com/folio-org/stripes-components/tree/v13.1.0) (2026-04-14)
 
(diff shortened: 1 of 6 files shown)


slaemmer:
Won't be done as per changed requirements [the Actions menu and Tags button should not change when the Version history pane is opened](https://folio-org.atlassian.net/browse/UIU-3566).

PR #2541, opened 2026-05-08
STCOM-1472 update timezone data

zburke:
Pull timezone data from v2026b of www.iana.org's public domain zonenow.tab database. This is substantially shorter (~100 entries) than using the complete database containing all timezones (~300) that have been valid since 1970. We are not interested in historical data reaching back that far, so this seems a fair compromise.

Refs [STCOM-1472](https://folio-org.atlassian.net/browse/STCOM-1472)


util/timezones.js
```diff
@@ -1,468 +1,103 @@
...
+  ['Pacific/Bougainville', 'Russia; Kosrae; Bougainville; Solomons'],
+  ['Pacific/Norfolk', 'Norfolk Island'],
+  ['Asia/Kamchatka', 'Russia; Tuvalu; Fiji; etc.'],
+  ['Pacific/Auckland', 'New Zealand (NZST/NZDT)'],
+  ['Pacific/Chatham', 'Chatham Islands'],
+  ['Pacific/Tongatapu', 'Kanton; Tokelau; Samoa (western); Tonga'],
+  ['Pacific/Kiritimati', 'Kiritimati'],
+  ['UTC', 'UTC'],
+]
+  .map(entry => ({ name: entry[0], value: entry[1] }))
```
JohnC-80: @zburke FYI the logic within `tenant settings` - https://github.com/folio-org/ui-tenant-settings/blob/master/src/settings/Locale/LocaleForm.js#L33
`value` field is rendered in the dropdown (which is `name` in this changed code) and also what [DayJS](https://day.js.org/docs/en/timezone/timezone) and browser [Intl API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) like to use for values - although semantically, it's a name AND a value - we're best to incur less friction XD ... so `value=[0] and name=[1]` is what we need.

JohnC-80 (changes requested):

util/timezones.js
zburke: Ew the naming there is terrible, but I guess we can continue to use it. How unfortunate. Thanks for pointing this out. I never in a million years would have thought to use `value` to mean `key` given the notion of "key-value pairs".

JohnC-80 (approved):

PR #2544, opened 2026-05-14
STCOM-1507 - Use sha for storybook action

JohnC-80:
Since we control this action in stripes-components, it's only fair we maintain/update its actions.


zburke (approved):
LGTM as-is 🙌 but the PR description hints at something that remains apparent in `build-storybook.yml` itself where `NODEJS_VERSION` is set to v20, which was EOLed a few months ago. 

Yes, let's merge this now. Let's also figure out the process/cadence/responsibility for ongoing maintenance.

PR #2547, opened 2026-05-22
STCOM-1491 Select placeholder/disabled should have proper contrast

JohnC-80:
## [STCOM-1491](https://folio-org.atlassian.net/browse/STCOM-1491)
This PR aligns the coloration of placeholder text of `<Select>` with that of `<TextField>`. On top of the unification, this also improves the contrast of placeholders in disabled `<Select>`'s to a minimum contrast of 4.5:1

In the screenshot, the `<Select>`'s 'select a snack' is a placeholder.

Specific CSS changes:
User agent styles (the browser) wants to set `disabled` selects to a .75 opacity by default. The change in this PR resolves that, setting Opacity to one. Also, since 'disabled' selects are darker/grayer, the placeholder is made darker to retain contrast.
 

Screenshot 2026-05-26 095646

EDITS:
- Rely only on CSS for styling vs JS value checking. The current implementation is a little awkward - setting color for the entire control means all of the options within the dropdown get the color (as the C(SS)ascade dictates.) This change creates different styling for the non-selected , non-placeholder options so they all get grayed out, and only sets the color when the specific placeholder `<option>` is selected. 🎉 
- turns out that react/jsx gripes about the `selected` attribute being used on `<options>` and I don't want to set a universal `defaultValue` at this component level, so back to requiring something extra for placeholders to work. Documented those needs.


lib/Select/Select.js
```diff
@@ -66,7 +66,7 @@ class Select extends Component {
       css.selectControl,
       // placeholder styling
       {
-        [`${css.placeholder}`]: this.props.placeholder && !this.props.value
+        [`${css.placeholder}`]: this.props.placeholder && this.props.value === '',
```
zburke: Why are we specifically comparing to empty string here?

lib/Select/Select.js
```diff
@@ -112,7 +112,7 @@ class Select extends Component {
 
     if (this.props.placeholder) {
       options.push(
-        <option value="" key="x" disabled>{placeholder}</option>
+        <option value="" key="x" selected disabled>{placeholder}</option>
```
zburke: Why is this changing?

lib/sharedStyles/sharedInputStylesHelper.js
```diff
@@ -34,7 +34,7 @@ export default (props) => {
     { [`${formStyles.isValid}`]: isValid },
     { [`${formStyles.hasFeedback}`]: hasFeedback },
     { [`${formStyles.readOnly}`]: props.readOnly },
-    { [`${formStyles.disabled}`]: props.disabled },
+    { [`${formStyles.isDisabled}`]: props.disabled },
```
zburke, line 37: Is this the "alignment" mentioned in the description? i.e. the bug was that `/lib/sharedStyles/form.css` is written in terms of `isDisabled` but we passed in `disabled` ... and then dropped it on the floor?

zburke (approved):

lib/Select/Select.js
JohnC-80: This ensures that a placeholder is selected/visible if no `value` prop is passed in. It probably wasn't implemented this way previously because marking something as 'selected' seems a little suspicious since that should be up to the user - however, this does not interfere with selection. Otherwise, apps would have to pass `''` to get the placeholder to show up. Since it's a disabled option, `<Select>` just doesn't want to work like that.

lib/sharedStyles/sharedInputStylesHelper.js
JohnC-80: There is no `formStyles.disabled` class - so it was rendering `undefined` in the mark-up :/ There's no `.undefined` class either :P

lib/Select/Select.js
JohnC-80: Changed this implementation. Was able to rely solely on CSS for this. Thanks for mentioning it - it didn't feel great to be so specific about it.

lib/Select/Select.js
```diff
@@ -146,7 +142,7 @@ class Select extends Component {
       >
         {options}
         {
-          React.Children.map(children, (child) => {
+          React.Children.map(children, (child, i) => {
```
zburke: Looks like `i` is unused in this loop. Maybe this is left over from an earlier refactor?

zburke (approved):
The `data-placeholder` attribute is a nice, very deliberate way to handle this.

lib/Select/Select.js
JohnC-80: Thanks - that'd be me flailing and trying to find where a necessary `key` prop was missing... removed.

PR #2550, opened 2026-06-03
[STCOM-1510] Keypresses caught by `<HasCommand>` are consumed

MikeTaylor:
See discussion in the Jira issue
https://folio-org.atlassian.net/browse/STCOM-1510

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Supply Personal Data Disclosure form. Refs STCOM-1449.
 * Re-incorporate component Readme's into storybook. Refs STCOM-1258.
 * Switch from `cloneDeep` in `<Callout>` component's state updates- a fix for potential stack overflows due to cloned react component trees 😬 . Refs STCOM-1508.
+* Keypresses caught by `<HasCommand>` are consumed. Refs STCOM-1510.
 
 ## [13.1.0](https://github.com/folio-org/stripes-components/tree/v13.1.0) (2026-04-14)
 
--- a/lib/Commander/HasCommand.js
+++ b/lib/Commander/HasCommand.js
@@ -45,7 +45,10 @@ class HasCommand extends React.Component {
     const handlers = {};
     commands.forEach(c => {
       if (c.handler) {
-        handlers[c.name] = c.handler;
+        handlers[c.name] = (e) => {
+          e.preventDefault();
+          return c.handler(e);
+        }
       }
     });
     return handlers;


zburke (commented):
This problem is slippery. While the problem-description in the ticket leading to this solution feels obviously correct at first glance, there is a nod to this known behavior in the README:

> ## Browser shortcuts.
> Care must be taken if a known browser shortcut is supplied to `<CommandList>`, e.g. `mod+s` is used by browsers to save the actual html page locally but is a comfortable convention for users when saving a form. In most cases, the shortcut should be changed to something else. To avoid the browser's behavior, you must call `e.preventDefault()` within the shortcut's handler.
> ```
> handleKeyboardSave(e) {
>     e.preventDefault();
>     executeSave();
> }
> ```

Omitting `e.preventDefault()` is what allows us to be alerted that we have chosen a shortcut that is already bound (as we did in [ERM-4070](https://folio-org.atlassian.net/browse/ERM-4070)). In other words, the existing documentation reads, 

> It is OK to override default behavior, but you should do so knowingly, and probably rarely. 

This PR flips that logic, making overrides more difficult to detect** and implementing them universally. Are we confident in this change? Are we confident that a few English-speaking/English-keyboard-using developers can make this decision, or is this something we should raise to a wider community of developers (e.g. via Slack#folio-development) or users (maybe there's a SIG to ask?). 

** Likely, instead of QAs getting FOLIO behavior + default behavior and telling devs, "I expected A but I got A + B", this behavior is likely to be "shifted right" from QA into production where end-users may encounter it as, "I expected A, but I got B instead".

[Additional context from the Slack discussion](https://open-libr-foundation.slack.com/archives/C210UCHQ9/p1779893826878769), for the historical record: 
> maybe it’s time to talk about how difficult, nay :horse: , impossible it is to find universal cross-locale keyboard shortcuts that don’t resemble emacs-on-crack, and so we should pursue managing keyboard shortcuts the same way we do translations, i.e. every locale gets their own.
> 
> * Good: we can have sane shortcuts like ctrl-e instead of ctrl-alt-meta-backslash-option-esc-esc-up-down-up-down-left-right-b-a-e
> * Bad: it’s something else to maintain, and documentation about keyboard-shortcuts is no longer locale-neutral

MikeTaylor:
Thanks, @zburke, that's exactly the kind of analysis I was looking for.

So I _think_ your position can be summarized as: you agree that the present behaviour is a bug, but it helps us to diagnose other bugs (choice of shortcuts that conflict with browser commands) that might otherwise go unrecognized.

is that accurate?

zburke:
> So I think your position can be summarized as: you agree that the present behaviour is a bug, but it helps us to diagnose other bugs (choice of shortcuts that conflict with browser commands) that might otherwise go unrecognized.
> 
> is that accurate?

It's close, @MikeTaylor. My position is that existing documentation states `preventDefault()` should be the exception, not the rule, because we aim to choose shortcuts that are free of conflicts. If we supply 20 shortcuts and only one calls `preventDefault()` then this is a sensible guideline. But if we supply 20 shortcuts and 19 call `preventDefault()` then it's time to reconsider our guidelines. 

I haven't done any kind of data collection and analysis to have a sense of how rare/common `preventDefault` is, i.e. to know if it's time to reconsider because our implementation (lots of conflicts) and our guidelines (aim to avoid conflicts) are out of sync.

MikeTaylor:
@zburke 
> It's close, @MikeTaylor. My position is that existing documentation states `preventDefault()` should be the exception, not the rule.

Well, I'm not sure that's really the position. If [the document](https://github.com/folio-org/stripes-components/tree/main/lib/Commander#browser-shortcuts) intends that, it doesn't say explicitly that that's the reasoning. All it says is that you may mind it necessary to do your own `preventDefault`, but there's nothing that says it's been deliberately made this way. (That _may_ be the case, but the doc doesn't say it.)

[The history](https://github.com/folio-org/stripes-components/commit/619abd80ad1c6f715e0b08f465adb02ab269d158) also seems a bit ambivalent about this point. If feels to me like reverting that commit would address your concern about our not finding out when we've inadvertently picked a combination in use elsewhere.

zburke:
I'm interpreting 
> In most cases, the shortcut should be changed to something else. 

to mean "`preventDefault()` should be the exception, not the rule, because we aim to choose shortcuts that are free of conflicts" but you're right that it does not state that explicitly. Do you think better documentation is an acceptable course of action here, or do you think we should follow through with changing the behavior?

PR #1550 / STCOM-836 is about removing a run-time warning that would show up in end-users' consoles.

MikeTaylor:
Reading further back in the history, it seems that "In most cases, the shortcut should be changed to something else" comes from @JohnC-80's initial commit of this file. So if we're going to be originalists, I guess we need John's take on what he intended when he wrote that.

But _are_ we originalists? That was written eight years ago. A lot has changed since then, not least the use of FOLIO in many many locales that were not on the radar back then. Perhaps we should instead think about what's best to do _now_ rather then trying to read the framers' original intentions in `git blame`'s chicken entrails?

And I think my own take is still that it's better the fix the behaviour than to tweak the documentation around it. (We do know that this behaviour has been observed in appplications other than Agreements.)

zburke:
You are correct, @MikeTaylor, that we have added support for many more locales since the original implementation landed. Consequently, since each locale includes its own regionally-defined shortcuts, the pool of universally-unmapped shortcuts that are available to FOLIO keeps shrinking. 

Given "choose shortcuts that are free of conflicts" is becoming a difficult position to hold, we have a choice: 
1. Hold it anyway, yay originalism. Consider each conflicting shortcut individually. Potentially accept increasingly-elaborate shortcuts. We are likely to learn about conflicts from QA during testing.
2. Abandon it, yay short shortcuts. Allow FOLIO's shortcuts to override locale-specific shortcuts en masse. Accept that some locale-specific shortcuts will silently cease to function depending on where keyboard-focus is at the time the shortcut is invoked (see [STCOM-901](https://folio-org.atlassian.net/browse/STCOM-901), [ERM-3311](https://folio-org.atlassian.net/browse/ERM-3311)). We are likely to learn about conflicts from end-users who file a bug in production. 
3. Rework the implementation of shortcuts to make them locale-specific, providing us with a larger pool of available shortcuts but giving us more stuff to maintain.

I'm not ready for (2) yet. While at first glance the [ERM-4070](https://folio-org.atlassian.net/browse/ERM-4070) behavior that led to this PR felt like "we found a bug: FOLIO is not correctly consuming its shortcut-events", I view it differently: "we learned about a shortcut conflict and now have to decide between (1) overriding the conflict (and consuming the event) and (2) choosing a non-conflicting shortcut".

MikeTaylor:
And there is another factor, too. Changing an existing shortcut is a terrible thing to do. The whole point[1] of shortcuts is that they enable experienced users to operate on muscle memory. When we change an existing shortcut we wipe out that ability. Users who have learned the shortcuts would be rightly upset.

So I think that, whatever the solution we come up with, it can't be changing existing shortcuts (although it _could_ involve _adding_ redundant shortcuts, e.g. Alt-E as well as Shift-Alt-E for edit).

With that in mind, and accepting that localizing shortcuts is a whole nother ball of wax, I think a version of solution 1 (originalism) is appealing: we stay with the strategy of trying to pick shortcuts that won't collide with anyone's combination of OS and locale. But we accept that sometimes there will be a conflict despite our best efforts. In that case, the question comes back to the one we started with: what is the least worst outcome?
1A. Let the FOLIO shortcut win.
1B. Somehow make the FOLIO shortcut not happen.
1C. Do both.

I'm assuming we're not even trying to do 1B — especially since, in the contexts where this occurs, there's typically nothing to type the `€` sign (or whatever it is) into. And it seems that 1C (which we do now) is just obviously, painfully, embarrassingly wrong. Which is what leads me back to 1A as the only realistic candidate. I think.

--

[1] OK, not quite the _whole_ point, there are also accessibility benefits for users who can't use a mouse, but I think the point stands.

MikeTaylor:
With all that said, I think we could both benefit from hearing what @JohnC-80 has to say. (I assume he's on vacation, otherwise he'd have commented by now.)

JohnC-80:
@MikeTaylor Sorry for late reply. I believe that usage of `preventDefault()` is best handled case-by-case rather than a blanket solution for all cases. System functionality can be *anything from entering a special character to a browser shortcut key. Preventing is an inconvenience to a browser-breaking bug. It's best to leave it up to the app (given what we've got at our disposal) and why we provide it as cautionary documentation rather than baked-in functionality. The app has better context of what it needs to do and what the user is involved in at the time. 

I'm curious if this problem can be fixed by a different means at the Agreements level... Do you know if this behavior occurs within `ui-users` as well? For agreements where the shortcut opens a completely different route/component and the default functionality is *then applied - far after the fact in JS-event-handling time - is a strange one.

As the numerous facets that you guys have brought up here show, there really isn't a silver bullet for this kind of issue. Even at the per-app level doesn't change the risk of stepping into user-intent/mind-reading territory. There's numerous potential solutions at the system level, but they all take more work than a conversation, unfortunately. It would be nice for us to get out of the 'coming up with defaults that please everyone' business for hotkeys, since it's an impossible undertaking and any possibility of having those continues to be whittled away with each bug like this we encounter. I recently read an article around implementation of keyboard shortcuts that actually discouraged using the `shift` key at all - while [accessibility guidelines discourage non-modifier shortcuts under most circumstances, as those will conflict with assistive technology (phew!)](https://www.w3.org/WAI/WCAG22/Understanding/character-key-shortcuts.html). [KeyboardLayout API that's still left on the table by non-Chromium browsers](https://developer.mozilla.org/en-US/docs/Web/API/Keyboard_API#api.KeyboardLayoutMap) also is an indicator that 'We' as a whole are STILL in the process of figuring all this out.

Solutions in the long game of things (not an all-inclusive list, but the ones I like the most):
- A new tool at our disposal is user preferences - but we have yet to break ground on re-binding shortcuts with this (although it seems feasible)
- Funnel all of these hotkey-able commands through a [single UI/single shortcut via some sort of 'Genie' UI](https://github.com/kentcdodds/genie) - this is another viable way to make FOLIO 'less clicky' and make a potentially good compromise between musophobic users, power users, average joe.

I wouldn't mind either - it all just has to be prioritized on top of all other things.

MikeTaylor:
Thanks for chipping in, @JohnC-80. No problem about the delay, it quickly became apparent that this fix wasn't going to happen in time for Trillium, so in that light a few more days doesn't make too much difference.

> I believe that usage of preventDefault() is best handled case-by-case rather than a blanket solution for all cases. System functionality can be *anything from entering a special character to a browser shortcut key. Preventing is an inconvenience to a browser-breaking bug.

I would like help imagining how this could be a browser-breaking bug. I'm not seeing it. I'm also still strugging to see how the current behaviour can _ever_ be correct.

> It's best to leave it up to the app (given what we've got at our disposal) and why we provide it as cautionary documentation rather than baked-in functionality. The app has better context of what it needs to do and what the user is involved in at the time.

I do accept this as a general principle. But it rather suggests we're going to have to fix this in a couple of hundred places instead of one. Or, much more likely, we'll fix it in two or three places, then spend the next few years getting bug reports like ERM-4070 on a case-by-case basis. Or, more likely still, most users will just accept that FOLIO shortcuts don't work properly, and work around it. ("Oh, yeah, that happens. Just delete the `€` character once you get into the edit screen. It's what we all do. You get used to it after a while.")

> I'm curious if this problem can be fixed by a different means at the Agreements level... Do you know if this behavior occurs within ui-users as well?

ui-users avoids this problem by not focusing an editable field in the new-record page. So the `€` is still emitted (so far as I can tell) but gets dropped on the floor because there's nowhere for it to be pasted. (I even had it suggested that the "fix" for the problem in Agreements was to remove the code that focuses the first editable field. Yay for usability!)

Felix Hemme has noticed that [this also happens in the Licenses app](https://folio-org.atlassian.net/browse/ERM-4070?focusedCommentId=307859). I have no idea how many more places it's happening in.

> For agreements where the shortcut opens a completely different route/component and the default functionality is *then applied - far after the fact in JS-event-handling time - is a strange one.

I agree; yet it consistently happens.

> As the numerous facets that you guys have brought up here show, there really isn't a silver bullet for this kind of issue. Even at the per-app level doesn't change the risk of stepping into user-intent/mind-reading territory. There's numerous potential solutions at the system level, but they all take more work than a conversation, unfortunately.

I think our problem is more fundamental than that. We can't decide (or can't agree) what the right solution even _is_. Once we get that sorted out, we can think about how to impement that solution.

> Solutions in the long game of things (not an all-inclusive list, but the ones I like the most):
>
> * A new tool at our disposal is user preferences - but we have yet to break ground on re-binding shortcuts with this (although it seems feasible)

This is a bit of a tragedy for me, given that [I laid out how to do this nine years ago](https://github.com/folio-org/stripes-core/blob/main/doc/hotkeys.md).

> I wouldn't mind either - it all just has to be prioritized on top of all other things.

Yeah.

So it looks like our options are:
1. Accept the change in this PR and modify the documentation accordingly.
2. Reject the PR and update the documentation with a rationale for why we think the current behaviour is more correct than the alternatives.
3. Something more radical which we all know ain't gonna happen in our lifetimes.

What's it to be?

MikeTaylor:
I guess that answers my question.

JohnC-80:
If we apply `preventDefault()` everywhere, we're saying that regardless of whatever the browser or assistive technology or OS has implemented, we want to completely ignore that in place of our own functionality. This includes a lot of things - we would be saying 'even if you, the user want to type a special character in this field, you cannot because we've mapped it to something else.' I do not think this is the blanket fix we're looking for. 

Overlapping hotkeys with those of browser, OS, etc is NOT the better practice - the tooling within our library has tried to prevent this behavior, but given different keyboard layouts, possibly other cases that we have yet to discover - we've failed to fully cover all the bases.

The case with Agreements, and Licenses is a curious one and deserves more scrutiny. The Users app does have focus the first field on edit, and this bug does not happen there. It's obviously a race condition, as the Users edit view is loading when it initially mounts - the Agreements/Licenses transition is immediate. This solution could be anywhere from where that hotkey is active to how `autoFocus` is handled.

@MikeTaylor, we've got targeted work around hotkeys for Umbrellaleaf - among other changes, `stripes-react-hotkeys` will be brought in as part of stripes-components. At that point we'll be able to focus on the shortcuts issue at the platform level.

MikeTaylor:
Thanks, @JohnC-80. Is there a Jira I can point to for the Umbrellaleaf work?

(I still can't believe we couldn't find a better flower name than Umbrellaleaf, but I guess that is somewhere in the 99th centile of our problems.)

JohnC-80:
@MikeTaylor the ticket we have - seems aged due to the lengthy cycle of Trillium dev- but it's in our radar for a while: https://folio-org.atlassian.net/browse/STCOM-1356

PR #2558, opened 2026-06-22
STCOM-1513  `<Pane>` - add a new `actionMenuToggleProps` prop

BogdanDenis:
## Description
Add a new `actionMenuToggleProps` prop to `<Pane>` and `<PaneHeader>` to be able to pass props to the action menu toggle element.

Required by https://github.com/folio-org/ui-marc-authorities/pull/508

## Issues
[STCOM-1513](https://folio-org.atlassian.net/browse/STCOM-1513)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Switch from `cloneDeep` in `<Callout>` component's state updates- a fix for potential stack overflows due to cloned react component trees 😬 . Refs STCOM-1508.
 * Update placeholder text color for Select to improve contrast. Refs STCOM-1491.
 * Update timezone data. Refs STCOM-1472.
+* `<Pane>` - add a new `actionMenuToggleProps` prop. Refs STCOM-1513.
 
 ## [13.1.0](https://github.com/folio-org/stripes-components/tree/v13.1.0) (2026-04-14)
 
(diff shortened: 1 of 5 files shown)


zburke (approved):
The component path here is Pane > PaneHeader > Dropdown > Button. This PR provides prop-drilling in Pane via `actionMenuToggleProps` because even though Dropdown's `renderTrigger` function accepts `ariaProps`, at present we don't have a way to drill through to it from the outside. That is, even though `Dropdown::renderTrigger()` accepts `ariaProps`, `Pane` does not drill them through, so this PR supplies them via `actionMenuToggleProps`.

I am OK with this approach, though I am a bit leery. It supplies the ability to provide all kinds of props, not just the aria-related props mentioned in parent ticket, [UIMARCHAUTH-521](https://folio-org.atlassian.net/browse/UIMARCAUTH-521). Expanding the scope is not necessarily bad. Maybe it's even good, collecting all the action-menu props under a single object instead of letting them proliferate as `actionMenuFoo` and `actionMenuBar`. 

Alternatives would be specific props just for aria, or maybe fetching them from context instead of drilling. @JohnC-80, what do you think?

Dmitriy-Litvinenko (approved):

JohnC-80 (approved):
Yeah, I'm okay with this as a way to expand the versatility for specifying props for the `actionMenuToggle`. It's not unheard of that we would actually need to pass other app-level props into it. A11y-wise, `aria-*` isn't always the way, but about 80% with this kind of thing. Audits run using some sort of voice command software will run into issues, since the visible label of the elements are still identical - but it's a different design discussion about how the label is dealt with.

PR #2561, opened 2026-07-08
STCOM-1523 - Clear outline from PaneHeader for focus styling

JohnC-80:
Focus of the PaneHeader specifically gets used in `stripes-core`'s `<Settings>` component. This is an aesthetic part of the fix at the [STCOR-1083 PR](https://github.com/folio-org/stripes-core/pull/1747).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Update timezone data. Refs STCOM-1472.
 * bugfix - `<SessionConfirmationModal>` shouldn't call its `onConfirm` prop every render. Refs STCOM-1512.
 * `<Pane>` - add a new `actionMenuToggleProps` prop. Refs STCOM-1513.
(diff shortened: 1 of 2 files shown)


zburke (commented):
@JohnC-80, I understand why we needed the STCOR PR -- we should've been passing the ref, not its referant -- but I don't understand why we need this work. Is this, too, a bug that's been lurking undetected for several years? What caused us to notice it now/not to have it prior to that STCOR PR when focus was properly being set?

JohnC-80:
@zburke This struck _me_ as a rough-edge inconsistency in styles. There are a handful of places in FOLIO that do send focus to the header ex: Inventory with a single search result sends focus to the header of the details Pane. Let me confer with Kimie for this change as it is a cosmetic issue - not so much an a11y one since we do have a focus indicator there at the pane header.

zburke (approved):
Design guru says, "Yep, this was a bug we just hadn't noticed".

PR #2564, opened 2026-07-13
STCOM-1524 MCL test flake - scroll less on a teeny grid in test

JohnC-80:
The MCL in the test is virtualized - so it's getting rid of rows as you scroll. Scroll too far, and you've removed the very row you are trying to assert on or click.

Failing tests would shout something like 'I can't find MCL row 11'...
and the listing of almost-matching stuff starts at 14 or 15... scrolled too far.
How does one scroll on pixels that they cannot see? Are they any bigger or smaller because you CAN'T see them?

Locally, the test renders a teeny grid:
image

And you can't see it, but in a blink, the grid actually shrinks from a large width, which can cause the height of elements to shift as rows of text wrap, down to 300.

Row 2 is selected, the grid shrinks (scrolls as it tries to keep row 2 visible), scrolls more(the action `await mcl.scrollBy({ direction: 'top', value: 200 });`, select again (if possible - where our tests were failing before), assert the selected class is applied to the recently selected row.

So a few safety measures in the fix - 
- wait for the grid to shrink (`mcl.has({width: 300}`)
- scroll less
- select a row that's roughly in the middle of the visible list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * bugfix - `<SessionConfirmationModal>` shouldn't call its `onConfirm` prop every render. Refs STCOM-1512.
 * `<Pane>` - add a new `actionMenuToggleProps` prop. Refs STCOM-1513.
 * `<Pane>` - remove focus outline from `<PaneHeader>`. Refs STCOM-1523.
+* Fix flaky test with virtualized `<MultiColumnList>`. Refs STCOM-1525.
 
 ## [13.1.0](https://github.com/folio-org/stripes-components/tree/v13.1.0) (2026-04-14)
 
--- a/lib/MultiColumnList/tests/MultiColumnList-test.js
+++ b/lib/MultiColumnList/tests/MultiColumnList-test.js
@@ -454,7 +454,7 @@ describe('MultiColumnList', () => {
 
         describe('adjusting to width change', () => {
           beforeEach(async () => {
-            await mcl.scrollBy({ direction: 'top', value: 600 });
+            await mcl.scrollBy({ direction: 'top', value: 400 });
           });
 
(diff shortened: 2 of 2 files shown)


zburke (approved):
Thank you! I figured it was sth like this, but also couldn't figure out why a table with 100 rows or less would be virtualized, nor did I see anything about virtualization in the test setup :/ I think the implication is that virtualization is enabled by default. That feels backwards to me; am I in the minority here? 

Also, grateful for the fix, but I still don't understand why this started popping up now after years of functioning correctly :/ :/

JohnC-80:
@zburke This test uses `<MCLHarness>`, which, as a test-harness *does have `virtualize` set to true by default and just gives a setup to test other things like loading data, resize, columns, etc. `MultiColumnList` itself does not virtualize by default. I didn't live enough in the meta around these tests as much as focus on getting the flaky test to have some consistent pass rate. I believe that the variability came with the length of the strings that faker was generating and the affect that had on the scroll position of elements when the width of the MCL shrinks during the test run (The width change is also a part of MCL harness behavior.) Bigger fish, but will probably see this again with some optimization work to try and improve performance at the scale case we have with Eureka.

zburke:
@JohnC-80, that explanation sort of makes sense, but if faker's version did not change, why is its output suddenly reliably shorter than previously? That explains why the test is failing, but not why we are suddenly encountering this failure. 

Anyway, for me, this again comes down to "9 times out of 10, I find the side-effects of virtualization to be worse than its benefits". Given my druthers, I would lean less on "make virtualization great again" and more on "avoid virtualization when possible, choosing alternatives such as paging". I understand there is no silver bullet here and that paging has its own wrinkles.

PR #2565, opened 2026-07-15
STCOM-1356 directly implement HotKeys and FocusTrap

zburke:
Instead of importing `<HotKeys>` and `<FocusTrap>` from `@folio/stripes-react-hotkeys`, implement those components locally. 

A future update can enable React v19 compliance by removing the `ReactDOM.findDOMNode()` fallback in HotKeys::cDM(). This will make the `attach` prop to `<HotKeys>`, and therefore the `scope` prop to `<HasCommand>`, required; previously, they were optional previously.

Refs [STCOM-1356](https://folio-org.atlassian.net/browse/STCOM-1356)


zburke:
@JohnC-80, while test coverage via Commander was good, I just couldn't bear the idea of introducing hundreds of lines of code without corresponding tests. For the most part, after the addition of `attach` to handle cases where the `ReactDOM.findDOMNode()` fallback kicked in, this went smoothly, but there are a few persistent failures I haven't been able to reconcile.
* several tests related using `noWrapper` expect to see `tabindex="-1"` in the rendered output, but I cannot figure out how that would happen given the return value is merely `children` in that case. Am I missing something?
* ~a few tests for nested handlers do not behave the way I expect.~ typo; fixed; nevermind.

Can you take a look at these skipped tests and let me know how you think we should proceed? We may decide that the change in behavior is necessary due to the changes in how attach/scope/refs are handled and that the test needs to be updated to reflect this new API, or maybe we just need to adjust how the test is written.

JohnC-80:
@zburke the test failures are a definite example of how hard this component leaned on `findDOMNode` in the first place. 

The cDM of `FocusTrap` executes before the cDM of `HotKeys` - and `FocusTrap`'s `element` prop wouldn't get updated without a re-render (which line 142 of `HotKeys` might have caught if it wasn't borked to start with - `attach === 'undefined'` 💀).

The test having no further updates to state/no wrapping component with a state update to refresh leaves it for a failure since word hasn't gotten to `FocusTrap` yet that it has an `element`.

Also - this test suite runs in karma, but it's a very jest suite - no async beforeEach and pieces of the test suite from S-R-H that was all its own, complete with the dependency on `jest-config-stripes`. 

I think we can do a quick refactor on these and lose some of the baggage it's carrying along to conform it to the other tests he have. Happy to pick it up from here.

lib/HotKeys/HotKeys.js
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  componentDidMount() {
+    const { attach } = this.props;
+
+    // handle refs...
+    if (attach) {
+      this.attachment = attach.current ? attach.current : attach;
+    } else {
+      this.attachment = ReactDOM.findDOMNode(this);
```
JohnC-80, line 127: There's a fallback mechanism in `stripes-react-hotkeys` that goes along with this - 
https://github.com/folio-org/stripes-react-hotkeys/blob/master/lib/HotKeys.js#L142 `this.forceUpdate()` - which will force a re-render now that an attachment has been found via the deprecated `findDOMNode`. The old code also passed `this.attachment` through to `<FocusTrap>` instead of `this.props.attach`... I know we've mostly covered it in the usage that's only in `CommandList`/`HasCommand`.. but 😬 .

lib/HotKeys/HotKeys.js
```diff
@@ -0,0 +1,301 @@
...
+      'keyMap',
+      'handlers',
+      'focused',
+      'attach',
+      'HKcontext'
+    );
+
+    return (
+      <HotKeysProvider value={this.state.HKcontext}>
+        <FocusTrap {...focusTrapProps} onFocus={this.onFocus} onBlur={this.onBlur} element={this.props.attach}>
```
JohnC-80: The old code sets `this.attachment` for element rather than `this.props.attach`.
https://github.com/folio-org/stripes-react-hotkeys/blob/master/lib/HotKeys.js#L288

JohnC-80 (changes requested):

JohnC-80 (approved):

JohnC-80:
The changes look correct, but I'm glaring at the tests, so I don't think we're there just yet. Will look when I have a moment.

PR #2566, opened 2026-07-17
ModalFooter documentation

zburke:
Describe the useful-but-unusual way `<ModalFooter>` expects its children elements to be specified to simplify focus management:

> Note the primary button is the first element in the example above, but the
> `row-reverse` rule in this component's CSS reverses the display order. While this
> may seem counterintuitive, it is not without reason: providing the primary
> element first allows the default focus-handling `<Modal>` to be very simple,
> focusing on the first focusable element.


JohnC-80:
@zburke Thank you for adding this bit. It wouldn't hurt to also include the `dismissible` advice in `Modal`'s readme - 
Something like:
```
This prop is only necessary if your Modal is information-only and does not provide the user with a 'Cancel' button. 
````

JohnC-80 (approved):

Dmitriy-Litvinenko (approved):

PR #2567, opened 2026-07-21
STCOM-1529 avoid initially focusing the close 'X' in Modals

JohnC-80:
## [STCOM-1529](https://folio-org.atlassian.net/browse/STCOM-1529)

Focusing the ‘close button’ first causes an odd UX. It’s like greeting a guest by showing them to the exit. The focus logic of the modal, and possibly stripes-components' focus utilities should be adjusted so that the close button is no longer found as the candidate to focus.
This could happen by providing exception classes/selectors to the getFirstFocusable, or implement an attribute that could be applied to elements that need to be focus-ignored when finding the first element data-focus-exclude.
- If the Modal contains a form, the first interactive element of that form should be focused.
- If there is no form, if the Modal has a footer, the ‘primary’ action of the footer should be focused first.
- No form or footer, focus will be on the container, with the close button only a tab away.

### Approach
Apply the data-attribute `data-focus-exclude` to the close 'X'. It will cause the `getFocusableElement` utilities to ignore the element it's applied to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * `<Pane>` - remove focus outline from `<PaneHeader>`. Refs STCOM-1523.
 * bugfix - `<Selection>` - Escape regex metacharacters when filtering, so filtering by `*` (or other special characters) no longer throws. Refs STCOM-
 * Fix flaky test with virtualized `<MultiColumnList>`. Refs STCOM-1525.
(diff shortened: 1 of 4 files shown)


zburke (approved):
Thank you! OMG, tests AND documentation 😍 I think the Sonar failure is due to a transient network blip. Hopefully it passes after a re-run.

PR #2568, opened 2026-07-21
STCOM-1531 - Use 'nodeRef' prop with react-transition-group components to avoid `findDOMNode` for R19 migration.

JohnC-80:
## [STCOM-1531](https://folio-org.atlassian.net/browse/STCOM-1531)

- `Callout`/`CalloutElement.js` — added a `useRef`, attached it to the transitioned wrapper div, passed as `nodeRef`.
- `Popover` — reused the existing `popoverRef` (already attached to the overlay div) as `nodeRef`.
- `MessageBanner`  — since the forwarded ref can be `null`, added an `internalRef` fallback (matching Popover.js's existing `ref || internalRef` idiom) and used that combined ref for both the DOM node and `nodeRef`.
- `Popover`/`LegacyPopover` — added a new `this.transitionRef`, attached it to the transition wrapper div, passed as `nodeRef`

Also - MCL - it was a comment! thanks for the heartburn, past self!


zburke (approved):
😌 `<Transition>` handles this ref for us. Thank you for finding, filing, and fixing this!

PR #2570, opened 2026-07-24
STCOM-1535 - Revisit for false positives with Axe updates

JohnC-80:
This update primarily affects MCL. It's never been flagged in accessibility audits, but axe continues to disagree. We really need to move our axe version forward, but false-positives make things difficult for everyone. Failures in stripes-components will mean failures reported in modules, false or not. As long as whatever we do to appease axe does not disrupt the clean accessibilty tree, I consider us in fine shape...

Before: Axe report: (Full disclosure, one of the 2 errors were in the storybook UI)
image

Before: accessibility tree (chrome dev tools)
image

After: - Axe report...
image

After - Accessibility tree:
image

Approach/what happened.
Shifting the `role=rowgroup` from the MCL RowContainer to the MCL Scrollable seems to have made the difference that axe was looking for. My A11y tree is unharmed.
Additionally, I applied the `rowgroup` role to the header container, since it's effectively a `head` of a table.

Also, removed the offending usage examples from storybook - no more `onRowClick` in example code.

It's bound to be an error that UI-modules will see if they are *still not implementing links in their first columns.

Update: checked in Inventory - Before
image

After -
image

### DIFF
--- a/.storybook/IntlWrap.js
+++ b/.storybook/IntlWrap.js
@@ -10,7 +10,7 @@ function updatePageDirection(direction) {
 
 const IntlWrap = ({ messages, children }) => {
   const [locale, updateLocale] = useState('en');
-  const [direction, updateDirection ] = useState('LTR');
+  const [direction, updateDirection] = useState('LTR');
 
   const handleDirectionChange = (e) => {
     updateDirection(e.target.value);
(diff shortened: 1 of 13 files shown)


zburke (approved):
Thanks for pursuing this! Addressing axe's complaints at the source in MCL, rather than [ignoring them as "false positives" after the fact during testing](https://github.com/folio-org/jest-config-stripes/pull/33/changes/fd40fa3882858679be88f18876b379b0fef123f0#diff-0f6b98b3e85f2156b97b1336d580264e804562f09d05945ce907cd4c3f644887R33-R50), is definitely a cleaner approach. I'll update that PR and remove the "false-positives" filter. 

My only concern is that this tweak to appease axe appears to irritate Sonar. 😒 Do you think there is anything we can/should do to placate Sonar, or we can we safely disregard its concern?

JohnC-80:
Merging this one.. sonar sees the 'maybe it has tab-index, maybe not' as a reason to warn, as well as the 'oh this is a role=row, should be a 'tr' in a separate component rather than understanding it's not rendered in a `tbody`...

PR #2573, opened 2026-07-29
STCOM-1536 rnd 2 - MCL CSS vs JS

JohnC-80:
This essentially converts the MCL layout to a flexbox layout with `flex-direction: column` set - so that the positioning/height of its major containers will just be handled without requiring measuring and calculation by JS.

Why? Why don't we just let the browser do the work rather than the uh.. browser/library/developer do the work - and we'll get to carve out a bunch of the JS in the process.

This removes multiple reads of `offsetHeight` - which are sure-fire culprits for layout-thrashing (inefficiency) since they incur  a reflow in the browser when they're executed every render. This is the stuff that janky UI's are made of and we don't need that life anymore.

The bug this ultimately does resolve:
Before:

https://github.com/user-attachments/assets/73f9c233-c87e-4637-a344-c45b0b714f39

After:

https://github.com/user-attachments/assets/ee012f0c-757f-4bae-ac1a-549a47f63f4c

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * `<Modal>` no longer focuses the close 'X' on initial open. Refs STCOM-1529.
 * Apply `nodeRef` prop to `react-transition-group` components to avoid `findDOMNode`. Refs STCOM-1531.
 * Update MCL rendering to conform to axe tests. Refs STCOM-1535.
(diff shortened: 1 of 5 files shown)


zburke (approved):
Results grids in user-search, inventory-search, plugin-find-user all look good to me. I moved the story from UIIN to STCOM :)

PR #2575, opened 2026-08-04
Implement/Add useCrossTabSessionStorage hook

JohnC-80:
Based on comments for https://folio-org.atlassian.net/browse/STCOM-1460?focusedCommentId=276289
done to enable implementation of  https://folio-org.atlassian.net/browse/UISAUTHCOM-98

We've ideanated about something that carries the ephemerality of session storage, but exists cross-tab like localstorage... this is a bit of both.

Localstorage uses a tab registry (unique id's + timestamp) per tab. Values are stored in sessionStorage on a key designated by the app.

A new tab registration (writing to localstorage) will be picked up in a session event. The catching tab will broadcast its value to sync the new tab.

Current 'live' registrations prune on a 5-second heartbeat with a stale threshold of 15-seconds. Each tab is responsible for keeping itself alive and updating this timestamp. If it's old, it gets cleaned.

Pushing this PR up in draft for discussion.

EDIT: alternative library that does something similar, lightweight but kinda new (in 0 major range) but zero deps! [TabStateSync](https://github.com/robertluiz/TabStateSync)

### DIFF
--- a/hooks/tests/useCrossTabSessionStorage-test.js
+++ b/hooks/tests/useCrossTabSessionStorage-test.js
@@ -1,0 +1,197 @@
+import {
+  describe,
+  beforeEach,
+  afterEach,
+  it,
+} from 'mocha';
+import { converge } from '@folio/stripes-testing';
+import { getHookExecutionHarness } from '../../tests/helpers/getHookExecutionResult';
+import { mount } from '../../tests/helpers';
+import useCrossTabSessionStorage, {
+  getValueKey,
+  getRegistryKey,
+  STALE_THRESHOLD_MS,
+} from '../useCrossTabSessionStorage';
+
+const PREFIX = '@folio/stripes-components::crossTabSession::';
+
+const clearAllSyncedStorage = () => {
+  Object.keys(localStorage)
+    .filter((k) => k.startsWith(PREFIX))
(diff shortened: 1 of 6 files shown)


zburke:
@JohnC-80, I wonder about flipping the implementation here, i.e. instead of using BroadcastChannel with SessionStorage and working hard to avoid LocalStorage, what if this were a stripes-core implementation and we just added a `FolioSessionData` LocalStorage key gets initialized on login and destroyed on logout? 

That might be simpler to maintain, with many of the same perks. If the value is a serialized object, it should be straightforward to implement [the Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage) (clear, getItem, removeItem, setItem). LocalStorage events also emit across tabs so you could implement listeners (e.g. 'rerender if X changes') without having to fuss with the BroadcastChannel. I don't see any flaws in this implementation, but at the same time it feels very ... sweaty.

JohnC-80:
Closing for now. Current cases with stuff-that-doesn't-belong-in-local (STDBL) can just be resolved by *not putting the STDBL in there. Maybe this will work for multi-tab support, but also, what if we don't? If anything gets over that hurdle, the branch'll be here....

PR #2576, opened 2026-08-05
STCOM-1541 - Color contrast issues in stripes-components. 

JohnC-80:
## [STCOM-1541](https://folio-org.atlassian.net/browse/STCOM-1541)
Inventory and ui-lists were cited in a11y audits for color contrast issues. The cases were obscure, but worthy of coverage - focus styles, specifically for elements in selected mcl rows and `<IconButton>`s.

As most styling is left up to stripes-components, the fixes reside here.

As an additional piece, added the first draft of a color contrast/chooser tool to help developers choose a contrasting color, at the very least, a provided CSS variable that they can use in their own code.

Resulting contrast of some cited elements:
Before:
image

After:
image

The contrast tool/page:
Allows the user to pick a color for a background color, and gives resulting contrast ratios. More useful, a handful of component examples to test interaction styles/colors with within stripes-components.

https://github.com/user-attachments/assets/1fe6ce0c-31f6-4c54-a3f4-eb5089fdf7f4


zburke (approved):
❤️ Big thanks for adding the tooling to assess future color-changes in addition to fixing the contrast issues in the current palette. As somebody with poor color vision, this is something I encounter almost every day, both online and IRL.

PR #2578, opened 2026-08-14
STCOM-1543 <TransitionGroup> creates random div with <Modal>

JohnC-80:
### Problem
In this day an age of flexbox, random divs can bork your nice layout. `<TransitionGroup>` commits this sin. It makes elements wrap unexpectedly, throwing the user into another world of distraction. It's enough to make a designer's eye twitch.

### Approach
Add component={null} prop to `<TransitionGroup>` in `<Modal>`

### Offensive

https://github.com/user-attachments/assets/5652113a-7ce6-42f2-86c0-defc70fece48


zburke (approved):
Børk! Børk! Børk! 🇸🇪 🧑‍🍳 

If we can easily-ish find the PR that added the flexbox styling that introduced this glitch, it would be nice to link it up. I'm not trying to point fingers so much as have additional context for the next time we're trying to troubleshoot similar issues.

PR #2579, opened 2026-08-14
STCOM-1544 - Dark/disappearing button text in the active state.

JohnC-80:
### Purpose
This blemish has been a peeve for a while. A better focus on our CSS variables brought this to light, so pushing this fix.

### Before
peeve

### After
fixed


zburke (approved):
❤️ This is nice polish. Thank you. 

This is so pedantic I cannot believe I'm actually mentioning it, but the CHANGELOG doesn't mention the actual fix, which was to supply a missing/sufficiently specific style in Button.css. The other changes merely replace a hard-coded value with a reference to the same value, now stored in a variable.

PR #2581, opened 2026-08-14
STCOM-1533 Improve performance on pane resizing.

JohnC-80:
Data-heavy, component-heavy views can cause some pane-resizing jank. This PR works to help Paneset's end of the deal and not add more jank due to its implementation.

Draft for now with the end of the Spike. Ready after a bit more sleuthing for jank-culprits.

So far:
- Swapped `<Paneset>`'s `setStyle` to a noop - it's only there to fulfill Paneset's own expectation when treating a nested Paneset. This avoids `getBoundingClientRect`.
- moved handle position computation *out of the standard render process, so it only updates when resize, blocking status, new panes, etc and not when the handle is being dragged.

### DIFF
--- a/lib/Paneset/PaneResizeContainer.js
+++ b/lib/Paneset/PaneResizeContainer.js
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect, useCallback } from 'react';
+import React, { useState, useRef, useEffect, useCallback, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import cloneDeep from 'lodash/cloneDeep';
 import isEqual from 'lodash/isEqual';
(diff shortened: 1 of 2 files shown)


lib/Paneset/PaneResizeContainer.js
```diff
@@ -184,37 +182,48 @@ const PaneResizeContainer = ({ isRoot, children, onElementResize, parentElement,
...
+    [computeHandlePositions, update, blocking]
+  );
 
-    return handleElements;
-  }, [activeHandle, handleHandleMouseDown, handles]);
+  // No DOM reads here, so this is cheap to rerun on every render (mousedown/mousemove/mouseup).
+  const renderHandles = () => handlePositions.map((pos) => {
+    if (!pos || !pos.render) return null;
+    return (
+      <PaneResizeHandle
```
zburke, line 218: Maybe a stronger way to write this is to check for the conditions you need, else return null?
```
if (pos?.render) { return <PaneResizeHandle ...> }
return null;
```
or 
```
return (pos?.render) ? <PaneResizeHandle ...> : null;
```

PR #2584, opened 2026-08-19
STCOM-1545 Shield `onConfirm()` with ref so that it only happens once per open.

JohnC-80:
### Problem
1. Click → `handleSessionConfirm` runs → `onConfirm()` → `onConfirm()` Logic fires. No guarantee that `open` was set to `false`.
2. box was checked, so `handleSessionConfirm` then calls `setSuppressStorage(true)`.
3. Re-render: `suppressStorage=true`, `open` still `true` → effect condition `suppressStorage && open` is now true → `onConfirm()` runs a second time . 💥 

### Approach
Shield it up with a ref. `onConfirm()` is set to run if `suppressStorage` is true AND `open` is true (the suppressed case where the use checked the 'never display again' box) - this can run a second time as described above, so if the effect has run, set a ref when it runs with the user's click that will prevent it from being run if the modal hasn't been closed and re-open.

Very effect-y... so comments!

Also in this PR:
Don't even render the ConfirmationModal if `suppressStorage=true` (can prevent flashes of the modal from showing up.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Update MCL rendering to conform to axe tests. Refs STCOM-1535.
 * Convert MCL layout elements to a `flex-box` column. Resolves issues where pagination buttons would not show up on the first render. Refs STCOM-1536.
 * Resolve color contrast issues on selected MCL rows, IconButtons, focus styles. Refs STCOM-1541.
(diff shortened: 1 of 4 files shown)


zburke (approved):
Yes, very effect-y. Thanks for the explanatory comments in the PR and in the code!

PR #2585, opened 2026-08-25
STCOM-1547 correctly parse locale strings in Datepicker

zburke:
Use `Intl.Locale` to parse locale-strings since that API is now baseline widely available. Overly simplistic parsing in the previous implementation faltered when locales included number format settings, e.g. `es-u-nu-latn`. While there was not a good reason to provide such settings (number format only needs to be specified when you want to override a locale's default), providing it shouldn't cause errors.

Refs [STCOM-1547](https://folio-org.atlassian.net/browse/STCOM-1547)


JohnC-80 (approved):
Beats the heck out of manual parsing! Dog-earing in mind for `Intl.locale.getWeekInfo()` That could hopefully help us use those static first days...

zburke:
> Beats the heck out of manual parsing! Dog-earing in mind for Intl.locale.getWeekInfo() That could hopefully help us use those static first days...

Good thinking, filed [STCOM-1549](https://folio-org.atlassian.net/browse/STCOM-1549)

PR #2586, opened 2026-08-26
STCOM-1538 AutoSuggest failing under axe-core 4.13.0

JohnC-80:
## [STCOM-1538](https://folio-org.atlassian.net/browse/STCOM-1538)

```
FAILED TESTS:
  AutoSuggest
    rendering
      entering an existing value
        ✖ has no axe errors - open list
          Chrome 151.0.0.0 (Mac OS 10.15.7)
        AxeError:  Axe violation(s): 
         Issue #1: aria-allowed-attr - Elements must only use supported ARIA attributes
            description: Ensure an element's role supports its ARIA attributes
          helpUrl: https://dequeuniversity.com/rules/axe/4.12/aria-allowed-attr?application=axeAPI
              html: <div class="downshift---WzkTb" data-test-autosuggest="true" aria-expanded="true">
            target: .downshift---WzkTb
              Failure Summary:
                Fix all of the following:
              -  ARIA attribute is not allowed: aria-expanded="true"
      
```

As the failure says, we've got `aria-expanded="true"` in a place where it doesn't belong. This PR removes that attribute and `extend`s the local interactor until https://github.com/folio-org/stripes-testing/pull/7584 is merged. (It will have to be merged, or the `Autosuggest interactor's `open` filter will break, since it refers to the wrongfully-placed attribute that this PR removes.


zburke (approved):
Happy to approve this if you want, but since you already fixed this [at the source](https://github.com/folio-org/stripes-testing/pull/7584), I'd rather see us bump our stripes-testing dep to ^5.1.0 and call it a day.

PR #2587, opened 2026-08-27
STCOM-1548 Improve labels and icons positioning in `<RepeatableField>` component when some labels are not present

BogdanDenis:
## Description
Make margin of the `legend` element of the repeatable field consistend with other labels in folio
Add a `hasRemoveButtonMargin` prop to be able to control margins of the remove field button. Currently, when individual field's labels are not present - remove icon will be out of alignment with the field.

## Screenshots
image

## Issues
[STCOM-1548](https://folio-org.atlassian.net/browse/STCOM-1548)


zburke (approved):
LGTM. @JohnC-80, please take a look and confirm you agree :)

JohnC-80 (approved):

PR #2592, opened 2026-09-01
POC: new wysiwyg component implementation using the `lexical` library

BogdanDenis:
## Description
`react-quill` has accessibility issues, and we need to find a replacement for that library for our WYSIWYG needs.
[https://lexical.dev/](https://lexical.dev/)  is a text editor framework with available packages for React. One thing that it focuses on is accessibility, so it’s a good potential candidate to use for our needs.
Since it’s a framework - it only provides core functionality and utilities to handle field state and HTML/markup field manipulation and formatting. All the UI elements should be defined by us.

Expected outcome of this POC - a WYSIWYG text editor component, which supports following formatting options:
- heading selection
- bold
- italic
- underline
- link
- ordered list
- unordered list
- clearing formatting

Since the main concern is functionality - styling and icons are not a priority for this POC, and could differ from current Editor component and overall Folio styling

## Approach
New components and structure:
`<LexicalEditor>` (can be renamed, for now it's named like this to make it separate from the existing `<Editor>` component) -  defines high-level configuration to the `lexical` library: available formatting nodes, formatting styles, prepopulating with HTML string from props.
`<LexicalEditorUI>` - composes together several `lexical`'s plugins. This component can access lexical context, so state reads/updates could be placed here, or somewhere between this component and `<LexicalEditor>` in the hierarchy.
`<ToolbarPlugin>` - this is the editor's toolbar component:
image
This component contains the state of the toolbar, defines/dispatches commands for each of the formatting buttons.
`<HeadingSelect>` - a heading select dropdown. Contains UI and logic to format selected text to be `h1`, `h2` etc...

Added a README with examples on how to integrate this component into a form, and notes for developers on how to use Lexical API to add simple formatting options.

## TODO:
Add localization to hard-coded text
Use better icons
Update styling/use another component for the heading selector
Check a11y - good news it that we have complete control over the rendered UI and can do whatever we want with it.


JohnC-80:
Thank you @BogdanDenis for this much-needed effort! Would you mind adding a storybook story for this?

BogdanDenis:
> Thank you @BogdanDenis for this much-needed effort! Would you mind adding a storybook story for this?

definitely, this work is still in progress - I will add storybook and a readme on this component

BogdanDenis:
Decided to go with TipTap WYSIWYG implementation instead of lexical
https://github.com/folio-org/stripes-components/pull/2594

PR #2594, opened 2026-09-09
STCOM-1554 POC replace React-Quill with TipTap editor

BogdanDenis:
## Description
`react-quill` has accessibility issues, and we need to find a replacement for that library for our WYSIWYG needs.
[TipTap](https://tiptap.dev/)  is a text editor framework with available packages for React.
Since it’s a framework - it only provides core functionality and utilities to handle field state and HTML/markup field manipulation and formatting. All the UI elements should be defined by us.

Expected outcome of this POC - a WYSIWYG text editor component, which supports following formatting options:
- heading selection
- bold
- italic
- underline
- link
- ordered list
- unordered list

Since the main concern is functionality - styling and icons are not a priority for this POC, and could differ from current Editor component and overall Folio styling

## Approach
Refactored `<Editor>` component to a functional component to be able to use TipTap's hooks. This component provides a high-level configuration of TipTap and keeps the original `<Editor>` component's API, so it's mostly backwards-compatible with the previous version (removed `modules` and `format` props that used to configure Quill's additional functionality)
New components and structure:
`<Editor>` -  defines high-level configuration of the `TipTap` library: available formatting nodes, formatting styles, prepopulating with HTML string from props.
`<Toolbar>` - defines the UI of the editor and manages the state of the toolbar (which formatting options are selected for a current text selection).
image
`<LinkControl>` - a custom control for the toolbar that pops up and allows to enter a link URL.
image

For this integration, we don't need to update the README much (only removed mentions of Quill, and removed `modules` and `format` props).
## TODO:
Use better icons


JohnC-80:
@BogdanDenis This is looking good and I think this is the direction we'll end up going in. The lexical editor was set up as a separate component, but this PR re-implements Editor in place. Was that intentional? I'm happy for it to be what we settle on, but maybe it's best to implement separately so that we can test the new component within modules first before switching completely.

zburke:
@JohnC-80, initially I shared your testing concern (maybe "implement separately so that we can test the new component within modules first"), but it turns out there are literally only two apps importing Editor (ui-remote-storage, SSC's Notes). The other places where I expected to find it import from stripes-template-editor, and that one directly invokes `<ReactQuill>`. I'll investigate ui-remote-storage and stripes-smart-components in more detail in [STRIPES-1047](https://folio-org.atlassian.net/browse/STRIPES-1047). 

Given the the blast-radius here looks pretty small, I lean toward accepting the existing, in-place implementation. That is, even if it requires some changes, changing only two places feels to me like a small price to pay for sticking to the existing component. What do you think?

JohnC-80:
@zburke Sounds like a reasonable plan, then. A good step on our path to reach v19. Hopefully this improvement can corral a few of the strays back to using this. I guess this still lacks Icon work, some testing, @BogdanDenis ?

BogdanDenis:
@zburke I think it's only the SSC Notes component that uses Editor. ui-remote-storage has it's own `Editor` component that does something completely unrelated, but I don't see any imports from SC based on [my search](https://github.com/search?q=repo%3Afolio-org%2Fui-remote-storage+editor+language%3AJavaScript&type=code&l=JavaScript)

And I made basically no changes to the `<Editor>`'s API, with the exception of `formats` and `modules` that were removed, but they weren't being used anyways...

And @JohnC-80 I've made some changes to the icons so they are more consistent in look and size and added some more tests.
Most of the formatting options are covered by tests, except for "bold", "italic", and "underline". For some reason value selection is lost when performing a click on an icon button. I think it's just because these clicks are simulated events, and they behave differently from real user interactions.

JohnC-80:
@BogdanDenis Giving this a test-drive in storybook, Looks like the content-editable divs are not currently focusable via the keyboard (tab). The rest of the toolbar elements are tabbable/announced fine. The formatting dropdown does need some focus-styling applied. Likely the `contenteditable` `div` will need this as well.

We should also focus the input for the 'link' overlay when that dialog opens.

Minor style - just a border-bottom on `.toolbar` element would be good to just visually separate the toolbar from the editable area.

Almost there!

